PR #6, opened 2024-01-09
feat: first draft of lint rules document

a-frantz:
Here we have it! The first draft of our comprehensive lint rule set. There are some controversial rules in here, so I expect a bit of debate to go down. I opted to open this PR directly instead of sharing in our Slack channel, as it will be good to have public record of _why_ we chose to enforce what we decide on. I doubt this document gets through review without a few rounds of revision.

This is **not** any kind of document we should feel beholden to in the future. As stated in the document, these rules are "up for debate" for now. In my imagining, we won't consider a rule "settled" until someone actually _implements_ the rule in code. As such, I would more think of this as a document to spark thought and debate. Maybe act as a reference for anyone who wants "cement" a rule in code.

So there it is! Feel free to rip this document to shreds while reviewing it. Let's try and polish this up as much as possible.

### DIFF
--- a/lint-rules.md
+++ b/lint-rules.md
@@ -209,14 +209,17 @@ The following tokens should **not** be **followed** by whitespace (including new
 Opening brackets (either `(` or `[`) should **not** be **followed** by a **space** but **may** be **followed** by a **newline**.
 Closing brackets (either `)` or `]`) should **not** be **preceeded** by whitespace **unless** that whitespace is **indentation**.
 
-Sometimes a long expression will exceed the maximum line width. In these cases, one or more linebreaks should be introduced. Line continuations should be indented one more level than the beginning of the expression. If bracketed content (things between `()` or `[]`) must be split onto multiple lines, a newline should follow the opening bracket, the contents should be indented an additional level, then the closing bracket should be de-indented to match the indentation of the opening bracket.
+Sometimes a long expression will exceed the maximum line width. In these cases, one or more linebreaks must be introduced. Line continuations should be indented one more level than the beginning of the expression. There should _never_ be more than one level of indentation change per-line.
 
-If you are line splitting an expression on an infix operator, the operator should be on the continued line.
+If bracketed content (things between `()` or `[]`) must be split onto multiple lines, a newline should follow the opening bracket, the contents should be indented an additional level, then the closing bracket should be de-indented to match the indentation of the opening bracket.
+
+If you are line splitting an expression on an infix operator, the operator and at least the beginning of the RHS operand should be on the continued line. (i.e. an operator should _not_ be on a line by itself.)
 
 If you are using the `if...then...else` construct as part of your expression and it needs to be line split, the entire construct should be wrapped in parantheses (`()`). The opening paranthesis should be immediately followed by a newline. `if`, `then`, and `else` should all start a line one more level of indentation than the wrapping paratheses. The closing paranthesis should be on the same level of indentation as the opening paranthesis.
+
 If you are using the `if...then...else` construct on one line, it does not need to be wrapped in parantheses. However, if any of the 3 clauses are more complex than a single identifier, they should be wrapped in parantheses.
 
-Sometimes a developer will choose to line split an expression despite it being able to all fit on one line that is <=90 characters wide. That is perfectly acceptable, though you may notice in the below examples the single line form can be more readable.
+Sometimes a developer will choose to line split an expression despite it being able to all fit on one line that is <=90 characters wide. That is perfectly acceptable, though you may notice in the below examples the single line form can be more readable. There is "wiggle" room allowed by the above rules. This is intentional, and allows developers to choose a more compact or a more spaced out expression.
 
 **Group**: `spacing`
 
@@ -228,33 +231,30 @@ Complex example that fits on one line:
 Int complex_value = w - x + (y * (z / (foo % bar)))
 ```
 
-Same example with line splitting:
+Same example with as much line splitting as permissible:
 
 ```wdl
-Int complex_value = w
+Int complex_value
+    = w
     - x
     + (
         y
         * (
             z
             / (
-                foo % bar
+                foo
+                % bar
             )
         )
     )
 ```
 
-Another complex example:
+Another complex example written as compactly as permissible:
 
 ```wdl
 Boolean complicated_logic = (
     if !(
-        a
-        && b
-        || c
-        && (
-            !d || !e
-        )
+        a && b || c && (!d || !e)
         == (
             if foobar
             then gak
@@ -264,6 +264,36 @@ Boolean complicated_logic = (
     then "wow"
     else "WOWOWOW"
 )
+```
+
+The same expression with as many newlines inserted as permissible:
+
+```wdl
+Boolean complicated_logic
+    = (
+        if
+            !(
+                a
+                && b
+                || c
+                (
+                    !d
+                    || !e
+                )
+                == (
+                    if
+                        foobar
+                    then
+                        gak
+                    else
+                        caz
+                )
+            )
+        then
+            "wow"
+        else
+            "WOWOWOW"
+    )
 ```
 
 ### `input_spacing`
--- a/lint-rules.md
+++ b/lint-rules.md
@@ -415,7 +415,7 @@ Bad:
     runtime {
         memory: '4 GB'
         disks: '~{disk_size_gb} GB'
-        container: "docker://ghcr.io/stjudecloud/util@sha256:c0583fe91d3e71fcfba58e2a57beb3420c7e907efd601f672fb5968086cd9acb"  # tag: 1.3.0
+        container: 'docker://ghcr.io/stjudecloud/util@sha256:c0583fe91d3e71fcfba58e2a57beb3420c7e907efd601f672fb5968086cd9acb'  # tag: 1.3.0
         maxRetries: 1
     }
 ```
@@ -948,7 +948,7 @@ task greet {
     runtime {
         memory: "4 GB"
         disks: "10 GB"
-        container: "docker://ghcr.io/stjudecloud/util:1.3.0"
+        container: "docker://ghcr.io/stjudecloud/util@sha256:c0583fe91d3e71fcfba58e2a57beb3420c7e907efd601f672fb5968086cd9acb"  # tag: 1.3.0
         maxRetries: 1
     }
 }
@@ -1022,7 +1022,7 @@ task greet {
     runtime {
         memory: "4 GB"
         disks: "10 GB"
-        container: "docker://ghcr.io/stjudecloud/util:1.3.0"
+        container: "docker://ghcr.io/stjudecloud/util@sha256:c0583fe91d3e71fcfba58e2a57beb3420c7e907efd601f672fb5968086cd9acb"  # tag: 1.3.0
         maxRetries: 1
     }
 }
@@ -1056,7 +1056,7 @@ task greet_anonymous {
     runtime {
         memory: "4 GB"
         disks: "10 GB"
-        container: "docker://ghcr.io/stjudecloud/util:1.3.0"
+        container: "docker://ghcr.io/stjudecloud/util@sha256:c0583fe91d3e71fcfba58e2a57beb3420c7e907efd601f672fb5968086cd9acb"  # tag: 1.3.0
         maxRetries: 1
     }
 }
--- a/lint-rules.md
+++ b/lint-rules.md
@@ -827,7 +827,7 @@ Bad:
 
 ### `inpersistent_container`  && `persistent_container_not_tagged`
 
-All tasks should run in a persistent container. This ensures reproducibility across time and environments. `wdl-grammar` and `wdl-ast` will look for a `sha sum` tag in your `container` declarations and warn if one is missing. A `sha` digest will always point to the exact same image, whereas tags are mutable. This mutability makes even versioned tags problematic when we want to ensure reproducibility.
+All tasks should run in a persistent container. This ensures reproducibility across time and environments. `wdl-grammar` and `wdl-ast` will look for a `:SHASUM` tag in your `container` declarations and warn if one is missing. A `sha` digest will always point to the exact same image, whereas tags are mutable. This mutability makes even versioned tags problematic when we want to ensure reproducibility.
 
 While the confidence in persistence gained by using `sha` digests for pulling container images is valuable, it comes at a cost. Which is a lack of human readability. So we enforce preceeding `container` entries with a comment which gives a human readable name to the image being pulled. It is imperative that the `sha` digest and the human readable tag are kept in sync. It's extra overhead compared to just using a tag directly, but we consider it well worth it.
 
--- a/lint-rules.md
+++ b/lint-rules.md
@@ -1,0 +1,1266 @@
+# WDL Lint Rules
+
+The below rules will all be added as lints to `wdl-grammar` and `wdl-ast` (if they aren't already). Most rules will be assigned a "group" that they belong to. The intention of having these groups is twofold. One, it will allow an easy method for disabling multiple related rules at the same time. Two, it will make the documentation easier to organize and search. We're still refining what these final groups will be, so not every rule in this document will be assigned one yet. Some rules may never be assigned as a group if they are truly stand-alone. The current groups are: `spacing`, `naming`, `sorting`, `completeness`, `container`, and `deprecated`.
+
+This document only concerns what we are calling "lint warnings". Lint warnings are distinct from what we call "validation errors". A validation error results when the specification (v1.x at time of writing) is not followed. These errors render your WDL document invalid. On the other hand, lint warnings are just our (the writer's of `wdl-grammar` and `wdl-ast`) opinion on what "good" WDL should look like. They are matters of style, readability, consistency, or for encouraging proper documentation. Your document may be littered with lint warnings, but if there aren't any validation errors, then your document is valid WDL.
+
+We are aiming to automate as much as we can to make adherence to these lint rules as easy as possible. The "auto-formatter" will be packaged as part of our [`sprocket` tool](https://github.com/stjude-rust-labs/sprocket). Auto-formatting has not yet been implemented as this project is still early in development, but while reading this document, please assume that anything easily automated (and perhaps complex to do manually) will eventually be handled by the formatter built into `sprocket`.
+
+Some of these lint warnings we will admit are arbitrary. There are times where several different methods for styling a document could be considered equally appealing. The important part is picking one such style and adhering to it consistenly across a document or codebase. In such cases, this document will only present one option. In the future, that singular option may just be the default, with several more-or-less equivalent styles presented as alternative options available through configuration. If you feel we are picking the wrong defaults, please let us know what you think would be better!
+
+At the time of writing (January 2024), we consider all rules herein to be up for debate. Please open an issue on this repository if you would like to suggest changes to this document.
+
+## Rules
+
+### `version_declaration_placement`
+
+The WDL version declaration is required for any v1 WDL document to be parsed. A missing version declaration will result in a validation error. However an incorrect placement of the decleration will be considered a lint warning. The version decleration should be the very first line in the document, unless there are header comments. In which case, there should be a blank line between the header and the version declaration. There should _always_ be a blank line following the version declaration.
+
+**Group**: `spacing`
+
+#### Example
+
+Good:
+
+```wdl
+version 1.1
+
+...
+```
+
+Also good:
+
+```wdl
+## [Homepage](https://example.com/)
+
+version 1.0
+
+...
+```
+
+### `import_placement`
+
+All import statements should follow the WDL version declaration (with one empty line between the version and the first import statement).
+
+**Group**: `spacing`
+
+#### Example
+
+Good:
+
+```wdl
+version 1.1
+
+import "../../tools/fastqc.wdl" as fastqc_tasks
+import "../../tools/fq.wdl"
+import "../../tools/kraken2.wdl"
+import "../../tools/md5sum.wdl"
+import "../../tools/mosdepth.wdl"
+import "../../tools/multiqc.wdl" as multiqc_tasks
+import "../../tools/ngsderive.wdl"
+import "../../tools/picard.wdl"
+import "../../tools/qualimap.wdl"
+import "../../tools/samtools.wdl"
+import "../../tools/util.wdl"
+import "./markdups-post.wdl" as markdups_post_wf
+```
+
+### `import_sort`
+
+Import statements should be sorted by the lexicographical ordering (GNU `sort` with `LC_COLLATE=C`) of each line. No extra white space is allowed between symbols or lines.
+
+**Group**: `sorting`
+
+#### Example
+
+Good:
+
+```wdl
+import "../../tools/fastqc.wdl" as fastqc_tasks
+import "../../tools/fq.wdl"
+import "../../tools/kraken2.wdl"
+import "../../tools/md5sum.wdl"
+import "../../tools/mosdepth.wdl"
+import "../../tools/multiqc.wdl" as multiqc_tasks
+import "../../tools/ngsderive.wdl"
+import "../../tools/picard.wdl"
+import "../../tools/qualimap.wdl"
+import "../../tools/samtools.wdl"
+import "../../tools/util.wdl"
+import "./markdups-post.wdl" as markdups_post_wf
+```
+
+### `blanks_between_elements` && `no_blanks_between_elements`
+
+There should be a blank line between each WDL element at the root indentation level (such as the import block and any task/workflow definitions) and between sections of a WDL task or workflow. Never have a blank line when indentation levels are changing (such as between the opening of a workflow definition and the meta section). There should also never be blanks _within_ a meta, parameter meta, input, output, or runtime section. See example for a complete WDL document with proper spacing between elements. Note the blank lines between meta, parameter meta, input, the first call or first private declaration, output, and runtime for the example task. The blank line between the workflow definition and the task definition is also important.
+
+**Group**: `spacing`
+
+#### Example
+
+Good:
+
+```wdl
+version 1.1
+
+import "../../tools/unused-example.wdl" as unused_example
+
+workflow hello_world {
+    meta {
+        description: "Greets the user of the workflow."
+        output: {
+            statement: "The WDL generated statement for the user.",
+        }
+    }
+
+    parameter_meta {
+        name: "Name to greet"
+        greeting: "Phrase to use while greeting `name`"
+    }
+
+    input {
+        String name = "World"
+        String greeting = "Hello"
+    }
+
+    call greet { input:
+        name = name,
+        greeting = greeting,
+    }
+
+    output {
+        String statement = greet.statement
+    }
+}
+
+task greet {
+    meta {
+        description: "Generates a statement from a name and a greeting"
+        output: {
+            statement: "The result of combining the input name and greeting",
+        }
+    }
+
+    parameter_meta {
+        name: "Name to greet"
+        greeting: "Phrase to use while greeting `name`"
+    }
+
+    input {
+        String name
+        String greeting
+    }
+
+    String statement = "~{greeting}, ~{name}"
+
+    command <<<
+        echo "~{statement}" > statement.txt
+    >>>
+
+    output {
+        String statement = read_string("statement.txt")
+    }
+
+    runtime {
+        memory: "4 GB"
+        disks: "10 GB"
+        container: 'docker://ghcr.io/stjudecloud/util:1.3.0'
+        maxRetries: 1
+    }
+}
+
+```
+
+### `inconsistent_newlines`
+
+Files should not mix `\n` and `\r\n` line breaks. Pick one and use it consistently in your project. Prefer unix line-endings (`\n`) when possible.
+
+### `line_width`
+
+WDL lines should be less than 90 characters wide whenever possible. Exceptions would be long strings that WDL doesn't allow to be broken up. This restriction applies to embedded code in the `command` block as well.
+
+### `expression_spacing`
+
+The following rules lead to consistently readable expressions. Note that the below rules only apply to places in WDL code where an **expression** is being evaluated. There are separate rules for whitespace in other code locations.
+
+The following tokens should be **surrounded** by whitespace when used as an infix in an expression:
+
+* `=`
+* `==`
+* `!=`
+* `&&`
+* `||`
+* `<`
+* `<=`
+* `>`
+* `>=`
+* `+`
+* `-`
+* `*`
+* `/`
+* `%`
+
+The following tokens should **not** be **followed** by whitespace (including newlines) when used as a prefix in an expression:
+
+* `+`
+* `-`
+* `!`
+
+Opening brackets (either `(` or `[`) should **not** be **followed** by a **space** but **may** be **followed** by a **newline**.
+Closing brackets (either `)` or `]`) should **not** be **preceeded** by whitespace **unless** that whitespace is **indentation**.
+
+Sometimes a long expression will exceed the maximum line width. In these cases, one or more linebreaks should be introduced. Line continuations should be indented one more level than the beginning of the expression. If bracketed content (things between `()` or `[]`) must be split onto multiple lines, a newline should follow the opening bracket, the contents should be indented an additional level, then the closing bracket should be de-indented to match the indentation of the opening bracket.
+
+If you are line splitting an expression on an infix operator, the operator should be on the continued line.
+
+If you are using the `if...then...else` construct as part of your expression and it needs to be line split, the entire construct should be wrapped in parantheses (`()`). The opening paranthesis should be immediately followed by a newline. `if`, `then`, and `else` should all start a line one more level of indentation than the wrapping paratheses. The closing paranthesis should be on the same level of indentation as the opening paranthesis.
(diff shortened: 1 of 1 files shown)


adthrasher (approved):

lint-rules.md
```diff
@@ -0,0 +1,1266 @@
...
+    then "wow"
+    else "WOWOWOW"
+)
+```
+
+### `input_spacing`
+
+When making calls from a workflow, it is more readable and easier to edit if the supplied inputs are each on their own line. This does inflate the line count of a WDL dcoument, but it is worth it for the consistent readability. An exception _can_ be made (but does not have to be made), for calls with only a single parameter. In those cases, it is permissable to keep the input on the same line as the call.
+
+When newline separating inputs to a call, always follow each input with a trailing comma. See examples below.
```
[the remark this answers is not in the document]
a-frantz: Can you elaborate? I have reasons for preferring it, but I'm not above having my mind changed on the matter.

My reasons for being pro-trailing-comma
* Cleaner diffs
    * New values can be appended to a list "cleanly"
    * No trailing comma, and the former-last value will appear "changed" in the diff as a comma must be appended to an otherwise unchanging line
* In my experience, trailing commas are easier to stay compliant with when changing an existing list. I've definitely been burned by compilers that complained about a missing comma when I add to a list. Or conversely in languages where trailing commas are disbarred (JSON) been burned by deleting the last element and forgetting to clean up the now-trailing-comma that must be deleted. IMO it's just "easier" to always have a comma at the end. Very subjective point here. Not my strongest argument, but I stand by it lol.

lint-rules.md
```diff
@@ -0,0 +1,1266 @@
...
+
+### `no_newline_eof`
+
+End the file with a newline.
+
+**Group**: `spacing`
+
+### `comment_whitespace`
+
+Comments on the same line as code should have 2 spaces before the `#` and one space before the comment. Comments on their own line should match the indentation level around them and have one space between the `#` and the comment. Keep in mind that even comments must be kept below the 90 character width limit.
```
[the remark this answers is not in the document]
a-frantz: I have a strong dislike for that comment style. I have to admit, it does _look_ a lot cleaner. But I just find manually aligning comments as I write code to be a huge pain. I'm more concerned with my experience as a developer writing comments than I am with someone else reading my comments. I guess I'm selfish in that regard.

More importantly, I find the first example (following my rules) perfectly readable. Mainly because of the 'double space' in it providing a bit of a break between code and comment.

Followup Q about your example: Are the 2 comments there related? Or are they separate ideas? _If_ those are meant to be one sentence (or one thought), I would argue both examples are wrong. I would write that "multiline spanning comment" like this:
```
input {
    # This is a long comment with a lot going on. I need
    # more space to write it out.
    Int x
    String name
}
```
_If_ they are separate thoughts, why would they be aligned? I guess it's more visually appealing, but it doesn't strike me are more logical or coherent. If the comments are unrelated, they shouldn't be aligned. Maybe "shouldn't" is strong, but they certainly don't _need_ to be.

lint-rules.md
```diff
@@ -0,0 +1,1266 @@
...
+4) `struct`
+5) `Object`
+6) `Map[*, *]`
+7) `Pair[*, *]`
+8) `String`
+9) `Boolean`
+10) `Float`
+11) `Int`
+
+For ordering of the same compound type (`Array[*]`, `struct`, `Map[*, *]`, `Pair[*, *]`), drop the outermost type (`Array`, `Map`, etc.) and recursively apply above sorting on the _first_ inner type `*`, with ties broken by the _second_ inner type. Continue this pattern as far as possible. Once this ordering is satisfied, it is up to the developer for final order of inputs of the same type.
```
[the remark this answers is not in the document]
a-frantz: Our [Best Practices](https://github.com/stjudecloud/workflows/blob/main/best-practices.md) does have something to say on the matter. However it's not something we can feasibly automate, which is why it's left out of this document.

> * After the input sorting rules in style-guide.md have been applied, follow the below rules for further sorting.
>     * "sample" files come before "reference" files
>     * If present, use_all_cores should be the last Boolean in its block
>     * the ncpu parameter comes before inputs that allocate memory, which come before inputs that allocate disk space
>    * This block of 2-3 inputs should come after all other inputs.

lint-rules.md
```diff
@@ -0,0 +1,1266 @@
...
+    command {
+        echo "Hello, World!"
+    }
+```
+
+### `inpersistent_container`  && `persistent_container_not_tagged`
+
+All tasks should run in a persistent container. This ensures reproducibility across time and environments. `wdl-grammar` and `wdl-ast` will look for an `sha sum` tag in your `container` declarations and warn if one is missing. An `sha` digest will always point to the exact same image, whereas tags are mutable. This mutability makes even versioned tags problematic when we want to ensure reproducibility.
+
+While the confidence in persistence gained by using `sha` digests for pulling container images is valuable, it comes at a cost. Which is a lack of human readability. So we enforce preceeding `container` entries with a comment which gives a human readable name to the image being pulled. It is imperative that the `sha` digest and the human readable tag are kept in sync. It's extra overhead compared to just using a tag directly, but we consider it well worth it.
```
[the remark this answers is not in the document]
a-frantz: I was unaware of this syntax at time of writing, and I would say this is the ideal syntax to enforce. _BUT_, Andrew T looked into it and found out that `singularity` doesn't support this syntax. So we can't use it internally.

I think this has to be one of the rules where we offer alternatives. However, I opted to leave all alternatives out of this first draft. Mainly due to the great length without going over variations. I didn't want to prolong the document.

My vote is update this rule to use the syntax proposed by @adthrasher . It seems ideal in all regards to me. However we can't use it, for now at least, so when it comes to implement this rule in code, we'll need to add some alternatives available to toggle. But the above stays the default, us as users would need to configure to something else.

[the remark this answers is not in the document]
adthrasher: This issue (https://github.com/containers/buildah/issues/1407) provides some discussion of the `:tag@digest` format. Particularly this comment

```
Earlier versions of Docker, including 1.13.1, interpret this as "ask for the image with the specified tag, then fail if the manifest does not match the specified digest". That's fairly brittle but ultimately plausible (where the tag ~documents the intent and the digest ensure immutability and prevents moving the tag maliciously).

Later versions of Docker interpret this as "ask for the image with the specified digest, ignore the tag completely", which is, in my view, mostly pointless and can be actively misleading. If the tag is ignored in the code, the users' intent can just as well be noted in a comment after the command instead of inside the image specification, and that would be more accurate WRT what is actually happening.
```

Assuming that is accurate and docker is ignoring the `tag`, then I would tend to agree with the author that putting the `tag` in a comment is a better solution (and it aligns with your format).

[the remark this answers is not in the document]
a-frantz: The "earlier" interpretation is what I assumed, so thank you for proving that assumption wrong. I would say the "later" interpretation is misleading and I'm surprised that's what Docker moved to. I think this needs some more investigation+discussion before settling on what the "most sane" default is.

I'll elaborate on what I intend to accomplish with this rule then pose a question to all the reviewers.

IMO WDL should strive to be reproducible across time and environments. The same WDL code should produce the same results regardless of where and when it was run. With this understanding, "moving" tags like `latest` should be avoided.

For context, our current `main` for `stjudecloud/workflows` uses BioContainer release tags (not `sha` digests) OR pulls `ghcr.io` images we own. To my understanding, the BioContainer tags are static and are never updated in place (rendering an `sha` digest redundant). We are in control over whether our `ghcr.io` images are overwritten, and we don't do that (though our CI _would_ allow it if we deviated from our standard practice). I think this is a perfectly acceptable set up.

The issue is that while writing a linter, the linter can't know the pulled tags are static or "safe". Which is why I argue the default for the linter should include an `sha` digest. However this also poses a problem, in that plain `sha` digests don't convey what version of a tool/image is being pulled. Hence the documents rule of also specifying a human readable version in a comment.

I understand that this rule, requiring `sha` digests in pulled container images, might be the most controversial rule in the document. So I wonder if we should list out what we consider "acceptable" alternatives for this rule. I'm hesitant to list alternate configurations for this rule, because it might obligate us (me) to list alternatives to other rules. I'd rather not do that, as the word count on this document is high enough as it is.

So, does this topic have enough nuance+import that it warrants a set of alternative configurations? Should any other rules get alternatives as well?

lint-rules.md
```diff
@@ -0,0 +1,1281 @@
...
+
+### `disallowed_output_name`
+
+Any output name matching these regular expressions will be flagged: `/^[oO]ut[A-Z_]/`, `/^output/i` or `/^..?$/`. It is redundant and needlessly verbose to use an output's name to specify that it is an output. Output names should be short yet descriptive. Prefixing a name with `out` or `output` adds length to the name without adding clarity or context. Additionally, names with only 2 characters can lead to confusion and obfuscates the content of an output. Output names should be at least 3 characters long.
+
+**Group**: `naming`
+
+#### Example
+
+All the below input names would be flagged:
```
claymcleod, line 795:
```suggestion
All the below output names would be flagged:
```

lint-rules.md
```diff
@@ -0,0 +1,1281 @@
...
+```wdl
+    input {
+        File outBam  # This name will be flagged twice, once for the 'out' prefix, and once for being camalCase when it should be snake_case
+        File out_bam
+        File output_gtf
+        String af
+    }
+```
+
+### `no_curly_commands`
```
claymcleod, line 806: What is the group for this one?

lint-rules.md
```diff
@@ -0,0 +1,1281 @@
...
+
+Bad:
+
+```wdl
+    command {
+        echo "Hello, World!"
+    }
+```
+
+### `inpersistent_container`  && `persistent_container_not_tagged`
```
claymcleod, line 828: First of all, if we keep this rule, I would use the word "immutable" instead of "persistent".

lint-rules.md
```diff
@@ -0,0 +1,1281 @@
...
+
+All tasks should run in a persistent container. This ensures reproducibility across time and environments. `wdl-grammar` and `wdl-ast` will look for a `sha sum` tag in your `container` declarations and warn if one is missing. A `sha` digest will always point to the exact same image, whereas tags are mutable. This mutability makes even versioned tags problematic when we want to ensure reproducibility.
```
claymcleod, line 830:
```suggestion
All tasks should run in a persistent container. This ensures reproducibility across time and environments. `wdl-grammar` and `wdl-ast` will look for a `:SHASUM` tag in your `container` declarations and warn if one is missing. A `sha` digest will always point to the exact same image, whereas tags are mutable. This mutability makes even versioned tags problematic when we want to ensure reproducibility.
```

lint-rules.md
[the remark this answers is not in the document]
claymcleod: I will go on record as saying I'm not a fan of this rule. I think there are very valid reasons to use non-SHASUM tags as long as you are aware of what you are doing. I also think the specifics of this rule make it hard to reasonable about. But, if you two are of the opinion we should keep it, then let's keep it.

lint-rules.md
```diff
@@ -0,0 +1,1281 @@
...
+
+While the confidence in persistence gained by using `sha` digests for pulling container images is valuable, it comes at a cost. Which is a lack of human readability. So we enforce preceeding `container` entries with a comment which gives a human readable name to the image being pulled. It is imperative that the `sha` digest and the human readable tag are kept in sync. It's extra overhead compared to just using a tag directly, but we consider it well worth it.
```
claymcleod, line 832: If we keep the rule, it needs to be word-smithed. For example, "Which is a lack of human readability." is a sentence fragment.

lint-rules.md
```diff
@@ -0,0 +1,1281 @@
...
+        disks: "10 GB"
+        container: "docker://ghcr.io/stjudecloud/util:1.3.0"
+        maxRetries: 1
+    }
+}
+```
+
+Although the above solves the lint warning issue, it's not very well written. It was a contrived example to illustrate one way to silence the lint warning (wrap the coercion in an `if (defined(<>))` block). We will leave it as an exercise for the reader what the "Best" option might be. (there are at least 2 ways: either rewrite the `greet` task to accept an optional `name`, or use the standard library's `select_first()` function with a default name. Either implementation will eliminate the lint warning).
+
+### `runtime_failable_nonempty_coercion`
```
claymcleod, line 1068: Personally, I would prefer "fallible" over "failable".

lint-rules.md
```diff
@@ -0,0 +1,1281 @@
...
+
+An import which is not used anywhere in the document. While relatively harmless, these unused imports can be confusing to readers and maintainers alike.
+
+### `unused_declaration`
+
+Nothing references a declaration. Much like unused imports, these can be confusing to readers and maintainers.
+
+### `select_nonoptional_array`
+
+It  is unnecessary to use `select_first()` or `select_all()` on an array which is non-optional. This usage is confusing.
```
claymcleod, line 1146:
```suggestion
It is unnecessary to use `select_first()` or `select_all()` on an array which is non-optional. This usage is confusing.
```

claymcleod (approved):

lint-rules.md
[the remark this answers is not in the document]
a-frantz: @claymcleod 

> I'm in favor of what the spec says FWIW.

I looked over the spec, and I don't see it making any comment on the style issue me and Andrew are discussing. In my reading, it just outlines how to make a comment/how a comment should be parsed. It does not clarify how to style comments.

Maybe I'm looking in the wrong spot. Can you link or quote?

lint-rules.md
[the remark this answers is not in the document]
adthrasher: I don't feel strongly enough about this, particularly if we support multiple options (as discussed via slack). I agree with you from the perspective of `diff`. However, for reading, I think omitting the trailing comma (as you would in standard written English) is cleaner. When I see a list with a trailing comma, I want to verify that something wasn't omitted from the list accidentally. I also think that a trailing comma encourages appending, which is not always the right choice when adding to a list.

lint-rules.md
```diff
@@ -0,0 +1,1281 @@
...
+
+# This comment has an incorrect level of indentation
+    Int z = x + y# NO spaces between code and comment is hard to read.
+```
+
+### `double_quotes`
+
+All quotes should be double quotes.
+
+**Group**: `naming`
```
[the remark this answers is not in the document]
a-frantz: This got lumped under "naming" because I went about the group concept by asking "what would likely be disabled at the same time?" I think double quotes and snake_case etc. rules make sense together from that perspective.

lint-rules.md
a-frantz: No group ATM. As stated in another comment, my approach for "groups" is "what rules would be turned off together?" I don't think this has any logical group it would fit into by that criteria. IMO not every rule needs a group. A "group-less" rule is just meant to be toggled on/off by itself.

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
+# WDL Lint Rules
```
rickymagner, line 1: ## Overall Remarks

- (**Important**) Summarize the list of lints with "one-liner" summaries for someone to check over at a glance, with more details in the full docs linked to from the summaries. This will help get feedback from more people, and I'd be happy to broadcast a bit more but can't imagine many folks would want to spend 30-60 minutes reading through the entire doc. They might spend 10 minutes however.
- Include in the doc a method for submitting feedback, whether it be via issues, commenting on this PR, etc. I think issues make more sense in the future since this PR is closed and can get quickly cluttered in the future.
- "Groups" for the lint rules could act like tags: you can be a part of multiple groups, e.g. `import_sort` is part of `sorting` and `import` ? Though I'm not sure you'd ever want to turn off all `import` rules for some reason. But it's worth considering at this stage if you ever foresee potential overlaps.
- Maybe declare at the top that "a blank line" means "exactly one blank line" throughout the doc so it's clear everywhere it doesn't mean "at least one blank line."
- It might also make sense to reorganize the doc to have lint group subheaders? Maybe someone knows they have strong feelings about whitespace but don't care about deprecation lints and can just jump to that section for example when reviewing. It can also make it easier for folks to debug which lints they want for their project if they're having issues with a particular flavor of warnings.

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+
+Some of these lint warnings are arbitrary. There are times where several different methods for styling a document could be considered equally appealing. The important part is picking one such style and adhering to it consistenly across a document or codebase. In such cases, this document will only present one option. In the future, that singular option may just be the default, with several more-or-less equivalent styles presented as alternative options available through configuration. If you feel we are picking the wrong defaults, please let us know what you think would be better!
+
+At the time of writing (January 2024), we consider all rules herein to be up for debate. Please open an issue on this repository if you would like to suggest changes to this document.
+
+## Rules
+
+### `version_declaration_placement`
```
rickymagner, line 15: The `version` line should be followed by *exactly* one blank line (this rule should enforce this rather than the `import` rules that follow since there might not be any imports)

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+```wdl
+version 1.1
+
+import "../../tools/fastqc.wdl" as fastqc_tasks
+import "../../tools/fq.wdl"
+import "./markdups-post.wdl" as markdups_post_wf
+import "https://raw.githubusercontent.com/stjude/seaseq/3.0/workflows/tasks/seaseq_util.wdl" as seaseq_util
+```
+
+### `blanks_between_elements`
```
rickymagner, line 64: - (Strong opinion) When `call`ing a task in a workflow, you should format it so that the `input:` line is indented on the next line alone, and then indent again for the input assignments, each on one line. So

```
   call greet { input:
        name = name,
        greeting = greeting,
    }
```
should be
```
call greet { 
    input:
        name = name,
        greeting = greeting,
    }
```
- This case should maybe be a separate lint about styling task calls. (See also thoughts on `input_spacing` below)
- Also (maybe) consider removing whitespace between input assignments like in Python function calling with keyword assignments, e.g. `name=name,` etc. (though this conflicts with `expression_spacing` later... maybe a separate lint for spacing within call blocks? But I don't care as much about this detail honestly)

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+
+### `inconsistent_newlines`
+
+Files should not mix `\n` and `\r\n` line breaks. Pick one and use it consistently in your project.
+
+### `line_width`
+
+WDL lines should be less than or equal to 90 characters wide whenever possible. This line width restriction applies to embedded code in the `command` block. Exceptions would be long strings that WDL doesn't allow to be broken up within the meta and parameter meta sections. Another exception is `container` lines inside the `runtime` block of a task. (See the rules `mutable_container`  && `immutable_container_not_tagged` for more information about permitted `container` lines.)
+
+### `expression_spacing`
```
rickymagner, line 153: - There should be a compact vs more newlines way to toggle the behavior of this lint to enforce "only one way to do something" for repos to automate... Maybe two separate lints that are incompatible?

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+                )
+            )
+        then
+            "wow"
+        else
+            "WOWOWOW"
+    )
+```
+
+### `input_spacing`
```
rickymagner, line 270: - Typo: "dcoument" to "document"
- Disagree with this lint generally; I prefer having `input:` on its own line, which helps emphasize that this is a call to a task, and also allows for WDL to evolve to allow other types of blocks to pass to the task (maybe `meta` info in the future? something else?) so that this lint would "scale" more naturally

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+        markdups.duplicate_marked_bam_index,
+        "undefined",
+    ]),
+    coverage_beds,
+    coverage_labels = parse_input.labels,
+    prefix = post_subsample_prefix + ".MarkDuplicates",
+}
+```
+
+### `snake_case`
```
rickymagner, line 320: - (Strong opinion) Disagree with this lint generally. Workflows and tasks should be PascalCase (e.g. `MyWorkflow` and `MyTask`), but variable names should be `snake_case`. The same as Python classes vs variables. I agree with structs as PascalCase as well.

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+
+    runtime {
+        memory: '4 GB'
+        disks: '~{disk_size_gb} GB'
+        container: 'docker://ghcr.io/stjudecloud/util@sha256:c0583fe91d3e71fcfba58e2a57beb3420c7e907efd601f672fb5968086cd9acb'  # tag: 1.3.0
+        maxRetries: 1
+    }
+```
+
+### `trailing_comma`
```
rickymagner, line 423: - Not sure I like this generally. I think it looks better to not have the trailing comma for lists, but understand I could just turn off the lint if I don't like it

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+        File star_log = prefix + ".Log.final.out"
+        File star_bam = prefix + ".Aligned.out.bam"
+        File star_junctions = prefix + ".SJ.out.tab"
+        File? star_chimeric_junctions = prefix + ".Chimeric.out.junction"
+    }
+    ...
+}
+```
+
+### `nonmatching_parameter_meta` && `input_not_sorted`
```
rickymagner, line 591: - (Strong opinion) I think this lint (`input_not_sorted`) should sort not by WDL type in the `parameter_meta`, but by the actual order they appear in the task declaration. I often group inputs together based on what fits together (e.g. `String sample_name` is next to `File input_vcf` for a task that extracts that sample from the vcf, etc.) so rearranging by type alone would break the scientific cohesion of the user-based groupings

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+        Int chim_multimap_n_max = 0
+        Int chim_multimap_score_range = 1
+        Int chim_non_chim_score_drop_min = 20
+        Int twopass1_reads_n = -1
+        Int ncpu = 8
+        Int modify_disk_size_gb = 0
+    }
+```
+
+### `disallowed_input_name`
```
rickymagner, line 767: - (Strong opinion) The second regex is fine (I'll have to change a few of my WDLs!) but the first is way too strict. What if I want an input named `Int indel_length_cutoff`? It would not be allowed by this lint!

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+
+```wdl
+    input {
+        File inBam  # This name will be flagged twice, once for the 'in' prefix, and once for being camalCase when it should be snake_case
+        File in_bam
+        File input_gtf
+    }
+```
+
+### `disallowed_output_name`
```
rickymagner, line 785: - (Strong opinion) Similar to previous: this is way too strict. What if my output is `File outer_join_table`? 
- It's also asymmetric to disallow 2-letter names here but allow for inputs. Maybe a separate lint for "no variable names under 3 chars"?

lint-rules.md
```diff
@@ -0,0 +1,1280 @@
...
+```wdl
+    runtime {
+        memory: "4 GB"
+        disks: "10 GB"
+        container: "docker://quay.io/biocontainers/htseq@sha256:04309f74909f7e48bc797ee5faa4e4388d7f581890c092a455d15bbcf5f6c537"  # tag: 2.0.4
+        maxRetries: 1
+    }
+```
+
+### `file_coercion`
```
rickymagner, line 860: - I'm not sure whether I'd just shut this lint off or if it should be modified, but its current form disallows e.g. setting defaults that point to public cloud locations, like a reference file input, etc. which is a common use case for the coercion

rickymagner (commented):
Here are some thoughts after reading the [lint-rules](https://github.com/stjude-rust-labs/wdl/blob/main/lint-rules.md) guide. I tried to highlight some "strong opinions" (which I was surprised I had), but of course it's all just opinions. I also have some overall remarks about the doc and ideas included first.

Thanks for all the hard work on this! Hoping it gains some traction.

lint-rules.md
a-frantz: > *  Summarize the list of lints with "one-liner" summaries for someone to check over at a glance, with more details in the full docs linked to from the summaries.

Great idea! Will put this summary on my TODO list :)

> * Include in the doc a method for submitting feedback

Good point, this was an oversight not including our "preferred method of feedback". Will converse with the team and decide what would be best for us to track and respond to.

> * "Groups" for the lint rules could act like tags

Good idea to consider. The term "tags" is maybe more quickly understood than "groups"; so even if we don't want to start assigning multiple to a rule (now) it could make the intent clearer.

> * Maybe declare at the top that "a blank line" means "exactly one blank line" throughout the doc

👍 

> * It might also make sense to reorganize the doc to have lint group subheaders?

Another good suggestion! I think this would be a pretty straight-forward improvement.

We really appreciate your input.

lint-rules.md
a-frantz: For communication purposes, let's refer to the style currently proposed by `lint-rules.md` as "inline-input" and the style you proposed as "indented-input". (Accepting alternative names)

I've seen indented-input in several repos. Do you think this is the style the community is coalescing around?

I strongly prefer inline-input. I think good universal rules are that opening brackets should be on the same indent level as their closing bracket, and that code should never "skip" an indent level by incrementing or decrementing by greater than 1 level at a time. Both of these rules are broken by indented-input.

Why do you prefer indented-input? 

-------------------

> * Also (maybe) consider removing whitespace between input assignments like in Python function calling with keyword assignments, e.g. `name=name`, etc. (though this conflicts with `expression_spacing` later... 

First, I find the whitespace padding on assignments easier to read. I wouldn't throw a hissy fit over losing that whitespace though if my opinion is the minority.

Second, the lints are written such that the `input_spacing` lint and the `expression_spacing` lints can "contradict" (though that may prove confusing). By that I mean we _could_ have a lint active for "no spaces around input assignments (`=`)" **and also** have a lint active for "whitespace padding around expression assignments (`=`)". They are different contexts. However my gut is that they should treat whitespace the same way around _all_ equal signs for simplicity.

lint-rules.md
a-frantz: Interesting that you say `same as Python classes vs variables`. I view workflows and tasks to be analogous to Python **functions** not classes (structs would be most similar to classes). And of course Python functions are snake_cased like variables.

I think all casing choices are more-or-less equivalent. I propose leaving it up to some democratic method to determine the `sprocket` default.

lint-rules.md
rickymagner: I agree with your second point. I've just gotten into the habit of no surrounding whitespace for `call`s since Python has the same weird dichotomy e.g. `x = 4` on its own but `my_func(x=4, y=7)` for function calls. WDL can be different and more consistent with whitespace everywhere.

---
I don't have a sense of the "larger community" and just have a biased view of the WDLs I've seen around Broad. But I never saw inline-input style until I read this lint guide. Here's an argument in favor of indented-input. What is the purpose of the syntax `input:` at all instead of e.g.
```
call MyTask {
    file=file,
    var=var,
}
```
Currently this syntax is conceptually OK, but for whatever reason the `input:` is required. This suggests there could be a future where other blocks of variables are fed to a task call with different headers, e.g.
```
File my_task_log_file

call MyTask {
    input:
        file=file,
        var=var,
    debug:
        log_file=my_task_log_file
}
```
In this case, using inline-input has no natural way to extend to the new `debug:` situation, and would look very ugly. Perhaps it's worth asking some WDL developers what the intent behind including `input:` in `call` blocks at all is to help determine how to style it.

lint-rules.md
rickymagner: Yeah, I think of a `task` sort of like an object that has different attributes: a command block, runtime parameters, meta data, and also takes input/output info of course. I can see why it could feel either like a function or like an object.

lint-rules.md
a-frantz: Wonderful point! You've convinced me to loosen this rule.

lint-rules.md
a-frantz: Good ideas, good ideas! Will revisit this.

lint-rules.md
a-frantz: This is another great point! Thank you! When I wrote this, I really only meant to exclude **local** file paths. But you are correct, as written, I've also excluded files which should be available to any environment with an internet connection. I'll have to workshop the language to be more clear on this issue.

lint-rules.md
rickymagner: I also might've misunderstood this lint, but my understanding was that it would take a task input and sort it based purely on the WDL type in the hierarchy described in the doc. So if my inputs looked like this at first:
```
File vcf
String sample_name

File bam
String read_group

File debug
File interval_list
```
the linted version would look like:
```
File bam
File debug
File interval_list
File vcf
String read_group
String sample_name
```
even though conceptually the "sample_name" and "vcf" inputs should be "near" each other because the "sample_name" references a sample in the vcf. 

In case I misunderstood the lint, here's what the ideal version would look like to me: `nonmatching_parameter_meta` makes sure that the `parameter_meta` input order matches the user-defined on in the task inputs, but would probably turn off `input_not_sorted` because I usually cluster my input based on conceptually similarity, regardless of WDL type/alphabetical order.

lint-rules.md
rickymagner: That makes sense, though might be tricky to distinguish "local" paths from "global" paths when verifying. If there's a good way to do that, I think it could make a good lint as `disallow_local_file_paths` or something.

lint-rules.md
adthrasher: I think there are some list-like structures that `miniwdl check` errors on if a trailing comma is included.

lint-rules.md
a-frantz: @adthrasher good points. I also think its reasonable for a user to want an alphabetical sort, and it's probably worth it to include that as an option. Between the 3 mentioned choices (our partially-opinionated-partially-arbitrary typed sort, alphabetical sort, and unsorted), IDK what's best as a default. [tangent] I don't think we've really settled on an exact philosophy for how we determine defaults.

lint-rules.md
a-frantz: Personally, I think trailing commas look a lot cleaner and are the more visually appealing style. Trailing commas also give me less friction when actually coding. Of course those are highly subjective opinions, and anyone could disagree.

I think this is another case where there is no "best" or "objective" answer. I think the best way to settle this sort of arbitrary debate is something democratic.

Idea:
1. What if we chose some selection of WDL repos (simple list would be the repos Clay collected for `wdl-gauntlet`)
2. checked them for consistency on commas (and any other "arbitrary" lint rules we're deciding on)
3. grouped the repos by "position" on each proposed rule
4. Summed each voting blocks' number of GitHub stars
5. The rule with the most votes becomes the default for `sprocket`!

lint-rules.md
a-frantz: https://github.com/openwdl/wdl/pull/635/files

In the 1.2 WDL release they are going to remove the requirement for the `input:` token. When that time comes, I am _definitely_ voting for deleting all uses of the token and updating the lint rules to match. With that said, I think we should adopt the style which will require the least amount of work when 1.2 is released. That would be the "inline-input" I originally proposed. A simple `sed -i` would be all that's needed to mass migrate to 1.2 style. It would be much more complicated to deal with the indentation changes required by the indented-input.

lint-rules.md
rickymagner: I think some method like this makes sense, though the "samples" may not be independent, e.g. different Broad repos probably influence different Broad repos to follow similar style, etc. Maybe it would make sense to advertise whatever method you decide so other folks can chime in ideas?

lint-rules.md
rickymagner: I definitely completely agree that when it becomes obsolete it should be linted away. I asked in Slack to confirm that this means there definitely will never be additional headers in the `call` blocks, since I could imagine the developers choosing that `input` is optional when it's the only header, but also still support multiple headers in the future where the `input` becomes mandatory to specify it. Given the possibility for this (at the current stage), I still think the indented-input is more flexible to handle these (and am obviously biased since I use that). I wouldn't worry about refactoring it away in the future as a reason to choose either: whatever autofixing lint command that's written shouldn't struggle over a find/replace within `call` blocks vs also fiddling some whitespace. If a "vote" does happen by scraping repos I'd be curious to see which of these styles (or even a third?!) end up being most common in the wild.

a-frantz: > Given the possibility for this (at the current stage), I still think the indented-input is more flexible to handle these (and am obviously biased since I use that).

I don't think it makes sense to choose our style based on a direction the SPEC may or may not take. Especially because that _potential_ change would necessitate new style lints if it ever came to fruition. i.e. if in the future they add non-input blocks to the call, we could just as easily change from inline-input to indented-input to some new 3rd option. Which brings me around to agreeing with your point:
> whatever autofixing lint command that's written shouldn't struggle over a find/replace within call blocks vs also fiddling some whitespace.

So I'll agree the point I made in the above comment wasn't very convincing lol. But the same logic applies to your comment as well.

>  If a "vote" does happen by scraping repos I'd be curious to see which of these styles (or even a third?!) end up being most common in the wild.

I proposed that repo scraping idea, but now that I've thought about it some more, it seems to fall flat. As you (@rickymagner) pointed out, the "sampling" will over-represent certain styles such as the Broad. I don't think what I proposed would lead to anything representative.

I think much less over-engineered way to go about this will be a simple Google poll or something like that which gets shared around. Or maybe even the lowest tech option would be a poll _within_ the openwdl Slack. Regardless, I think it's a bridge we don't need to cross now.

rickymagner: I agree something like a carefully thought out poll on a few lint decisions that gets spread around different groups somehow fairly sounds like a good idea. There's always risk of oversampling either way (I'd definitely advertise around Broad which could lead to oversampling/bias). There's also a possibility for divisive lints to come in non-compatible either-or groups, like an inline vs indented input in calls, etc.

PR #23, opened 2024-04-30
fix: ignore comments when parsing `Expression`.

### LINKED ISSUE #21: [BUG] Error parsing comment in multi-line boolean expression
Here's another bug in parsing, though I admit it's a bit of a silly edge case. If you have a multi-line boolean expression with a comment in the middle for some reason, the parser will not interpret it correctly. Here are the steps to reproduce:

1. Make a file called `MinimalParse.wdl` with the contents:
```
version 1.0

workflow MinimalParse {
    input {
        String? name
    }

    if (defined(name) &&
       # comment here
       defined(name)) {

    }
}
```
2. Try to parse with the `wdl` crate. Here is the snippet of code:
```
// Inside main
    let src = std::env::args().nth(1).expect("missing src");
    let contents = std::fs::read_to_string(src).expect("could not read file contents");

    // Generate the parse tree from the document source.
    let parse_tree = grammar::v1::parse(&contents)
        .expect("Cannot parse")
        .into_tree()
        .expect("Cannot convert to tree");

    // Generate the abstract syntax tree from the parse tree.
    let syntax_result = ast::v1::parse(parse_tree);
```
3. The program panics with output:
```
internal error: entered unreachable code: unknown primary in expression: COMMENT
```
4. Removing the comment line (keeping the multi-line boolean expression) parses just fine. Also everything in the Rust snippet works until the last line where the AST is parsed. 

Both `womtool validate` and cromwell are happy with the original file. I agree it's probably not a good way to write comments in WDL in general, but the error produced here at the very least is not helpful. I suspect this edge case was not considered since the error mentions "unreachable code" but do think this might be legal WDL that should be parsed correctly.

peterhuene:
This commit fixes the incorrect filter when iterating over the pairs provided to the `parse` function for `Expression`.

Fixes #21.

      changes (when appropriate).

### DIFF
--- a/wdl-ast/CHANGELOG.md
+++ b/wdl-ast/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-* Fix ignoring comments in expressions [#23](https://github.com/stjude-rust-labs/wdl/pull/23).
+* Fix ignoring comments in expressions ([#23](https://github.com/stjude-rust-labs/wdl/pull/23)).
 
 ## 0.1.0 — 12-17-2023
 
--- a/wdl-ast/CHANGELOG.md
+++ b/wdl-ast/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Fix ignoring comments in expressions [#23](https://github.com/stjude-rust-labs/wdl/pull/23).
+
 ## 0.1.0 — 12-17-2023
 
 ### Added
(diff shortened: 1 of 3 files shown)


wdl-ast/CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+* Fix ignoring comments in expressions [#23](https://github.com/stjude-rust-labs/wdl/pull/23).
```
claymcleod, line 10:
```suggestion
* Fix ignoring comments in expressions ([#23](https://github.com/stjude-rust-labs/wdl/pull/23)).
```

claymcleod (approved):
LGTM

PR #24, opened 2024-05-02
[MIRROR]  feat: add check for proper snake_case usage 

claymcleod:
This is a mirror of #13, as I do not have access to push to @simojoe's fork.

Closes #13.


claymcleod:
Closes #13.

PR #30, opened 2024-05-08
feat: implement a WDL lexer using `logos`.

peterhuene:
This commit introduces an `experimental` module as a placeholder for a new parser implementation, feature-gated by the `experimental` feature.

To start, a lexer implementation based on `logos` has been added.

The lexer may have several different token types depending on context, mainly to support interpolation in string literals and commands.

The implementation should be enough to lex a 1.x WDL document.

The parser implementation that will make use of the lexer is forthcoming.

      good, articulated reason as to why there shouldn't be more than one).

### DIFF
--- a/wdl-grammar/src/experimental/lexer.rs
+++ b/wdl-grammar/src/experimental/lexer.rs
@@ -218,6 +218,10 @@ version 1.1",
     #[test]
     fn test_version_draft3() {
         use VersionToken::*;
+        // Note: draft-3 documents aren't supported by `wdl`, but
+        // the lexer needs to ensure it can lex any valid version
+        // token so that the parser may gracefully reject parsing
+        // the document.
         let lexer = Lexer::<VersionToken>::new(
             "
 # Test for draft-3 documents
--- a/wdl-grammar/src/experimental/lexer.rs
+++ b/wdl-grammar/src/experimental/lexer.rs
@@ -1,0 +1,240 @@
+//! Module for the lexer implementation.
+
+use logos::Logos;
+use miette::SourceSpan;
+
+pub mod v1;
+
+/// Converts a logos `Span` into a miette `SourceSpan`.
+fn to_source_span(span: logos::Span) -> SourceSpan {
+    SourceSpan::new(span.start.into(), span.end - span.start)
+}
+
+/// Represents a token for lexing version directives in WDL documents.
+///
+/// A WDL parser may initially use this token to lex the `version`
+/// directive at the start of a WDL document.
+///
+/// Once the version directive has been parsed, the parser will then
+/// [morph][Lexer::morph] the lexer to the appropriate token for the
+/// document's WDL version and pass the lexer to the matching version
+/// of the WDL grammar.
+#[derive(Logos, Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+#[repr(u8)]
+#[logos(error = Error)]
+pub enum VersionToken {
+    /// Contiguous whitespace.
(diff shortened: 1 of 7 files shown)


wdl-grammar/src/experimental/lexer/v1.rs
```diff
@@ -0,0 +1,1339 @@
...
+    /// The `.` symbol.
+    #[token(".")]
+    Dot,
+
+    // WARNING: this must always be the last variant.
+    /// The exclusive maximum token value.
+    MAX,
+}
+
+// There can only be 128 tokens in a TokenSet.
```
peterhuene, line 432: FYI: `TokenSet` will be introduced in an upcoming PR; basically the parser can reference sets of tokens in a bitfield, which will be useful for error recovery and also "expected one of these tokens"-style diagnostics.

wdl-grammar/src/experimental/lexer.rs
```diff
@@ -0,0 +1,240 @@
...
+                (Ok(VersionKeyword), 26..33),
+                (Ok(Whitespace), 33..34),
+                (Ok(Version), 34..37)
+            ],
+            "produced tokens did not match the expected set"
+        );
+    }
+
+    #[test]
+    fn test_version_draft3() {
```
a-frantz, line 219: Someone can correct me if I'm wrong here, but I think for all of our sanity we only want to support 1.X versions of WDL. The drafts are all relatively ancient now, and I don't think much of anyone is using them actively.

wdl-grammar/src/experimental/lexer/v1.rs
```diff
@@ -0,0 +1,1339 @@
...
+            Self::TaskKeyword => write!(f, "`task` keyword"),
+            Self::ThenKeyword => write!(f, "`then` keyword"),
+            Self::TrueKeyword => write!(f, "`true` keyword"),
+            Self::VersionKeyword => write!(f, "`version` keyword"),
+            Self::WorkflowKeyword => write!(f, "`workflow` keyword"),
+            Self::ReservedDirectoryTypeKeyword => write!(f, "reserved `Directory` keyword"),
+            Self::ReservedHintsKeyword => write!(f, "reserved `hints` keyword"),
+            Self::ReservedRequirementsKeyword => write!(f, "reserved `requirements` keyword"),
+            Self::OpenBrace => write!(f, "`{{`"),
+            Self::CloseBrace => write!(f, "`}}`"),
```
a-frantz, line 484: Why are these braces doubled?

wdl-grammar/src/experimental/lexer.rs
peterhuene: Agreed on all points; this is more of a test of the regex to match a "version" than anything else (definitely not going to support `draft-3` files in the parser implementation). I'm happy to use a different version name to help reduce potential confusion over what we support.

rickymagner: FYI: Cromwell just released [v87](https://github.com/broadinstitute/cromwell/releases/tag/87) which retired the `womtool upgrade` command to convert draft-2 to version 1.0. The functionality obviously exists in older womtool versions, but I'd take this as a signal it's safe to stick to 1.x, and agree on the sanity aspect.

a-frantz: @peterhuene that makes sense! I think we can probably just add a comment stating the above instead of changing the test.

@rickymagner thanks for the input! Good to get some confirmation on this.

a-frantz (approved):
I think you know this, but for the time being we're going to ignore the Gauntlet CI(Though we should make an attempt at reworking it sooner rather than latter)

PR #25, opened 2024-05-02
feat: document_preamble lint rule

adthrasher:
This pull request adds a new rule to `wdl`.

- **Rule Name**: `document_preamble`
- **Rule Kind**: Lint warning
- **Rule Code**: `v1::W009`
- **Packages**: `wdl-grammar`

Implements document_preamble as defined in #11.

Before submitting this PR, please make sure:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.
- [x] Your changes are squashed into a single commit (unless there is a _really_
      good, articulated reason as to why there shouldn be more than one).

Rule specific checks:

- [x] You have added the rule as an entry within the the package-specific rule
      tables (`wdl-ast/src/v1.rs` for AST-based rules and 
      `wdl-grammar/src/v1.rs` for parse tree-based rules).
- [x] You have added the rule as an entry within the the global rule
      table at `RULES.md`.
- [x] You have added the rule to the appropriate `fn rules()`.
    - Validation rules added to `wdl-ast` should be added to `fn rules()` within 
      `wdl-ast/src/v1/validation.rs`.
    - Lint rules added to `wdl-ast` should be added to `fn rules()` within `wdl-ast/src/v1/lint.rs`.
    - Validation rules added to `wdl-grammar` should be added to `fn rules()` within 
      `wdl-grammar/src/v1/validation.rs`.
    - Lint rules added to `wdl-grammar` should be added to `fn rules()` within 
      `wdl-grammar/src/v1/lint.rs`.
- [x] You have added a test that covers every possible setting for the rule 
      within the file where the rule is implemented.
- [x] You have run `wdl-gauntlet --save-config` to ensure that all of the rules
      added/removed are now reflected in the baseline configuration file
      (`Gauntlet.toml`).

PR #32, opened 2024-05-09
feat: adding preamble_comment lint rule

simojoe:

- **Rule Name**: `preamble_comment`
- **Rule Kind**: Lint warning
- **Rule Code**: `v1::W010`
- **Packages**: `wdl-grammar`

Verify that comments before (and only before) the version declaration start with a double pound sign. 

The whitespace around the pound signs and the comment lines are checked in other rules.

- [X] Your changes are squashed into a single commit (unless there is a _really_

Rule specific checks:

- [X] You have added the rule as an entry within the the package-specific rule
- [X] You have added the rule as an entry within the the global rule
      table at `RULES.md`.
- [X You have added the rule to the appropriate `fn rules()`.
      `wdl-grammar/src/v1/lint.rs`.
- [X] You have added a test that covers every possible setting for the rule 
- [X] You have run `wdl-gauntlet --save-config` to ensure that all of the rules
      (`Gauntlet.toml`).


wdl-grammar/src/v1/lint/header_comment.rs
```diff
@@ -0,0 +1,178 @@
...
+use wdl_core::Version;
+
+use crate::v1;
+
+/// Detects header comments declaration
+#[derive(Debug)]
+pub struct HeaderComment;
+
+impl<'a> HeaderComment {
+    /// Creates a warning for header comments without double pound sign
```
a-frantz, line 22: @adthrasher do we want this part of the warning? When I wrote the rule I didn't intend to include this case, but maybe it should be warned on.

Historically we've had it both ways; and I'm fairly certain every time we had a case of mixing single-pound and double-pound header comments we've eventually either deleted the single ones or "upgraded" them to double-pound comments.

For additional context: double-pound header comments are meant to be copied into `sprocket` (or `wdldoc`) built docs, and single-pound "header" comments are just regular comments meant to be ignored by tooling.

A potential convention that keeps this bit of code would be:
```
## sprocket comment

version 1.1
# "regular" comment sprocket will ignore
# Copyright St. Jude Children's Research Hospital

import "../foo.wdl"
```
We've dubbed the header+version declaration the "document preamble", I think it makes sense to keep the preamble well defined (**only** header comments and the version).

wdl-grammar/src/v1/lint/header_comment.rs
```diff
@@ -0,0 +1,178 @@
...
+    /// Creates a warning for header comments after version declaration
+    fn header_comment_after_version(&self, location: Location) -> lint::Warning
+    where
+        Self: Rule<&'a Pair<'a, v1::Rule>>,
+    {
+        lint::warning::Builder::default()
+            .code(self.code())
+            .level(lint::Level::Low)
+            .group(self.group())
+            .subject("double pound sign are reserved for header comment")
```
a-frantz, line 51:
```suggestion
            .subject("double pound signs are reserved for header comments")
```

wdl-grammar/src/v1/lint/header_comment.rs
```diff
@@ -0,0 +1,178 @@
...
+            .body(
+                "Only full line comment before version declaration should start with a double \
```
a-frantz, line 53:
```suggestion
                "Only full line comments before the version declaration should start with a double \
```

a-frantz (changes requested):
@adthrasher please weigh in, but I'm thinking everywhere this PR says `header` should be updated to `preamble` so that we're consistent with the language adopted in #25 .

I coined the term "header comment" but I think before solidifying that in code we should pivot to calling them "preamble comments". Does that make sense?

adthrasher:
> @adthrasher please weigh in, but I'm thinking everywhere this PR says `header` should be updated to `preamble` so that we're consistent with the language adopted in #25 .
> 
> I coined the term "header comment" but I think before solidifying that in code we should pivot to calling them "preamble comments". Does that make sense?

I think we need to be careful here. For example in this text, `double pound signs are reserved for header comments`, the `double pound signs` would be a `preamble comment`, but they're reserved for use in the `document preamble`. I would reword it to something like `double pound signs/preamble comments are reserved for the document preamble`.

a-frantz:
> I think we need to be careful here. For example in this text, `double pound signs are reserved for header comments`, the `double pound signs` would be a `preamble comment`, but they're reserved for use in the `document preamble`. I would reword it to something like `double pound signs/preamble comments are reserved for the document preamble`.

Good catch that it will be more than a simple find and replace. But sounds like you're on board with the terminology change?

Going forward, we're going to retire the term "header" and instead use "preamble".

Besides some word smithing related to this change, the PR looks good to me. I requested @peterhuene as a reviewer to make sure I didn't miss anything on the Rust side of things.

peterhuene (approved):
From the Rust side of things, looks good 👍

adthrasher:
> > I think we need to be careful here. For example in this text, `double pound signs are reserved for header comments`, the `double pound signs` would be a `preamble comment`, but they're reserved for use in the `document preamble`. I would reword it to something like `double pound signs/preamble comments are reserved for the document preamble`.
> 
> Good catch that it will be more than a simple find and replace. But sounds like you're on board with the terminology change?
> 
> Going forward, we're going to retire the term "header" and instead use "preamble".
> 
> Besides some word smithing related to this change, the PR looks good to me. I requested @peterhuene as a reviewer to make sure I didn't miss anything on the Rust side of things.

Yes, I think that's a better approach moving forward.

simojoe:
I updated the naming as I understood from the conversation above! 

I kept the condition disallowing the normal comments in the preamble as was not clear if the converse was decided.

wdl-grammar/src/v1/lint/preamble_comment.rs
```diff
@@ -0,0 +1,180 @@
+//! Preamble comments are full line comments starting with a double pound sign
```
adthrasher, line 1:
```suggestion
//! Preamble comments are full line comments starting with a double pound sign and must occur before the version declaration
```

wdl-grammar/src/v1/lint/preamble_comment.rs
```diff
@@ -0,0 +1,180 @@
...
+    where
+        Self: Rule<&'a Pair<'a, v1::Rule>>,
+    {
+        lint::warning::Builder::default()
+            .code(self.code())
+            .level(lint::Level::Low)
+            .group(self.group())
+            .subject("preamble comment without a double pound sign")
+            .body(
+                "Header comments are full line comments before the version declaration and they \
```
adthrasher, line 33:
```suggestion
                "Preamble comments are full line comments before the version declaration and they \
```

adthrasher (approved):

PR #34, opened 2024-05-10
feat: implements true WDL string literal parsing

claymcleod:
This PR introduces true WDL string literal parsing (as opposed to simply punting it and storing it as a native `String`).

      changes (when appropriate).

### DIFF
--- a/wdl-ast/src/v1/document/expression/literal/string.rs
+++ b/wdl-ast/src/v1/document/expression/literal/string.rs
@@ -1,0 +1,330 @@
+//! Literal strings.
+
+pub mod inner;
+
+use grammar::v1::Rule;
+pub use inner::Inner;
+use pest::iterators::Pair;
+use wdl_grammar as grammar;
+use wdl_macros::check_node;
+
+/// A literal string.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum String {
+    /// A single quoted string.
+    SingleQuoted(Inner),
+
+    /// A double quoted string.
+    DoubleQuoted(Inner),
+}
+
+impl String {
+    /// If the [`String`] is a [`String::SingleQuoted`], a reference to the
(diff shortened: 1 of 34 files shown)


claymcleod:
There's something going on with `gauntlet` that won't allow it to accept the changes that needs to be looked at. Since I'm running short on time, I suggest we just code review and merge anyway for now and let the `gauntlet` strategy be rethought while I'm out.

wdl-ast/src/v1/document/expression/literal/string.rs
```diff
@@ -0,0 +1,330 @@
...
+use wdl_macros::check_node;
+
+/// A literal string.
+#[derive(Clone, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub enum String {
+    /// A single quoted string.
+    SingleQuoted(Inner),
+
+    /// A double quoted string.
+    DoubleQuoted(Inner),
```
a-frantz, line 18: @peterhuene since Clay is out I'm asking you:
I would think that 95% of the time we don't care about the distinction between single quotes and double quotes. Doesn't this structure complicate usage by requiring handling of the enum variants? I don't know what an alternative would look like, but could this be handled differently? In a way that obfuscates the quote character for general usage (but of course we still need that information stored _somewhere_).

Does this make sense?

peterhuene: I agree that it seems no longer necessary to make the distinction between single and double quoted strings in the AST; there's simply one "string literal" at this representation level.

In the `rowan`-based parser, there will only be a string literal "node" in the CST and the node would have child tokens for whatever opening/closing quote characters were in the parsed source.

peterhuene:
After discussion with folks on Slack, we've decided to close this PR for now in favor of the new parser implementation which has already implemented string representation (with full interpolation) in the AST.

If by the time Clay is back and we haven't switched over to the new parser implementation for some reason, then we can revisit this PR (but 🤞 that `pest` will be gone).

PR #39, opened 2024-05-20
feat: implement parsing `meta` and `parameter_meta` in experimental parser.

peterhuene:
This commit implements parsing of `meta` and `parameter_meta` sections in both tasks and workflows for the experimental parser.

This also includes a fix for lexing single-digit integers, a fix for a panic when parsing a `Map` key type that is not a primitive type name, and fixes to the `delimited` and `expect*` methods of `Parser`.

Error messages for unmatched braces and brackets have been introduced as well.

Unknown tokens are now properly added to the syntax tree too.

This commit also moves the `Event::Error` variant out into its own collection.

Doing so reduces the `Event` size from 48 bytes to 24 bytes; as there will be an event for every token in the file, the size of `Event` is important.

Errors are now tracked directly in the parser implementation and returned when the parser is finished.

      changes (when appropriate).


wdl-grammar/tests/parsing/metadata/source.wdl
```diff
@@ -0,0 +1,139 @@
+# This is a test of parsing task metadata sections.
```
a-frantz, line 1:
```suggestion
# This is a test of parsing task and workflow metadata sections.
```

wdl-grammar/tests/parsing/metadata/source.wdl
```diff
@@ -0,0 +1,139 @@
...
+        c: 5
+        d: -0xf
+        e: 1.0e10
+        f: -2.
+        g: true
+        h: false
+        i: null
+        j: {
+            a: [1, 2, 3],
+            b: ["hello", "world", "!"],
```
a-frantz, line 18: I believe metadata arrays must all be of the same type, but I don't see that tested/checked. What happens when the WDL looks like:
```suggestion
            a: [1, 2, 3, 4.5],
            b: ["hello", "world", "!", false],
```

peterhuene: You're right! However, this is just a syntax tree, which would accept arrays of any metadata value type regardless of the actual expressed type as that's what the grammar allows.

Type checking of metadata values can be done statically because they don't accept expressions and may be implemented during a future validation pass of the CST.

Type checking for (the soon-to-be PR'd) expressions would require a proper resolution of one or more ASTs (if there are `import` statements); we'll probably implement that in the future as part of a real execution engine.

peterhuene: That said, I've been trying to stick to as-syntactically-valid as possible tests so that a future validation pass doesn't require an update to these baselines (said validation would then have test casts for mismatched types in arrays).

a-frantz: All that makes sense! Thanks for the explanation.

a-frantz (approved):

PR #42, opened 2024-05-21
feat: implement bound decls in tasks and workflows in experimental parser.

peterhuene:
This commit implements bound declarations in tasks and workflows for the experimental parser.

As bound declarations support expressions, this also implements support for expressions, including in string interpolations.

A basic Pratt parser is implemented in the `expr_with_precedence` function.

      changes (when appropriate).


wdl-grammar/src/experimental/grammar/v1.rs
```diff
@@ -890,3 +1144,419 @@ fn call_statement(parser: &mut Parser<'_>, _marker: Marker) -> Result<(), (Marke
...
+        }
+    }
+
+    Ok(lhs)
+}
+
+/// Parses an atomic expression such as a literal.
+///
+/// Due to the WDL grammar having an ambiguity between parenthesized expressions
+/// and pair literals, this function also handles the former.
```
a-frantz, line 1279:
```suggestion
/// Due to the WDL grammar having an ambiguity between parenthesized expressions
/// and pair literals, this function handles the former in addition to pair literals.
```
I initially misread this comment. It was grammatically correct, but I found it confusing. I think this is a bit better.

wdl-grammar/tests/parsing/name-ref/source.wdl
```diff
@@ -0,0 +1,9 @@
+# This is a test of a name references.
+
+version 1.1
+
+task test {
+    Integer x = 0
+    Integer y = -x
+    Integer z = x + y * x
```
a-frantz, line 8:
```suggestion
    Int x = 0
    Int y = -x
    Int z = x + y * x
```
I think this is what was meant. What is currently being parsed is a custom type named `Integer` (that doesn't exist).

wdl-grammar/tests/parsing/postfix-exprs/source.wdl
```diff
@@ -0,0 +1,18 @@
+# This is a test of postfix expressions
+
+version 1.1
+
+task test {
+    Integer a = min(0, 1)
+    Integer b = min(max(100, a), 10)
+    Array[String] c = ["a", "b", "c"]
+    String d = d[a + b]
```
a-frantz, line 9: `d` is undefined when used on the RHS.

wdl-grammar/tests/parsing/postfix-exprs/source.wdl
```diff
@@ -0,0 +1,18 @@
+# This is a test of postfix expressions
+
+version 1.1
+
+task test {
+    Integer a = min(0, 1)
+    Integer b = min(max(100, a), 10)
```
a-frantz, line 7:
```suggestion
    Int a = min(0, 1)
    Int b = min(max(100, a), 10)
```

wdl-grammar/tests/parsing/precedence/source.wdl
```diff
@@ -0,0 +1,8 @@
+# This is a test of operator precedence
+
+version 1.1
+
+task test {
+    Boolean a = 0 || 1 && 1 == 0 != 1 < 0 <= 1 > 0 >= 1 + 2 - 3 * 4 / 5 % 6
+    Integer b = (1 + 2) - (3 * 4) / (5 % 6)
```
a-frantz, line 7:
```suggestion
    Boolean a = 0 || 1 && 1 == 0 != 1 < 0 <= 1 > 0 >= 1 + 2 - 3 * 4 / 5 % 6
    Int b = (1 + 2) - (3 * 4) / (5 % 6)
```

wdl-grammar/tests/parsing/precedence/source.wdl
```diff
@@ -0,0 +1,8 @@
+# This is a test of operator precedence
```
a-frantz, line 1: So in the Rust code I saw assignment and calculation of precedence. But am I correct that it isn't tested or reflected at this stage of the parse implementation? `source.tree` is just in the order it was found in the file, and doesn't have precedence information.

wdl-grammar/tests/parsing/prefix-exprs/source.wdl
```diff
@@ -0,0 +1,11 @@
+# This is a test of prefix expressions.
+
+version 1.1
+
+task test {
+    Boolean a = !true
+    Boolean b = !!!!!!!!!false
+    Int c = -2
+    Float d = -2.0
+    Int e = ---------42
```
a-frantz, line 10: Is this allowed? (I haven't checked the spec) I would hope/expect an error will be thrown for this at some stage.

wdl-grammar/tests/parsing/name-ref/source.wdl
peterhuene: Doh! `Integer` is legal an would be interpreted as a custom type name here; this is going to be a recurring problem in my WDL journey, I think 😭

wdl-grammar/tests/parsing/postfix-exprs/source.wdl
peterhuene: 😭

wdl-grammar/tests/parsing/precedence/source.wdl
peterhuene: I'm sensing a pattern...

wdl-grammar/tests/parsing/precedence/source.wdl
peterhuene: The precedence is captured in the depth of the tree, though. It just happens that the expressions are in lowest-to-highest order, so the tree shape matches.

I'll add a reverse test.

wdl-grammar/tests/parsing/prefix-exprs/source.wdl
peterhuene: It should be, it's just a series of negation expressions that are right-associative, no different from an explicit parenthesized expression of `-(-42)`, same [as in Rust](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=6b6ed2a2ac3075e63933fc0ee7366cd0).

a-frantz: Huh. Learned something new. IDK why, but I was working under the assumption that `-42` is the same as `--42` is the same as `---42` and so on. But in hindsight that doesn't make much sense. Ignore this then!

PR #45, opened 2024-05-23
feat: double_quotes

simojoe:

- **Rule Name**: `double_quotes`
- **Rule Kind**: Lint warning
- **Rule Code**: `v1::W012`
- **Packages**: `wdl-grammar`

Warns about strings defined with single-quotes. This ignores anything located inside the command section. Tests are not exhaustive, there are quite a lot of things that are using strings. Should they still be added?

- [X] You have added a few sentences describing the PR here.
- [ ] Your changes are squashed into a single commit (unless there is a _really_

Rule specific checks:

- [X] You have added the rule as an entry within the the package-specific rule
- [X] You have added the rule as an entry within the the global rule
      table at `RULES.md`.
- [X] You have added the rule to the appropriate `fn rules()`.
      `wdl-grammar/src/v1/lint.rs`.
- [ ] You have added a test that covers every possible setting for the rule 
- [X] You have run `wdl-gauntlet --save-config` to ensure that all of the rules
      (`Gauntlet.toml`).

### DIFF
--- a/wdl-grammar/src/v1/lint/double_quotes.rs
+++ b/wdl-grammar/src/v1/lint/double_quotes.rs
@@ -14,7 +14,7 @@ use wdl_core::Version;
 
 use crate::v1;
 
-///
+/// Detects strings that are not defined with double quotes.
 #[derive(Debug)]
 pub struct DoubleQuotes;
 
@@ -45,7 +45,7 @@ impl<'a> Rule<&'a Pair<'a, v1::Rule>> for DoubleQuotes {
     }
 
     fn tags(&self) -> lint::TagSet {
-        TagSet::new(&[lint::Tag::Naming])
+        TagSet::new(&[lint::Tag::Clarity, lint::Tag::Style])
     }
 
     fn check(&self, tree: &'a Pair<'_, v1::Rule>) -> lint::Result {
@@ -99,7 +99,7 @@ mod tests {
         assert_eq!(warnings.len(), 1);
         assert_eq!(
             warnings.first().to_string(),
-            "[v1::W012::[Naming, Style]::Low] string defined with single quotes (1:8-1:36)"
+            "[v1::W012::[Style, Clarity]::Low] string defined with single quotes (1:8-1:36)"
         );
 
         Ok(())
@@ -115,7 +115,7 @@ mod tests {
         assert_eq!(warnings.len(), 1);
         assert_eq!(
             warnings.first().to_string(),
-            "[v1::W012::[Naming, Style]::Low] string defined with single quotes (1:21-1:26)"
+            "[v1::W012::[Style, Clarity]::Low] string defined with single quotes (1:21-1:26)"
         );
 
         Ok(())
@@ -140,7 +140,7 @@ mod tests {
         assert_eq!(warnings.len(), 1);
         assert_eq!(
             warnings.first().to_string(),
-            "[v1::W012::[Naming, Style]::Low] string defined with single quotes (3:22-3:29)"
+            "[v1::W012::[Style, Clarity]::Low] string defined with single quotes (3:22-3:29)"
         );
 
         Ok(())
--- a/wdl-grammar/src/v1/lint/double_quotes.rs
+++ b/wdl-grammar/src/v1/lint/double_quotes.rs
@@ -1,0 +1,148 @@
+//! All quotes should be double quotes.
+
+use std::collections::VecDeque;
+
+use nonempty::NonEmpty;
+use pest::iterators::Pair;
+use wdl_core::concern::code;
+use wdl_core::concern::lint;
+use wdl_core::concern::lint::Rule;
+use wdl_core::concern::lint::TagSet;
+use wdl_core::concern::Code;
+use wdl_core::file::Location;
+use wdl_core::Version;
+
+use crate::v1;
+
+///
+#[derive(Debug)]
+pub struct DoubleQuotes;
+
+impl<'a> DoubleQuotes {
+    /// Creates a warning for strings defined using single-quotes
+    fn single_quote_string(&self, location: Location) -> lint::Warning
+    where
+        Self: Rule<&'a Pair<'a, v1::Rule>>,
+    {
+        // SAFETY: this error is written so that it will always unwrap.
+        lint::warning::Builder::default()
+            .code(self.code())
+            .level(lint::Level::Low)
+            .tags(self.tags())
+            .push_location(location)
+            .subject("string defined with single quotes")
+            .body("All strings should be defined using double quotes.")
+            .fix("Change the single quotes to double quotes.")
+            .try_build()
+            .unwrap()
+    }
+}
+
+impl<'a> Rule<&'a Pair<'a, v1::Rule>> for DoubleQuotes {
+    fn code(&self) -> Code {
+        // SAFETY: this manually crafted to unwrap successfully every time.
+        Code::try_new(code::Kind::Warning, Version::V1, 12).unwrap()
+    }
+
+    fn tags(&self) -> lint::TagSet {
+        TagSet::new(&[lint::Tag::Naming])
+    }
+
+    fn check(&self, tree: &'a Pair<'_, v1::Rule>) -> lint::Result {
+        let mut warnings = VecDeque::new();
+
+        for node in tree.clone().into_inner().flatten() {
+            if node.as_rule() == v1::Rule::string && node.as_str().starts_with('\'') {
+                let location = Location::try_from(node.as_span()).map_err(lint::Error::Location)?;
+                warnings.push_back(self.single_quote_string(location));
+            }
+        }
+
+        match warnings.pop_front() {
+            Some(front) => {
+                let mut results = NonEmpty::new(front);
+                results.extend(warnings);
+                Ok(Some(results))
+            }
+            None => Ok(None),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+
+    use pest::Parser as _;
+    use wdl_core::concern::lint::Rule as _;
+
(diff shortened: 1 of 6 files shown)


wdl-grammar/src/v1/lint/double_quotes.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+impl<'a> Rule<&'a Pair<'a, v1::Rule>> for DoubleQuotes {
+    fn code(&self) -> Code {
+        // SAFETY: this manually crafted to unwrap successfully every time.
+        Code::try_new(code::Kind::Warning, Version::V1, 12).unwrap()
+    }
+
+    fn tags(&self) -> lint::TagSet {
+        TagSet::new(&[lint::Tag::Naming])
```
a-frantz, line 48: I'm not sure this falls ender `Naming`? It's definitely `Style` (which is being auto-added bc of the `Naming` tag), and I think this rule calls for a new tag called `Consistency`. And if we're adding a `Consistency` tag, that should be added to a few of the existing rules: `whitespace`, `no_curly_commands`, `mixed_indentation`, `snake_case`? Open to hearing dissenting opinions on any of these specifics here.

a-frantz (commented):
Does this work for nested placeholder strings? I haven't checked out the branch to test myself. e.g. `String nested = "Hello ~{if alien then 'world' else planet}!"`. The inner `'world'` string should be flagged. There are other cases we probably want to check as well.

I'm fine with skipping over the `command` block for now. That seems like a bonus check to me, that shouldn't hold up the other test conditions.

wdl-grammar/src/v1/lint/double_quotes.rs
adthrasher: It is labeled `Naming` in the discussion (https://github.com/stjude-rust-labs/wdl/discussions/11), but I agree. I'm not sure that `Naming` is correct.

wdl-grammar/src/v1/lint/double_quotes.rs
```diff
@@ -0,0 +1,148 @@
...
+use wdl_core::concern::lint;
+use wdl_core::concern::lint::Rule;
+use wdl_core::concern::lint::TagSet;
+use wdl_core::concern::Code;
+use wdl_core::file::Location;
+use wdl_core::Version;
+
+use crate::v1;
+
+///
```
adthrasher, line 17: Does this need a comment?

wdl-grammar/src/v1/lint/double_quotes.rs
adthrasher: `mixed_indentation` uses `Style|Spacing|Clarity`. Is `Clarity` sufficient here or does `Consistency` convert something different?

adthrasher: `snake_case` similarly uses `Naming|Style|Clarity`

a-frantz: Hm. They are similar, but different.

Generally, I'd argue "too many tags" is a better problem to have than "not enough tags". Tags serve the (primary?) function of making the docs easier to navigate and the (secondary?) function of providing easy shorthand for groups of rules that a user would want to toggle at once. (For the sake of argument, we will pretend those functions exist today lol)

If those are our two main goals for tags, I don't think there's a real problem with having too many. The biggest downside (I can see) is that having lots of tags muddies the warning's shorthand notation (`[v1::W004::[Spacing, Style, Clarity]::High]`).

simojoe: What is the final decision here? Only `Clarity` or `Clarity|Style`?

wdl-grammar/src/v1/lint/double_quotes.rs
simojoe: Yes, forgot this one. Strangely none of the steps described in the README picked it up but the lint here did.

Will add.

wdl-grammar/src/v1/lint/double_quotes.rs
a-frantz: I misread @adthrasher 's comments. Let me try addressing this again lol (so @simojoe hold on before making any changes)

>  I would tend to favor `Clarity` here, ironically, for consistency with the already implemented rules.

I don't see how `Clarity` applies to `double_quotes`. This rule isn't a question of "what's clearer?".

As authors, everyone is faced with what is a more or less identical choice between single and double quotes. We've arbitrarily chosen double quotes to be the standard because switching back and forth is bad. But as long as it's consistent, it doesn't matter.

That's why I want to add `Consistency` here. And then in hindsight, `Consistency` _could_ apply to some of our other rules.

What's the argument against having both `Consistency` and `Clarity`?

adthrasher: @a-frantz I think it's worth trying to avoid too many tags if we can. When I look at the rules `snake_case`, `no_curly_commands`, and `mixed_indentation`, in all three cases we are choosing a preferred style for the purpose of `Clarity`. Technically, all would be `Consistency` as well.

a-frantz: options:
1. no new `Consistency` tag
    * IMO `Clarity` doesn't quite fit for `double_quotes`. Which means `double_quotes` would only have the tag `Style`.
2. replace all instances of `Clarity` with `Consistency`
    * `double_quotes` gets both `Consistency` and `Style`
3. Have both `Clarity` and `Consistency`.
    * They appear together for `snake_case`, `no_curly_commands`, and `mixed_indentation` _but not_ `double_quotes` (which only gets `Consistency`)
 4. ??? is there anything else worth considering?
 
I would rule out number 2. Sounds like you want to rule out number 3. I don't feel strongly about number 1, so maybe that's the compromise we settle on?

adthrasher: Copying/summarizing conversation with @a-frantz from Slack, but the conclusion is that `double_quotes` should get `Style` and `Clarity` tags.

Personally, I view `consistency` as a necessary component of `clarity`. If something is inconsistent (e.g. mixing single and double quotes), it lacks clarity. The choice of double quotes vs single quotes comes down to a matter of `Style` in my opinion as well. So `Style|Clarity` seems to fit.

There will also likely be an effort in the future to define a set of tags to support and write descriptions for each. With that in mind, keeping a smaller set of tags now makes that easier. We can always add tags to existing rules as part of that effort.

adthrasher:
> Does this work for nested placeholder strings? I haven't checked out the branch to test myself. e.g. `String nested = "Hello ~{if alien then 'world' else planet}!"`. The inner `'world'` string should be flagged. There are other cases we probably want to check as well.
> 
> I'm fine with skipping over the `command` block for now. That seems like a bonus check to me, that shouldn't hold up the other test conditions.

It does not appear to catch that condition.

adthrasher:
I rebased this to get the latest changes. Looking at the `Arena` results, it does appear to be checking within the `command` block.

a-frantz:
> I rebased this to get the latest changes. Looking at the `Arena` results, it does appear to be checking within the `command` block.

Ok so that's a good thing I think. As long as it doesn't apply to literal Bash code. e.g. look at this hypothetical command block
```
command <<<
    echo 'this Bash string should be ignored'
    echo "~{if foo then 'this should be flagged' else 'this one too'}"
>>>
```

adthrasher:
> > I rebased this to get the latest changes. Looking at the `Arena` results, it does appear to be checking within the `command` block.
> 
> Ok so that's a good thing I think. As long as it doesn't apply to literal Bash code. e.g. look at this hypothetical command block
> 
> ```
> command <<<
>     echo 'this Bash string should be ignored'
>     echo "~{if foo then 'this should be flagged' else 'this one too'}"
> >>>
> ```

To clarify, it filters to `v1::Rule::string` instances. So it will ignore anything that isn't part of a WDL instruction in the `command` block. So in your example, it ignores the first line, but will catch the two embedded strings.

a-frantz:
> > > I rebased this to get the latest changes. Looking at the `Arena` results, it does appear to be checking within the `command` block.
> > 
> > 
> > Ok so that's a good thing I think. As long as it doesn't apply to literal Bash code. e.g. look at this hypothetical command block
> > ```
> > command <<<
> >     echo 'this Bash string should be ignored'
> >     echo "~{if foo then 'this should be flagged' else 'this one too'}"
> > >>>
> > ```
> 
> To clarify, it filters to `v1::Rule::string` instances. So it will ignore anything that isn't part of a WDL instruction in the `command` block. So in your example, it ignores the first line, but will catch the two embedded strings.

That's perfect then. Let's add a my example as an actual test case for sanity, update Arena, and then this should be good to merge.

adthrasher:
> > > > I rebased this to get the latest changes. Looking at the `Arena` results, it does appear to be checking within the `command` block.
> > > 
> > > 
> > > Ok so that's a good thing I think. As long as it doesn't apply to literal Bash code. e.g. look at this hypothetical command block
> > > ```
> > > command <<<
> > >     echo 'this Bash string should be ignored'
> > >     echo "~{if foo then 'this should be flagged' else 'this one too'}"
> > > >>>
> > > ```
> > 
> > 
> > To clarify, it filters to `v1::Rule::string` instances. So it will ignore anything that isn't part of a WDL instruction in the `command` block. So in your example, it ignores the first line, but will catch the two embedded strings.
> 
> That's perfect then. Let's add a my example as an actual test case for sanity, update Arena, and then this should be good to merge.

Test added. I've updated Arena, but I was going to hold off until https://github.com/stjudecloud/workflows/pull/157 is merged. However, I can commit the Arena warnings and move ahead with this first if you prefer.

a-frantz (approved):

PR #46, opened 2024-05-23
 feat: implement input and output sections in the experimental parser.

peterhuene:
This commit implement parsing of input and output sections in the experimental
parser for both tasks and workflows.

Also implemented is the caching of a lexer peek operation which prevents the
lexer from having to tokenize the input again.

The `expect*` methods of `Parser` now peek and move next only if the token
matches what's expected; this will prevent consuming recovery tokens in
recovery contexts.

Before submitting this PR, please make sure:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added tests (when appropriate).
- [x] You have updated the README or other documentation to account for these
      changes (when appropriate).
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.
- [x] Your changes are squashed into a single commit (unless there is a _really_
      good, articulated reason as to why there should be more than one).

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.0.0/

PR #43, opened 2024-05-22
feat: add support for import statements to the experimental parser.

peterhuene:
This commit adds support for import statements to the experimental parser, along with a simple test for the import statement syntax.

Before submitting this PR, please make sure:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added tests (when appropriate).
- [x] You have updated the README or other documentation to account for these
      changes (when appropriate).
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.
- [x] Your changes are squashed into a single commit (unless there is a _really_
      good, articulated reason as to why there should be more than one).

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.0.0/

PR #47, opened 2024-05-23
feat: implement parsing command sections in the experimental parser. 

peterhuene:
This commit implements parsing of command sections in tasks for the
experimental parser.

      changes (when appropriate).

### DIFF
--- a/wdl-grammar/src/experimental/lexer/v1.rs
+++ b/wdl-grammar/src/experimental/lexer/v1.rs
@@ -111,14 +111,22 @@ pub enum HeredocCommandToken {
     #[token("~")]
     Tilde,
 
-    /// An ending angle bracket.
+    /// A single close angle bracket (not the end).
     ///
-    /// When three of these tokens are sequentially encountered,
-    /// [morph][super::Lexer::morph] the lexer to use [Token].
-    ///
-    /// Otherwise, consider the token to be part of the
-    /// literal text.
+    /// This token is part of the literal text.
     #[token(">")]
+    SingleCloseAngle,
+
+    /// A double close angle bracket (not the end).
+    ///
+    /// This token is part of the literal text.
+    #[token(">>")]
+    DoubleCloseAngle,
+
+    /// An ending triple close angle bracket.
+    ///
+    /// When encountered, [morph][super::Lexer::morph] the lexer to use [Token].
+    #[token(">>>")]
     End,
 }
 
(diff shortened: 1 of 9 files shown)


peterhuene:
This PR depends on #43 and #46 to review; when those merge, this will be rebased and marked as ready.

wdl-grammar/src/experimental/lexer/v1.rs
```diff
@@ -111,14 +111,22 @@ pub enum HeredocCommandToken {
...
-    /// literal text.
+    /// This token is part of the literal text.
     #[token(">")]
+    SingleCloseAngle,
+
+    /// A double close angle bracket (not the end).
+    ///
+    /// This token is part of the literal text.
+    #[token(">>")]
+    DoubleCloseAngle,
```
a-frantz, line 124: What's the purpose of these tokens? Shouldn't `>` and `>>` be treated as any other literal text?

wdl-grammar/src/experimental/tree.rs
```diff
@@ -203,6 +205,10 @@ pub enum SyntaxKind {
     InputSectionNode,
     /// Represents an output section node.
     OutputSectionNode,
+    /// Represents a braced command section node.
+    BracedCommandSectionNode,
+    /// Represents a heredoc command section node.
+    HeredocCommandSectionNode,
```
a-frantz, line 211: Just a thought: there's a good deal of redundant code created by treating these as separate Nodes. Does it make sense to consolidate the code for these cases into just a plain `CommandSectionNode`? That may prove more complicated than it's worth. I'm not sure.

peterhuene: > don't you need to differentiate because it changes how contained items are interpreted?

At this point we've already parsed the interpolation of the command contents, so it shouldn't matter for the CST.

wdl-grammar/src/experimental/lexer/v1.rs
peterhuene: Due to `logos` not being able to support backtracking as it only looks at an input exactly once, we can't have a pattern which effectively says "only match an unescaped `>>>`".  It's why the regular expression for literal text within the command is `[^\\~>]+`, which stops matching text at the first `>` that is encountered.

Therefore we then have to account for a single `>`, a double `>>`, and finally a truly closing `>>>`. The first two are simply treated as part of the text when we're parsing the interpolation.

a-frantz: Ok that makes sense. Thanks for explaining!

a-frantz (approved):

PR #52, opened 2024-05-29
fix: allow reserved names in the experimental parser.

peterhuene:
The experimental parser is too strict regarding reserved names in the parse, specifically in metadata and parameter metadata sections, where keys are allowed to be reserved names.

This commit fixes the parser to accept reserved names in addition to identifiers; the places where reserved names are not allowed will be checked during (forthcoming) validation of the tree instead.

This also fixes the incorrect recovery in metadata sections which might cause an infinite loop: the recovery was incorrectly set to recover at a value instead of any identifier, which meant that recovery would move to a token that would immediately fail, recovery would then not move, and the cycle would repeat.

Added tests for both problems.

Additionally the `NameNode` and `QualifiedNameNode` nodes have been removed from the tree in favor of simply looking for "any identifier token child" in the forthcoming AST.

      changes (when appropriate).

### DIFF
--- a/wdl-grammar/tests/parsing/reserved-meta-names/source.tree
+++ b/wdl-grammar/tests/parsing/reserved-meta-names/source.tree
@@ -19,7 +19,7 @@ RootNode@0..175
       OpenBrace@92..93 "{"
       Whitespace@93..102 "\n        "
       MetadataObjectItemNode@102..118
-        VersionKeyword@102..109 "version"
+        Ident@102..109 "version"
         Colon@109..110 ":"
         LiteralStringNode@110..118
           Whitespace@110..111 " "
@@ -35,7 +35,7 @@ RootNode@0..175
       OpenBrace@145..146 "{"
       Whitespace@146..155 "\n        "
       MetadataObjectItemNode@155..166
-        TaskKeyword@155..159 "task"
+        Ident@155..159 "task"
         Colon@159..160 ":"
         LiteralStringNode@160..166
           Whitespace@160..161 " "
--- a/wdl-grammar/tests/parsing/reserved-meta-names/source.tree
+++ b/wdl-grammar/tests/parsing/reserved-meta-names/source.tree
@@ -1,0 +1,49 @@
+RootNode@0..175
+  Comment@0..56 "# This is a test of r ..."
+  Whitespace@56..58 "\n\n"
+  VersionStatementNode@58..69
+    VersionKeyword@58..65 "version"
+    Whitespace@65..66 " "
+    Version@66..69 "1.1"
+  Whitespace@69..71 "\n\n"
+  TaskDefinitionNode@71..174
+    TaskKeyword@71..75 "task"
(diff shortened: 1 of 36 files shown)


wdl-grammar/tests/parsing/reserved-meta-names/source.tree
```diff
@@ -0,0 +1,49 @@
...
+    Whitespace@80..81 " "
+    OpenBrace@81..82 "{"
+    Whitespace@82..87 "\n    "
+    MetadataSectionNode@87..124
+      MetaKeyword@87..91 "meta"
+      Whitespace@91..92 " "
+      OpenBrace@92..93 "{"
+      Whitespace@93..102 "\n        "
+      MetadataObjectItemNode@102..118
+        VersionKeyword@102..109 "version"
```
a-frantz, line 22: I imagine I'm misunderstanding something about `rowan` or `logo`, but shouldn't the string `version` be parseable as an `Ident` `Token`? That's what it is in _this_ context, and `version` matches the regex used for `Ident` (`[a-zA-Z][a-zA-Z0-9_]*`).

peterhuene: That's a good point; let me figure out a way to substitute an ident token here, which will also simplify the AST as it then doesn't have to search for the reserved keyword tokens either.

a-frantz:
> Additionally the NameNode and QualifiedNameNode nodes have been removed from the tree in favor of simply looking for "any identifier token child" in the forthcoming AST.

Can you elaborate on this? Won't this make resolving namespaces more difficult, as instead of a `QualifiedNameNode`, we will have to `peek()` for `Dot` and then do some kind of join on following `Ident` tokens?

peterhuene:
> Can you elaborate on this? Won't this make resolving namespaces more difficult, as instead of a QualifiedNameNode, we will have to peek() for Dot and then do some kind of join on following Ident tokens?

The plan is to have a `task` method on the `CallStatement` AST node that returns `(Option<Ident>, Ident)`, where the first is the namespace and the second is the task name.

It will accomplish this by finding the first child ident token; it will then look for one more ident token child. If only one was found, it'll return `(None, task)`; if both are found, it'll return `(Some(namespace), task)`.

If we kept it as-is, the search for the two child tokens still occurs, it would just happen in a `QualifiedName` type rather than the `CallStatement` type, which would instead have a `name` method that returns `QualifiedName`.

This change simply reduces the depth of the tree slightly, as it isn't really needed.

peterhuene:
I'll also add that `peek()`ing will no longer be necessary as we've already constructed the CST from the input.

The forthcoming AST representation is a facade over the CST which knows how to navigate the tree to extract only semantically relevant information; users of the AST won't need to navigate the CST in most cases (an exception would be a lint rule that wants to see if the document ends with a newline; that's not represented in the AST, but one can easily answer that question from the CST).

a-frantz (approved):

PR #61, opened 2024-05-30
fix(wdl-grammar): false positives in preamble_comment

### LINKED ISSUE #50: Consolidate preamble comment warnings
See an example of a "bad" preamble here: https://github.com/aws-samples/amazon-omics-tutorials/blob/main/example-workflows/gatk-best-practices/workflows/germline-fastqs-to-vcf/sub-workflows/fastq-to-ubam.wdl

I'm not sure who owns this code (whether it's the Broad or Amazon), but they are using double-pound sign comments the way we use them, but _after_ the version declaration. This causes a LintWarning to be thrown for every line of "the preamble" of that file. We should update to detect a continuous preamble _in the wrong location_ and warn once on the whole span.

a-frantz:
This pull request ~adds a new~ modifies a rule in `wdl`.

- **Rule Name**: `preamble_comment`
- **Rule Kind**: Lint warning
- **Rule Code**: `v1::W010`
- **Packages**: `wdl-grammar`

Closes #50 . This will only throw one lint warning for continuous blocks of comments. It also permits triple+ pound sign comments.

Rule specific checks:

- [ ] You have added the rule as an entry within the the package-specific rule
- [ ] You have added the rule as an entry within the the global rule
      table at `RULES.md`.
- [ ] You have added the rule to the appropriate `fn rules()`.
      `wdl-grammar/src/v1/lint.rs`.
- [x] You have added a test that covers every possible setting for the rule 
- [ ] You have run `wdl-gauntlet --refresh` to ensure that there are no unintended
      changes to the baseline configuration file (`Gauntlet.toml`).
- [x] You have run `wdl-gauntlet --refresh --arena` to ensure that all of the rules
      (`Arena.toml`).

### DIFF
--- a/wdl-grammar/src/v1/lint/preamble_comment.rs
+++ b/wdl-grammar/src/v1/lint/preamble_comment.rs
@@ -166,7 +166,7 @@ version 1.0
         assert_eq!(warnings.len(), 1);
         assert_eq!(
             warnings.first().to_string(),
-            "[v1::W010::[Style]::Low] preamble comment without a double pound sign (1:1-1:12)"
+            "[v1::W010::[Style]::Low] preamble comment(s) without a double pound sign (1:1-1:12)"
         );
         Ok(())
     }
@@ -189,7 +189,7 @@ version 1.0
         assert_eq!(warnings.len(), 1);
         assert_eq!(
             warnings.first().to_string(),
-            "[v1::W010::[Style]::Low] preamble comment without a double pound sign (1:1-3:21)"
+            "[v1::W010::[Style]::Low] preamble comment(s) without a double pound sign (1:1-3:21)"
         );
         Ok(())
     }
--- a/wdl-grammar/src/v1/lint/preamble_comment.rs
+++ b/wdl-grammar/src/v1/lint/preamble_comment.rs
@@ -30,13 +30,13 @@ impl<'a> PreambleComment {
             .code(self.code())
             .level(lint::Level::Low)
             .tags(self.tags())
-            .subject("preamble comment without a double pound sign")
+            .subject("preamble comment(s) without a double pound sign")
             .body(
                 "Preamble comments are full line comments before the version declaration and they \
                  start with a double pound sign.",
             )
             .push_location(location)
-            .fix("Add a pound sign at the start of the line.")
+            .fix("Add a pound sign at the start of each offending comment.")
             .try_build()
             .unwrap()
     }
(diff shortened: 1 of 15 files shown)


a-frantz:
On Slack we discussed some other comment cases we wanted to throw warnings on. Should this rule be expanded to cover those cases? Or should those additional cases go into another rule? I think as a general rule we want to prefer multiple targeted rules over fewer broad rules.

We've already split this rule (preamble_comment) from the document_preamble rule. But maybe we should only have one rule for the preamble??? I'm not sure how we want to divvy up/classify these warnings.

wdl-grammar/src/v1/lint/preamble_comment.rs
```diff
@@ -135,7 +166,30 @@ version 1.0
...
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn it_catches_badly_formatted_preamble_comment_block() -> Result<(), Box<dyn std::error::Error>>
+    {
+        let tree = Parser::parse(
+            Rule::document,
+            r#"# a comment
```
adthrasher, line 179: So we're explicitly disallowing single pound comments in the preamble?

a-frantz: Yes. That's not new in this PR. I thought we had a discussion about it back when this rule was first implemented.

adthrasher: In the `document_preamble` lint check, I only look for `COMMENT` before the version. For some reason, I thought this rule (`preamble_comment`) was only about disallowing the double-pound syntax outside of the preamble.

adthrasher: I don't think we should change it if we've discussed previously. I was just confirming that this was our intention here.

adthrasher:
Did we decide on whether no spacing after a `#` is allowable, e.g. `##this is a comment`? I'm guessing if so, that should be a separate rule.

wdl-grammar/src/v1/lint/preamble_comment.rs
a-frantz: The fact you had to ask to clarify whether this was an intended case or not is more evidence that we need to write a `body` (as defined in #12 ) for each rule. The current situation is that you need to code dive to discern what exactly a rule covers. We should have it written in plain text somewhere.

adthrasher (approved):

PR #63, opened 2024-05-31
fix: fix disambiguation of struct literals in the experimental parser.

peterhuene:
This commit fixes the disambiguation of struct literal expressions in the experimental parser.

Previously, we looked ahead from the identifier by one token to see if it is an open brace; if so, we would assume it is a struct literal.

However, this conflicts with other grammar rules that expect an open brace following an expression, such as conditional statements in workflows.

The fix is to look ahead three tokens for `{`, \<ident>, and `:` to consider it a struct literal; if those tokens aren't present, then the expression will be treated as a name reference.

Note that this means that empty struct literals (and by extension struct definitions themselves) are invalid; while the WDL spec doesn't clarify this requirement, the example WDL parsers from OpenWDL do require at least one member specifier in a struct literal as well.

      changes (when appropriate).


peterhuene:
CI is in a bad state due to the failed publish of `wdl-core`. I think if we succeed in publishing just `wdl-core`, stop there, re-run this CI, merge this, then continue on publishing for the remaining crates we'll be fine.

wdl-grammar/src/experimental/grammar/v1.rs
```diff
@@ -1798,21 +1802,37 @@ fn literal_struct_or_name_ref(
...
-    if !matches!(parser.peek(), Some((Token::OpenBrace, _)) | None) {
-        // This is actually a name reference.
-        return Ok(marker.complete(parser, SyntaxKind::NameReferenceNode));
+    // To disambiguate between a name reference and a struct literal,
+    // we need to peek ahead 3 tokens for `{`, <any-ident>, `:`.
+    // Start by peeking one for open brace and then do the slightly
+    // more expensive `peek3` operation.
+    if let Some((Token::OpenBrace, _)) = parser.peek() {
+        if let Some(peek) = parser.peek3() {
+            dbg!(&peek);
```
a-frantz, line 1811: Should line be here?

peterhuene: Whoops, it should not.

a-frantz (approved):

PR #64, opened 2024-05-31
feat: implement an AST for the experimental parser.

peterhuene:
This commit adds an `experimental` feature to `wdl-ast` that feature-gates a new AST implementation based off of `rowan` and the experimental parser implementation in `wdl-grammar`.

The AST implementation is a facade over the `SyntaxNode` representing the concrete syntax tree produced by `wdl-grammar`'s experimental parser. The AST is therefore cheap to construct and clone at any level. Changes to the tree require replacing nodes in the tree to form a new tree (see [SyntaxNode::replace_with](https://docs.rs/rowan/latest/rowan/api/struct.SyntaxNode.html#method.replace_with)). A simplified editing API will likely be added in the future.

The new AST implementation parses given WDL source and produces an AST if the version in the source was supported; this differs from the existing implementation where the version has to be specified up-front before parsing.

An example of parsing WDL source into the new AST:

```rust
match Document::parse(source).into_result() {
    Ok(document) => {
        match document.ast() {
            Ast::V1(ast) => {
                // Document was 1.x
            }
            _ => bail!("unsupported WDL version"),
        }
    }
    Err(errors) => {
        // The document had parse errors
    }
}
```

Additionally, a new `Visitor` trait is added for the `visit` method of the AST.

The visitor will be called upon entry and exit of each node of the AST during a preorder traversal.

Hopefully this abstraction will simplify lint rules that concern themselves with inspecting specific node types (e.g. a lint rule that inspects every literal string in the tree).

      changes (when appropriate).

### DIFF
--- a/wdl-ast/src/experimental/v1/expr.rs
+++ b/wdl-ast/src/experimental/v1/expr.rs
@@ -944,6 +944,24 @@ impl AstToken for StringText {
 pub struct Placeholder(pub(super) SyntaxNode);
 
 impl Placeholder {
+    /// Returns whether or not placeholder has a tilde (`~`) opening.
+    ///
+    /// If this method returns false, the opening was a dollar sign (`$`).
+    pub fn has_tilde(&self) -> bool {
+        self.0
+            .children_with_tokens()
+            .find_map(|c| match c.kind() {
+                SyntaxKind::PlaceholderOpen => Some(
+                    c.as_token()
+                        .expect("should be token")
+                        .text()
+                        .starts_with('~'),
+                ),
+                _ => None,
+            })
+            .expect("should have a placeholder open token")
+    }
+
     /// Gets the options for the placeholder.
     pub fn options(&self) -> AstChildren<PlaceholderOption> {
         children(&self.0)
@@ -2362,16 +2380,9 @@ task test {
         let parts: Vec<_> = s.parts().collect();
         assert_eq!(parts.len(), 3);
         assert_eq!(parts[0].clone().unwrap_text().as_str(), "Hello, ");
-        assert_eq!(
-            parts[1]
-                .clone()
-                .unwrap_placeholder()
-                .expr()
-                .unwrap_name_ref()
-                .name()
-                .as_str(),
-            "name"
-        );
+        let placeholder = parts[1].clone().unwrap_placeholder();
+        assert!(!placeholder.has_tilde());
+        assert_eq!(placeholder.expr().unwrap_name_ref().name().as_str(), "name");
         assert_eq!(parts[2].clone().unwrap_text().as_str(), "!");
 
         // Fourth declaration
@@ -2382,10 +2393,10 @@ task test {
         let parts: Vec<_> = s.parts().collect();
         assert_eq!(parts.len(), 3);
         assert_eq!(parts[0].clone().unwrap_text().as_str(), "String");
+        let placeholder = parts[1].clone().unwrap_placeholder();
+        assert!(placeholder.has_tilde());
         assert_eq!(
-            parts[1]
-                .clone()
-                .unwrap_placeholder()
+            placeholder
                 .expr()
                 .unwrap_literal()
                 .unwrap_string()


wdl-ast/src/experimental.rs
```diff
@@ -0,0 +1,392 @@
...
+mod test {
+    use pretty_assertions::assert_eq;
+
+    use super::*;
+
+    #[test]
+    fn document_has_preamble_comments() {
+        let parse = Document::parse(
+            r#"
+## This is a preamble comment.
```
a-frantz, line 370: @adthrasher @peterhuene another instance where we want to be careful about pedantic definitions (because we're making up the terms):

**(a) Are all comments within the preamble "preamble comments"? OR (b) are _only_ double-pound-sign comments "preamble comments"?**

(a) is more intuitive. But then we need a new term for double-pound-sign comments. "Doc comments"? Then the only "valid" doc comments would be preamble doc comments. I think double-pound-sign comments outside the preamble would still be referred to as doc comments, they just aren't supported in any special way (and produce warnings).

wdl-ast/src/experimental/v1/expr.rs
```diff
@@ -0,0 +1,4922 @@
...
+    #[test]
+    fn literal_string() {
+        let parse = Document::parse(
+            r#"
+version 1.1
+
+task test {
+    String a = "hello"
+    String b = 'world'
+    String c = "Hello, ${name}!"
```
a-frantz, line 2326: Is it captured by the AST whether a placeholder uses `~` or `$`?
And related, should we have a lint for ensuring `~` is always used?

a-frantz (commented):
Close to done reviewing, but wanted to get these 2 comments out there. I only highlighted one, but there are multiple instances in `expr.rs` where `Integer` should be `Int`.

wdl-ast/src/experimental/v1/task.rs
```diff
@@ -0,0 +1,1119 @@
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MetadataObjectItem(pub(super) SyntaxNode);
+
+impl MetadataObjectItem {
+    /// Gets the name of the item.
+    pub fn name(&self) -> Ident {
+        token(&self.0).expect("expected a name")
+    }
+
+    /// GEts the value of the item.
```
a-frantz, line 495:
```suggestion
    /// Gets the value of the item.
```

wdl-ast/src/experimental/v1/task.rs
```diff
@@ -0,0 +1,1119 @@
...
+        let commands: Vec<_> = tasks[0].commands().collect();
+        assert_eq!(commands.len(), 1);
+
+        // First task command
+        assert!(commands[0].is_heredoc());
+        let parts: Vec<_> = commands[0].parts().collect();
+        assert_eq!(parts.len(), 3);
+        assert_eq!(
+            parts[0].clone().unwrap_text().as_str(),
+            "\n        printf \"hello, "
```
a-frantz, line 916: At what point will this leading whitespace be stripped? Should that be done by the AST?

a-frantz (changes requested):
Woowee that was a doozy of a PR. Please see comments, but overall this looks great 🤠

wdl-ast/src/experimental.rs
a-frantz: > we could also just move its implementation into the preamble lint rule as well if this helps avoid confusion.

I think this is the right call. As of right now, the official WDL spec has no concept of a preamble or preamble comment. Those are entirely our own inventions (that we hope to spread), but IMO that makes this function out of place. The CST and AST should be more-or-less devoid of this team's opinions/conventions/etc, and focus on the "letter of the law". `wdl-lint` is where we can put our "spirit of the law" interpretations.

wdl-ast/src/experimental/v1/expr.rs
adthrasher: I think a lint to ensure `~{}` is consistently used (even outside `command` blocks) is a good choice.

wdl-ast/src/experimental/v1/expr.rs
```diff
@@ -0,0 +1,4922 @@
...
+
+    #[test]
+    fn less_equal() {
+        let parse = Document::parse(
+            r#"
+version 1.1
+
+task test {
+    Integer a = 1
+    Integer b = 2
```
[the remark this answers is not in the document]
peterhuene: I'm apparently incapable of not having at least one of these in a PR! 😞

wdl-ast/src/experimental/v1/task.rs
peterhuene: I think the AST should give the literal text of the command and it's up to a lint rule to check it and an execution engine to format it to whatever script file it's writing out.

wdl-ast/src/experimental/v1/expr.rs
```diff
@@ -0,0 +1,4922 @@
...
+        let tasks: Vec<_> = ast.tasks().collect();
+        assert_eq!(tasks.len(), 1);
+        assert_eq!(tasks[0].name().as_str(), "test");
+
+        // Task declarations
+        let decls: Vec<_> = tasks[0].declarations().collect();
+        assert_eq!(decls.len(), 3);
+
+        // First declaration
+        assert_eq!(decls[0].ty().to_string(), "Integer");
```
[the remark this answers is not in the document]
peterhuene: If we called it `decl_type`, would that help or be redundant that it's on a declaration?

wdl-ast/src/experimental.rs
peterhuene: Sounds good, I'll remove it and let the lint rule implement it.

wdl-ast/src/experimental/v1/task.rs
a-frantz: Ok. Makes sense as the execution engine's responsibility (i.e. the AST does not need to do any whitespace trimming)

wdl-ast/src/experimental/v1/expr.rs
a-frantz: Sounds good all around 👍

wdl-ast/src/experimental/v1/expr.rs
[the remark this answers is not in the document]
a-frantz: I'm not following what you mean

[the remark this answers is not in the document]
peterhuene: I think I misunderstood what you're asking then; what I thought you were asking was "is the method name `ty` (in lieu of `type` as that's a keyword) insufficient in conveying that it's returning the WDL type of the declaration vs. a native Rust type?"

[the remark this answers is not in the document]
a-frantz: Sorry, I was vague. What I meant was should the return of `ty()` be "smarter" than just returning the literal text it finds? Should there be some kind of validation here? (The answer very well could be "no, it should just return the raw text")

The thought here is triggered by `Integer` not being a valid WDL type in this context but the AST doesn't have any indication of that error.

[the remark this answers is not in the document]
peterhuene: It's actually not returning the literal text it finds, but formatting the string based on the AST representation of the type.

The problem is that `Integer` is very much a legal WDL type name, it's just treated as a type reference to a hypothetical `Integer` struct.

So the `Display` implementation just prints back the name of the type being referenced.

wdl-ast/src/experimental/v1/expr.rs
peterhuene: Added a `Placeholder::has_tilde` convenience method.

a-frantz (approved):

PR #66, opened 2024-06-04
fix: properly validate negative numbers in metadata sections.

peterhuene:
Metadata sections do not allow expressions, so a minus token is intentionally
parsed to be part of the literal integer or float, if present.

However, validation was not taking this into account properly; the fix is to
check to see if the minus token is present in the literal (which only happens
in metadata sections) when validating the value.

This also cleans up how `LiteralInteger` values work, splitting handling
negation into its own method.

Also includes some code clean-up to use the `token` function from `rowan` and
to use the existing `SyntaxElement` instead of `NodeOrToken` where appropriate.

      changes (when appropriate).


peterhuene:
Marking this as draft as I think there's a better way to handle this that isn't so confusing.

a-frantz (approved):

PR #68, opened 2024-06-06
feat: introduce a `wdl-lint` crate.

peterhuene:
This commit introduces a new crate named `wdl-lint` that will be the singular place where lint rules are defined; the rules are based on the new parser implementation.

`wdl-lint` builds on top of `wdl-ast` to provide a version-specific `Rule` trait and the rules that implement that trait.

As the rules are implemented using the `Visitor` trait, they plug into a `Validator`:

```rust
let rules = wdl_lint::v1::rules();
let mut validator = Validator::default();
validator.add_v1_visitors(rules.iter().map(|r| r.visitor());
let result = validator.validate(&document);
```

After some discussion, we determined that `miette` wasn't going to meet our needs for diagnostic reporting; therefore, this commit includes removing `miette` as a dependency.

Instead a `Diagnostic` type was introduced to `wdl-grammar` and also re-exported from `wdl-ast`. The same type is now used to report parser diagnostics (in `wdl-grammar`), validation diagnostics (in `wdl-ast`), and lint rule diagnostics (in `wdl-lint`), which unifies error handling.

Three existing lint rules were ported as part of these changes, with more coming:

* `double_quotes`
* `ending_newline` (formerly `newline_eof`)
* `missing_runtime` (formerly `missing_runtime_block`)

      changes (when appropriate).


wdl-lint/tests/lints/multiple-eof-newline/source.wdl
```diff
@@ -0,0 +1,8 @@
+# This is a test of multiple EOF newlines lint
+
+version 1.1
+
+workflow test {
+
+}
+
```
a-frantz, line 8:
```suggestion



```
I'd like to make sure that the `span` reported covers _every_ superfluous newline.

wdl-lint/tests/lints/missing-runtime-block/source.errors
```diff
@@ -0,0 +1,8 @@
+warning[missing_runtime]: task `bad` is missing a runtime section
```
a-frantz, line 1: Is `<severity>[<ID>]` clear enough that ID is an _identifier_? Let's assume a user who has never encountered a lint warning before. (I think that's a real possibility with some biologists being introduced to programming with WDL)

Is the `[<ID>]` clear enough that _this is what to look up in the documentation_? I think it becomes clear if there's an additional message at the bottom saying `look at 'sprocket explain <ID>' for more info`, but that message doesn't belong as built into the `Diagnostic`. This problem is similarly avoided if we provide a URL. But without those two, I'm worried we're not being clear enough.

wdl-lint/src/v1/missing_runtime.rs
```diff
@@ -0,0 +1,76 @@
...
+/// The identifier for the missing runtime rule.
+const ID: &str = "missing_runtime";
+
+/// Creates a "missing runtime section" diagnostic.
+fn missing_runtime_section(task: &str, span: Span) -> Diagnostic {
+    Diagnostic::warning(format!("task `{task}` is missing a runtime section"))
+        .with_code(ID)
+        .with_label("this task is missing a runtime section", span)
+        .with_fix(
+            "add a runtime section to the task with desired cpu, memory and container/docker \
```
a-frantz, line 24:
```suggestion
            "add a runtime section to the task with desired cpu, memory and container \
```
The spec used to only use docker images (hence the validity of the `docker` key), but it's been expanded to (I think, going off the top) any OCI registry. We should remain implementation agnostic here and just say `container`.

a-frantz (commented):
Went a little out of order in the review here. Will look over `wdl-ast` in the morning. But overall I'm excited for these changes. Great job 🚀

wdl-lint/src/v1/missing_runtime.rs
a-frantz: Another thought. We have another lint planned for "you should really define these runtime keys" and I don't remember what all they are off the top of my head. But this language should include all those keys... Or no keys 🤔 But mentioning `cpu` and not `maxRetries` strikes me as inconsistent (once that rule is implemented). I'd lean towards not mentioning any specific keys in this message, as mentioning everything we will require might be too verbose.

wdl-lint/src/v1.rs
```diff
@@ -0,0 +1,70 @@
+//! Module for V1 lint rules.
```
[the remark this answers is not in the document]
peterhuene: In the parser re-implementation, I tried to abstract places where we'd want to re-use things between different versions of the language; for example `Parser` takes the token type it works with as a type argument, so presumably we would just implement a new token type for V2 and re-use the parser; as `SyntaxKind` is expected to be the _union_ of all the WDL versions, we could potentially reuse V1 grammar rule functions directly from V2 grammar functions as well.

For the `Visitor` trait, however, I made it tied to a particular version of the AST. My hope is that we'll try to make use of V1 AST representations from a V2 AST for things that haven't changed. A `v2::Visitor` trait might be `: v1::Visitor` and only include members for the V2 AST representations (i.e. things that have changed).

I don't know exactly what the right story is for supporting different major versions and it definitely doesn't feel like we're there yet, but what I absolutely do know is that users of the API should not have to decide up-front what version to use, rather the API should respond to the version declared in the document and "do the right thing".

wdl-lint/tests/lints/missing-runtime-block/source.errors
peterhuene: I think the majority of users will see these messages in the context of `sprocket` and it will be pretty clear with the additional provided context ("see `sprocket explain`" and "here's the rule's URL for more information").

That said, I'm happy to make this more clear in usage outside of sprocket; right now we're piggy backing on `codespan-reporting`'s notion of `code` here, but we can format it anyway we like.

Possible alternative?

```
warning: task `bad` is missing a runtime section [lint: missing_runtime]
```

wdl-lint/src/v1.rs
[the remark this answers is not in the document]
peterhuene: I think we could possible make `Visitor` and `Rule` versionless, and by extensions the rules are all unversioned. 

When a new major version comes out, we parse it into a CST like we do now, keeping the same `SyntaxKind` for the elements that haven't changed and add new `SyntaxKind` variants for those that have (e.g. `SyntaxKind::TaskDefinitionV2`). We add AST representations for those V2 elements and extend `Visitor` to visit them, passing the V2 representations.

A rule that concerns itself with whitespace or comment tokens would continue to work as those are (presumably) not going to change from version to version. Likewise, if a rule checks a syntax construct that is identical from version to version it doesn't need to do anything to continue to work.

I will have to think on this a bit more once the dust settles of the parser implementation change over, but it sounds like this is all doable.

wdl-grammar/src/experimental/diagnostic.rs
```diff
@@ -0,0 +1,260 @@
...
+
+impl From<logos::Span> for Span {
+    fn from(value: logos::Span) -> Self {
+        Self::new(value.start, value.len())
+    }
+}
+
+/// Represents the severity of a diagnostic.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub enum Severity {
```
a-frantz, line 60: I think the notion of `Severity` still needs some workshopping. Continuing the conversation from #12, here's my idea for `Severity`:
- `Error` will be anything and everything coming out of `wdl-grammar` and `wdl-ast`
    - it means "something is _wrong_ with the provided code" and "it violates the WDL specification"
    - `wdl-lint` should never emit an `Error`
    - Is there a time where `wdl-grammar` or `wdl-ast` should emit a `Warning` or `Note`? I think not.
- `Warning` this is emitted by `wdl-lint`
    - It means "this is not 'good' WDL. A best practice has been ignored."
    - I've changed my mind from a previous position and now think every lint violation should be a `Warning`
    - In `sprocket lint` these should have a non-zero exit code
- `Note` - I don't see a purpose for `Note`
    - Maybe I just need a good example pointed out to me though 🤷

wdl-lint/tests/lints/missing-runtime-block/source.errors
a-frantz: > I think the majority of users will see these messages in the context of sprocket and it will be pretty clear with the additional provided context ("see sprocket explain" and "here's the rule's URL for more information").

This is true, so maybe it's unnecessary to make _this_ clearer. But it can't hurt, right?

IMO the `error`/`warning`/`note` distinction isn't really serving us at the diagnostic/output stage. (See [this comment](https://github.com/stjude-rust-labs/wdl/pull/68#discussion_r1630434927) for my argument on that)

So I'd recommend a format like:
```
[missing_runtime]: task `bad` is missing a runtime section
```
I think boxing it (with `[]`), putting it at the front, and the fact it's snake_cased, all work together to bring attention to the ID. Although... What if we PascalCased our IDs? I think starting with a capital letter could be a clearer demarkation that this is an ID of some sort.
```
[MissingRuntime]: task `bad` is missing a runtime section
```
PascalCase is looking more "official" or something to me. IDK the right words to describe it, but it has a better ~~~vibe~~~.

That said, I also see your suggestion (`[lint: missing_runtime]` at the end) working. That's pretty explicit that "this is the lint title/ID". If we agree to drop `warning`/`note` then I vote we go with my format. If we decide to keep `warning`/`note`, we can go with your format. I vote for PascalCasing the IDs either way though

wdl-grammar/src/experimental/diagnostic.rs
peterhuene: Keeping in mind that `note`, `warning`, and `error` all colorize differently in normal terminal output, I still see value in `note` conveying what we used to convey with a `Low` lint level: this particular lint isn't all that important, but it is valuable for style and consistency.

It would convey to users that the lint is more opinion than "something's potentially wrong here, please investigate" and would be a reasonable thing to disable if one doesn't agree with the lint.

> Is there a time where `wdl-grammar` or `wdl-ast` should emit a Warning or Note? I think not.

Real-world compilers emit warnings and I don't want to limit this for the future or make the tooling "linting-oriented"; while the parser itself might not emit warnings, certainly a future WDL evaluator would emit warnings (e.g. "unused variable", "unreachable code", etc.) in addition to errors (e.g. "unknown name reference", "type mismatch", etc.) and I don't want to have to use a different `Diagnostic` type for that either.

We should also keep in mind that linting is _opt-in_ too; there will be plenty of other tools we may want to build in the future that will produce warnings and errors (and maybe even notes).

wdl-lint/src/v1/ending_newline.rs
```diff
@@ -0,0 +1,111 @@
...
+        .with_code(ID)
+        .with_label("expected a newline to follow this", span)
+        .with_fix("add an empty line at the end of the file")
+}
+
+/// Creates a "multiple ending newline" diagnostic.
+fn multiple_ending_newline(span: Span) -> Diagnostic {
+    Diagnostic::note("multiple empty lines at the end of file")
+        .with_code(ID)
+        .with_label("duplicate newline here", span)
```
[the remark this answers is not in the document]
peterhuene: Updated to print either `newline` or `newlines` depending on how many extraneous newlines were found.

wdl-grammar/src/experimental/diagnostic.rs
a-frantz: Good points! Making me put my thinking cap on 🤠 (I wear a cowboy hat when deep in thought)

> Real-world compilers emit warnings and I don't want to limit this for the future or make the tooling "linting-oriented"; while the parser itself might not emit warnings, certainly a future WDL evaluator would emit warnings (e.g. "unused variable", "unreachable code", etc.) in addition to errors (e.g. "unknown name reference", "type mismatch", etc.) and I don't want to have to use a different Diagnostic type for that either.
> 
> We should also keep in mind that linting is opt-in too; there will be plenty of other tools we may want to build in the future that will produce warnings and errors (and maybe even notes).

This has convinced me to lighten up on my position. I initially wanted to see this `Severity` `enum` to only have `Error` and `Warning`, but you're right that there are plenty of use cases for `Note`. _But_, I don't think any of those `Note` uses are being implemented today/in this PR. I still feel all lints, even the whitespace ones, belong at the `Warning` level.

> this particular lint isn't all that important, but it is valuable for style and consistency.
> It would convey to users that the lint is more opinion than "something's potentially wrong here, please investigate" and would be a reasonable thing to disable if one doesn't agree with the lint.

I was clearly at one point in favor of having a `Low` category, and by extension a past version of me would be agreeing with this logic. But @adthrasher changed my mind.

We should keep our ultimate goal in mind. We _are_ opinionated. We think these lints are the best way to write WDL. We want people to adopt the lints we're implementing. The current state of community WDL is _a problem we're trying to **fix**_.

If we want people to adopt our lints, why would we say some of them are less important? It feels like encouraging them to disable certain lints. I'd rather present a unified front, that all of these lints matter and should be adopted.

wdl-lint/src/v1.rs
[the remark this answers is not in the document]
peterhuene: I'm going to resolve this for now but I think I'll have a follow-up PR next week to make `Visitor` and `Rule` versionless.

wdl-grammar/src/experimental/diagnostic.rs
peterhuene: > If we want people to adopt our lints, why would we say some of them are less important? It feels like encouraging them to disable certain lints. I'd rather present a unified front, that all of these lints matter and should be adopted.

I still have some (mild) concerns over treating all lint rules as having the same importance, and also a default of "non-zero exit on warning", as being _barriers_ to adoption rather than _promoters_ of adoption, though.

I've turned on new linting tools on large, existing source bases before and it can be _a lot_ of work to get a linter happy. Sometimes knowing that a thing the linter is complaining about isn't all that important (vs ones that need to be immediately fixed) and disabling it _for now_ is the right call to make to progress towards the goal of being linter "clean".

With that all said, I propose we keep `Diagnostic` with the severity levels for future flexibility, but change all of the lint rules to `Diagnostic::warning` to meet our goal of promoting consistency to the larger WDL community.

We can always revisit that for the future based on user feedback.

wdl-lint/tests/lints/missing-runtime-block/source.errors
peterhuene: One argument that I might put forth to keep the severity prefix for the output is that my intention is to have both the validation rules and the lint rules run in a single pass over the AST when you do `sprocket lint` (conversely, `sprocket check` would do basically the same thing except it wouldn't add the lint rules to the validator).

Thus one might get a mix of errors (coming from the validation rules) and warnings (coming from the lint rules) in the output, so I think it'd be good to have the severity label present.

There's also something to be said about consistency with most source-parsing tools out there that fairly universally print a severity label before a message.

wdl-grammar/src/experimental/diagnostic.rs
a-frantz: That's a valid perspective! I think we're both bringing real concerns to the table.

> We can always revisit that for the future based on user feedback.

Yes, I meant to include something to that effect. We don't need to feel tied to any decision we make today. I would go so far as to say we need to keep ourselves open to the community's feedback on this.

> I've turned on new linting tools on large, existing source bases before and it can be a lot of work to get a linter happy. Sometimes knowing that a thing the linter is complaining about isn't all that important (vs ones that need to be immediately fixed) and disabling it for now is the right call to make to progress towards the goal of being linter "clean".

I've been there, and that is one reason I consider it a high priority to get `sprocket fmt` implemented.

With all this in mind, I'd be happy charging ahead with all lint's being at the `Warning` level. It will be trivial to update pending more input on the issue.

wdl-lint/src/v1/missing_runtime.rs
peterhuene: I'm also fine with `add a runtime section to the task` with perhaps an additional note of `for a list of mandatory runtime attributes, see https://github.com/openwdl/wdl/blob/wdl-1.1/SPEC.md#mandatory-runtime-attributes`?

wdl-lint/tests/lints/missing-runtime-block/source.errors
a-frantz: I had not considered either of those points. I think you're right, let's keep the `Severity` prefix (and move the rule name to the end,)
```
warning: task `bad` is missing a runtime section [lint: MissingRuntime]
```
Thoughts on PascalCase?

wdl-lint/src/v1/missing_runtime.rs
peterhuene: Although, that part of the spec is more geared towards what an execution engine _must_ support rather than what is required of WDL authors, it seems.

Perhaps just `add a runtime section to the task` and let the other lint rule validate the expected required runtime keys for WDL authors.

wdl-lint/tests/lints/missing-runtime-block/source.errors
peterhuene: I don't have a preference for either, to be honest. Whatever folks think would look good both from a diagnostic output standpoint but also from a configuration standpoint (in a TOML/JSON/YAML file or in a source comment like `# allow: <id>`).

wdl-lint/src/v1/missing_runtime.rs
a-frantz: A little tangent: "Mandatory" has a bit of a dubious meaning in that document, because they aren't actually mandatory to specify. They are mandatory to have a default, but that's an implementation detail for engines that users don't need to be concerned with.
Me and Andrew have complained previously that the WDL spec doc is one of the best sources for new WDL users/writers to learn. And that being the best is not a compliment 😅 There should be a better place to link to for these things, instead of reading through a lengthy SPEC doc. So I'd rather not link there in these contexts, but IDK where _to_ link instead.

For now, maybe just leave it at `add a runtime section to the task` and we can revisit later? IDK I'm torn on the right approach here.

peterhuene: I opted for `add a runtime section to the task` and not mention any keys to delegate that to the upcoming rule that will ensure the expected keys are present.

wdl-grammar/src/experimental/diagnostic.rs
peterhuene: I've updated them to the warning level.

wdl-lint/src/v1/missing_runtime.rs
a-frantz: O we wrote comments at the same time lol. Looks like we came to the same conclusion though. Sounds good! 👍

wdl-lint/tests/lints/missing-runtime-block/source.errors
a-frantz: I'm partial to PascalCase in all those contexts. In my mind that leading capital imparts a certain weight or importance that is good for a rule to have. Authority 🧑‍⚖️

peterhuene: I've pushed up a change to pascal case and to also put the rule in the suffix rather than after the diagnostic.

Let's see what people think.

a-frantz (approved):
Looks great to me! Thanks for all this. One more PR to reach rule parity. And then one PR after that to officially swap in the experimental and delete the current parser! 🎉

a-frantz (approved):

wdl-grammar/src/experimental/diagnostic.rs
adthrasher: >   * Is there a time where `wdl-grammar` or `wdl-ast` should emit a `Warning` or `Note`? I think not.

Peter highlighted a couple of places below. Another is language features slated for deprecation. They are valid for the current version, but the parser should `warn` or `note` those to the user.

PR #71, opened 2024-06-08
feat: porting SnakeCase rule

a-frantz:

- **Rule Name**: `SnakeCase`

Enforces snake_case for workflows, tasks, and variables.

### DIFF
--- a/wdl-lint/src/v1/snake_case.rs
+++ b/wdl-lint/src/v1/snake_case.rs
@@ -59,9 +59,9 @@ const ID: &str = "SnakeCase";
 
 /// Creates a "snake case" diagnostic.
 fn snake_case(context: Context, name: &str, properly_cased_name: &str, span: Span) -> Diagnostic {
-    Diagnostic::warning(format!("{context} identifier `{name}` is not snake_case"))
+    Diagnostic::warning(format!("{context} name `{name}` is not snake_case"))
         .with_rule(ID)
-        .with_label("this identifier must be snake_case", span)
+        .with_label("this name must be snake_case", span)
         .with_fix(format!("replace `{name}` with `{properly_cased_name}`"))
 }
 
@@ -114,8 +114,8 @@ struct SnakeCaseVisitor {
 }
 
 impl SnakeCaseVisitor {
-    /// Determines current context.
-    fn determine_context(&self) -> Context {
+    /// Determines current declaration context.
+    fn determine_decl_context(&self) -> Context {
         if self.within_struct {
             Context::Struct
         } else if self.within_input {
--- a/wdl-lint/src/v1/snake_case.rs
+++ b/wdl-lint/src/v1/snake_case.rs
@@ -7,6 +7,9 @@ use convert_case::Boundary;
 use convert_case::Case;
 use convert_case::Converter;
 use wdl_ast::experimental::v1::BoundDecl;
+use wdl_ast::experimental::v1::InputSection;
+use wdl_ast::experimental::v1::OutputSection;
+use wdl_ast::experimental::v1::StructDefinition;
 use wdl_ast::experimental::v1::TaskDefinition;
 use wdl_ast::experimental::v1::UnboundDecl;
 use wdl_ast::experimental::v1::Visitor;
@@ -28,8 +31,14 @@ enum Context {
     Task,
     /// The warning occurred in a workflow.
     Workflow,
-    /// The warning occurred in a variable.
-    Variable,
+    /// The warning occurred in a struct.
+    Struct,
+    /// The warning occurred in an input section.
+    Input,
+    /// The warning occurred in an output section.
+    Output,
+    /// The warning occurred in a private declaration.
+    PrivateDecl,
 }
 
 impl fmt::Display for Context {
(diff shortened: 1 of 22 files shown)


wdl-lint/src/v1/snake_case.rs
a-frantz, line 1: I've flip flopped in referring to [un]bound-declarations as "variables" or "declarations". I think I've been staring at both words so long they seem weird to me now 😂  Neither feels right in all instances in `source.errors` and IDK what an alternative could be.

Point is, I'm looking for feedback on the terminology and phrasing used throughout these text fields.

wdl-lint/src/v1/snake_case.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    fn visitor(&self) -> Box<dyn Visitor<State = Diagnostics>> {
+        Box::new(SnakeCaseVisitor)
+    }
+}
+
+/// Implements the visitor for the snake case rule.
+struct SnakeCaseVisitor;
+
+impl Visitor for SnakeCaseVisitor {
```
a-frantz, line 94: The four functions below are highly repetitive, but IDK how to consolidate them any further. Is that just a consequence of the visitor pattern, or am I missing something?

peterhuene: What if we combined the conversion and check into a `check_name` function that looks like:

```rust
fn check_name(context: Context, name: &str, span: Span, diagnostics: &mut Diagnostics) { 
    let converter = Converter::new()
        .remove_boundaries(&[Boundary::DigitLower, Boundary::LowerDigit])
        .to_case(Case::Snake);
    let properly_cased_name = converter.convert(name);
    if name != properly_cased_name {
        let warning = snake_case(context, name, &properly_cased_name, span);
        diagnostics.add(warning);
     }
}
```

Which would simplify the visitor callbacks to just:

```rust
if reason == VisitReason::Exit {
    return;
}

let name = task.name();
check_name(Context::Task, name.as_str(), name.span(), state);
```

for `task_definition`, etc.

wdl-lint/src/v1/snake_case.rs
peterhuene: Indeed, both `variable` and `declaration` seem wrong for members of a struct, for example.

I think we could set a flag when we enter a `struct_definition` and clear the flag upon exit; if the flag is set and we see an unbound declaration, we name it a "struct member".

We could do the same for `input_section` and `output_section` so that we name either a bound/unbound decl inside an input section as "input" or a bound decl inside an output section as "output", respectively.

If we encounter a bound declaration not in an input or output section, then we call it a "private declaration", perhaps?

wdl-lint/src/v1/snake_case.rs
peterhuene: Obviously the above would need to be a little more complex with respect to `Context` if we want to give unbound/bound decls clearer names.

wdl-lint/tests/lints/snake-case/source.wdl
```diff
@@ -0,0 +1,33 @@
...
+    call good_task
+}
+
+task BadTask {
+    command <<<
+        echo "Hello World"
+    >>>
+    output {
+        File badOut = "out.txt"
+    }
```
peterhuene, line 16: Relating to the above comments, we should probably add `input` sections that have both bound and unbound decls.

wdl-lint/src/v1/snake_case.rs
a-frantz: I like all those ideas! Also gives me a reason to try out some other features of the new parser, so I'll see what I can do!

wdl-lint/src/v1/snake_case.rs
a-frantz: I didn't think of passing the `state`, but yes that's a good idea.

wdl-lint/src/v1/snake_case.rs
```diff
@@ -0,0 +1,229 @@
...
+#[derive(Debug, Default)]
+struct SnakeCaseVisitor {
+    within_struct: bool,
+    within_input: bool,
+    within_output: bool,
+}
+
+impl SnakeCaseVisitor {
+    /// Determines current context.
+    fn determine_context(&self) -> Context {
```
peterhuene, line 118: Naming nit:

```suggestion
    /// Determines current declaration context.
    fn determine_decl_context(&self) -> Context {
```

wdl-lint/src/v1/snake_case.rs
```diff
@@ -0,0 +1,229 @@
...
+        }
+    }
+}
+
+/// The identifier for the snake_case rule.
+const ID: &str = "SnakeCase";
+
+/// Creates a "snake case" diagnostic.
+fn snake_case(context: Context, name: &str, properly_cased_name: &str, span: Span) -> Diagnostic {
+    Diagnostic::warning(format!("{context} identifier `{name}` is not snake_case"))
```
peterhuene, line 62: Nit: I think from the user's perspective these are "names" (although from the grammar they're identifiers):

```suggestion
    Diagnostic::warning(format!("{context} name `{name}` is not snake_case"))
```

wdl-lint/src/v1/snake_case.rs
```diff
@@ -0,0 +1,229 @@
...
+        .with_rule(ID)
+        .with_label("this identifier must be snake_case", span)
```
peterhuene, line 64: Same nit comment as above:

```suggestion
        .with_label("this name must be snake_case", span)
```

peterhuene:
Looks like we're just missing a few doc comments for a green CI, but other than that and the few comments above, this looks really good!

peterhuene (approved):
Looks great, thanks for porting this!

PR #72, opened 2024-06-09
feat: port preamble comment and whitespace lint rules.

peterhuene:
This commit ports the preamble comment and whitespace lint rules to the new parser.

It adds supports for visiting whitespace and comment tokens as well.

Token visitations no longer get called twice as they never have children, so the `VisitReason` was removed as a parameter.

Also contains a fix to the comment tokenization so that comments don't capture a carriage return on `\r\n` line endings.

      changes (when appropriate).

### DIFF
--- a/wdl-lint/src/v1/preamble_whitespace.rs
+++ b/wdl-lint/src/v1/preamble_whitespace.rs
@@ -185,9 +185,25 @@ impl Visitor for PreambleWhitespaceVisitor {
                 // Whitespace token is valid
                 return;
             }
+
+            // Don't include the newline separating the previous comment from the whitespace
+            let offset = if s.starts_with("\r\n") {
+                2
+            } else if s.starts_with('\n') {
+                1
+            } else {
+                0
+            };
+
+            let span = whitespace.span();
+            state.add(unnecessary_whitespace(Span::new(
+                span.start() + offset,
+                span.len() - offset,
+            )));
+            return;
         }
 
-        // At this point, the whitespace is unnecessary
+        // At this point, the whitespace is entirely unnecessary
         state.add(unnecessary_whitespace(whitespace.span()));
     }
 }
--- a/wdl-lint/src/v1/preamble_whitespace.rs
+++ b/wdl-lint/src/v1/preamble_whitespace.rs
@@ -1,0 +1,192 @@
+//! A lint rule that checks for an incorrect preamble whitespace.
+
+use wdl_ast::experimental::v1::Visitor;
+use wdl_ast::experimental::AstNode;
+use wdl_ast::experimental::AstToken;
+use wdl_ast::experimental::Diagnostic;
+use wdl_ast::experimental::Diagnostics;
+use wdl_ast::experimental::Span;
+use wdl_ast::experimental::SyntaxElement;
+use wdl_ast::experimental::SyntaxKind;
+use wdl_ast::experimental::ToSpan;
+use wdl_ast::experimental::VersionStatement;
+use wdl_ast::experimental::VisitReason;
+use wdl_ast::experimental::Whitespace;
+
+use super::Rule;
+use crate::util::lines_with_offset;
+use crate::Tag;
+use crate::TagSet;
+
+/// The identifier for the preamble whitespace rule.
+const ID: &str = "PreambleWhitespace";
+
+/// Creates an "unnecessary whitespace" diagnostic.
+fn unnecessary_whitespace(span: Span) -> Diagnostic {
+    Diagnostic::note("unnecessary whitespace in document preamble")
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_fix("remove the unnecessary whitespace")
+}
+
+/// Creates an "expected a blank line" diagnostic.
+fn expected_blank_line(span: Span) -> Diagnostic {
+    Diagnostic::note("expected a blank line before version statement")
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_fix("add a blank line between the last preamble comment and the version statement")
+}
+
+/// Detects incorrect whitespace in a document preamble.
+#[derive(Debug, Clone, Copy)]
+pub struct PreambleWhitespaceRule;
+
+impl Rule for PreambleWhitespaceRule {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Ensures that documents have correct whitespace in the preamble."
+    }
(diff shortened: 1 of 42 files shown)


wdl-lint/src/v1/preamble_whitespace.rs
```diff
@@ -0,0 +1,192 @@
...
+                                usize::from(range.start()) + start,
+                                line.len(),
+                            )));
+                        }
+
+                        count += 1;
+                        if count == 2 {
+                            // If the previous byte isn't a newline, then we are missing a blank
+                            // line
+                            if text.as_bytes()[next_start - 1] != b'\n' {
```
a-frantz, line 124: Can you explain this check to me? I'm not following what this `if` statement accomplishes

wdl-lint/src/v1/preamble_whitespace.rs
```diff
@@ -0,0 +1,192 @@
...
+                            break;
+                        }
+                    }
+
+                    // We expected two lines
+                    if count < 2 {
+                        state.add(expected_blank_line(range.to_span()));
+                    }
+                } else {
+                    // Whitespace without a comment before it
```
a-frantz, line 148: This is another case that isn't clear to me.

wdl-lint/src/v1/preamble_whitespace.rs
```diff
@@ -0,0 +1,192 @@
...
+impl Rule for PreambleWhitespaceRule {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Ensures that documents have correct whitespace in the preamble."
+    }
+
+    fn explanation(&self) -> &'static str {
```
a-frantz, line 53: Your implementation of this rule is impressively thorough! You're catching lots of cases I hadn't thought about. I wonder if all the requirements for whitespace are being communicated via the below `explanation()`? I'm not sure 🤔 Maybe a simple addition like `No extraneous whitespace is allowed.` will cover all our bases?

wdl-lint/tests/lints/missing-comment-space/source.wdl
```diff
@@ -0,0 +1,8 @@
+##This preamble comment is missing a space.
+## But this one isn't!
```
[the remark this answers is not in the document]
peterhuene: Good catch, I'll fix.

wdl-lint/src/v1/preamble_whitespace.rs
peterhuene: It ensures that there were two actual lines as `lines_with_offsets` will return back a line even if there's no line terminator at the end (i.e. if it can't find a line terminator it returns the remainder of the string as the last "line").

Imagine you had the following:

```wdl
## Hello
  version 1.1
```

The whitespace token would be `\n  ` (a newline with two spaces after it). `lines_with_offsets` returns one blank line and one line with two spaces in it; if we don't check to see that the current line doesn't end with a `\n`, we'd fail to report the missing expected blank line between the last comment and the version statement.

Consequently, we'd report this as two warnings: unnecessary whitespace for the two spaces and a missing blank line between last comment and version statement.

a-frantz: That makes sense. Thanks!

wdl-lint/src/v1/preamble_whitespace.rs
peterhuene: This case can be hit in two ways.

The first is that there's a single whitespace token before the version statement:

```wdl
  
  
version 1.1
```

In the above case, the entire whitespace token is unnecessary.

The other is a parse error as only a comment token would normally precede a whitespace token in the preamble:

```wdl
task oops {
}

version 1.1
```

Here we actually end up emitting an error that a version statement must come before the task and the remainder of the file, including the version statement, is treated as "unparsed"; it would subsequently not be linted.

wdl-lint/tests/lints/preamble-comment-after-version/source.errors
```diff
@@ -1,15 +1,25 @@
+note: unnecessary whitespace in document preamble [rule: PreambleWhitespace]
+  ┌─ tests/lints/preamble-comment-after-version/source.wdl:1:38
+  │  
+1 │   ## This is a normal preamble comment.
+  │ ╭─────────────────────────────────────^
+2 │ │        ## Bad leading whitespace!
+  │ ╰───────^
```
a-frantz, line 7: How hard would it be to drop the leading newline from the reported span? It's being indicated as problematic, but the newline is fine (and is actually needed). The span should only cover the literal spaces before the comment.

wdl-lint/tests/lints/missing-comment-space/source.wdl
```diff
@@ -1,5 +1,7 @@
 ##This preamble comment is missing a space.
-## But this one isn't!
+##
+## But this one isn't and neither are the empty ones
+##
```
a-frantz, line 4: O this is a potentially interesting case. Should we allow line 4? I'm not sure. In the [future] Markdown conversion, (depending on implementation specifics) this could cause an "extra" blank line to be printed between elements. But I don't think that really matters, does it? What about a more extreme case?
```
##
##
##
##
##

version 1.1
```
Should any of those lines get a lint?
Similar case I just thought of:
```
##
##
##
## Those are a bunch of meaningless blank(ish) lines.
## Are they a problem worth linting?

version 1.0
```

a-frantz: I think both cases are harmless. They aren't ideal, but I don't think they're worth the hassle to lint. But my convictions here are weak.

wdl-lint/tests/lints/preamble-comment-after-version/source.errors
peterhuene: I should be able to remove it.

wdl-lint/tests/lints/missing-comment-space/source.wdl
peterhuene: Personally, I think the lint is prescriptive enough; if the blank lines do end up being rendered differently in documentation, I would err on the side of leaving it be and hope the user has a reason for having them.

At the very least, the fact that the lines are all `## ...` make it look like a consistent and intentional preamble.

a-frantz: Yea sounds good. No change needed!

a-frantz (approved):
Looks great

PR #73, opened 2024-06-10
feat: port the `MatchingParameterMeta` rule to `wdl-lint`.

peterhuene:
This commit ports the "matching parameter meta" rule to `wdl-lint`.

      changes (when appropriate).

### DIFF
--- a/wdl-lint/src/v1/matching_parameter_meta.rs
+++ b/wdl-lint/src/v1/matching_parameter_meta.rs
@@ -50,7 +50,7 @@ fn extra_param_meta(parent: &TaskOrWorkflow, extra: &str, span: Span) -> Diagnos
         TaskOrWorkflow::Workflow(w) => ("workflow", w.name()),
     };
 
-    Diagnostic::warning(format!(
+    Diagnostic::note(format!(
         "{context} `{parent}` has an extraneous parameter metadata key named `{extra}`",
         parent = parent.as_str(),
     ))
--- a/wdl-lint/src/v1/matching_parameter_meta.rs
+++ b/wdl-lint/src/v1/matching_parameter_meta.rs
@@ -1,0 +1,180 @@
+//! A lint rule for matching parameter metadata.
+
+use std::collections::HashMap;
+
+use wdl_ast::experimental::v1::InputSection;
+use wdl_ast::experimental::v1::ParameterMetadataSection;
+use wdl_ast::experimental::v1::TaskDefinition;
+use wdl_ast::experimental::v1::TaskOrWorkflow;
+use wdl_ast::experimental::v1::Visitor;
+use wdl_ast::experimental::v1::WorkflowDefinition;
+use wdl_ast::experimental::AstToken;
+use wdl_ast::experimental::Diagnostic;
+use wdl_ast::experimental::Diagnostics;
+use wdl_ast::experimental::Span;
+use wdl_ast::experimental::VisitReason;
+
+use super::Rule;
+use crate::Tag;
+use crate::TagSet;
+
+/// The identifier for the matching parameter meta rule.
+const ID: &str = "MatchingParameterMeta";
+
+/// Creates a "missing param meta" diagnostic.
+fn missing_param_meta(parent: &TaskOrWorkflow, missing: &str, span: Span) -> Diagnostic {
+    let (context, parent) = match parent {
+        TaskOrWorkflow::Task(t) => ("task", t.name()),
+        TaskOrWorkflow::Workflow(w) => ("workflow", w.name()),
+    };
+
+    Diagnostic::warning(format!(
+        "{context} `{parent}` is missing a parameter metadata key for input `{missing}`",
+        parent = parent.as_str(),
+    ))
+    .with_rule(ID)
+    .with_label(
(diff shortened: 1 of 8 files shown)


wdl-lint/src/v1/matching_parameter_meta.rs
```diff
@@ -0,0 +1,180 @@
...
+            i.declarations().map(|d| {
+                let name = d.name();
+                (name.as_str().to_string(), name.span())
+            })
+        })
+        .collect();
+
+    let actual: HashMap<_, _> = param_meta
+        .iter()
+        .flat_map(|m| {
```
a-frantz, line 112: I _think_ using a `flat_map` here is a problem. Meta can be arbitrarily nested, but we only want to check the first level. An example:
```
task {
    input { String foo }
    parameter_meta {
        not_foo: {
            also_not_foo: {
                bar: {
                    foo: "I might be describing some different `foo`!"
                }
            }
        }
    }
}
```
I guess the above would throw 3 warnings for `not_foo`, `also_not_foo`, and `bar`? But it would permit `foo`. I think instead We might only want to warn on the first level, (so ignore `also_not_foo` and `bar` _and_ `foo`).

Also I could be reading the Rust wrong lol.

wdl-lint/src/v1/matching_parameter_meta.rs
```diff
@@ -0,0 +1,180 @@
...
+}
+
+/// Creates an "extra param meta" diagnostic.
+fn extra_param_meta(parent: &TaskOrWorkflow, extra: &str, span: Span) -> Diagnostic {
+    let (context, parent) = match parent {
+        TaskOrWorkflow::Task(t) => ("task", t.name()),
+        TaskOrWorkflow::Workflow(w) => ("workflow", w.name()),
+    };
+
+    Diagnostic::warning(format!(
```
a-frantz, line 53:
```suggestion
    Diagnostic::note(format!(
```
Per [#3](https://github.com/stjude-rust-labs/sprocket/issues/3) of sprocket, our team is actually out of step with idiomatic WDL. Current common practice is to document outputs in the parameter_meta. I think we should drop the level of "extra_param_meta" from `warning` to `note` because we are going against the grain with this lint.

In a future release we'll want configuration to permit `output` in parameter_meta as well as inputs (But I think we hold our position with a default behavior of warning on outputs)

Open to other ideas as well here. A bit of a tricky situation

wdl-lint/src/v1/matching_parameter_meta.rs
peterhuene: The code is only checking the top-level items in the parameter metadata.

The first `iter()` call is on the `param_meta` option, which will either yield the section or nothing.

As a result, the `flat_map` is effectively called at-most once, passing it the section. What's being flattened is the section's items and we don't look into the item's values, so it's only the top names that are iterated over.

I can add a test case to ensure we don't accidentally check for the nested names, though.

a-frantz: Great, no code change needed then. But another test case would be a good sanity check 👍

wdl-lint/src/v1/matching_parameter_meta.rs
peterhuene: A note level makes sense to me.

Additionally, the spec does mention that both inputs and outputs should be present in the `parameter_meta` section and they _must_ correspond to an input or output name, but doesn't appear to put a restriction on a name that might appear as _both_ an input and an output; are users expected to document that the name is both an input and an output in the same entry, even if they might have different meanings?

Also, "parameter" definitely is a weird name for _results_ of a task or workflow. To an outsider coming into WDL, `parameter_meta` seems bolted on to the language. I hope that a future version can be steered towards documentation comments or at least separate `parameter_meta` into two sections: one for parameters/inputs and one for results/outputs.

a-frantz: > but doesn't appear to put a restriction on a name that might appear as both an input and an output

Inputs and outputs are defined in the same namespace, so this case actually is covered. You can't define an output with the same name as an input.

> Also, "parameter" definitely is a weird name for results of a task or workflow. To an outsider coming into WDL, parameter_meta seems bolted on to the language. I hope that a future version can be steered towards documentation comments or at least separate parameter_meta into two sections: one for parameters/inputs and one for results/outputs.

As to this, I think our whole team is in agreement on that. However, our team appears to be a minority. I've tried to register a formal complaint [here](https://github.com/openwdl/wdl/issues/637) that has been largely ignored.

peterhuene: > Inputs and outputs are defined in the same namespace, so this case actually is covered. You can't define an output with the same name as an input.

Good point, I had forgotten about that as we're not at a stage yet where we'd be evaluating the WDL to catch such errors.

> As to this, I think our whole team is in agreement on that. However, our team appears to be a minority. I've tried to register a formal complaint https://github.com/openwdl/wdl/issues/637 that has been largely ignored.

I'm happy to learn I'm not alone in my opinion 😅

a-frantz (approved):

PR #75, opened 2024-06-10
feat: port the "mixed indentation" rule to `wdl-lint`.

peterhuene:
This commit ports the "mixed indentation" rule to `wdl-lint`.

The lint looks for mixing tabs and spaces at the start of a line within a command section.

      changes (when appropriate).


wdl-lint/src/v1/mixed_indentation.rs
```diff
@@ -0,0 +1,116 @@
...
+use wdl_ast::experimental::Span;
+use wdl_ast::experimental::VisitReason;
+
+use super::Rule;
+use crate::util::lines_with_offset;
+use crate::Tag;
+use crate::TagSet;
+
+/// The identifier for the mixed indentation rule.
+const ID: &str = "MixedIndentation";
```
a-frantz, line 18: Q: eventually we want to check for mixed indentation through the whole document, but that is relatively inconsequential when it's not inside the command text (just super ugly and weird, but harmless). Will that be implemented under this same rule? I don't think so, because the `explanation()` will have to be different. Or at least significantly rephrased and extended.

If my hunch is correct and we want command-section-`MixedIndentation` _and_ a **non**-command-section-`MixedIndentation`, naming will be awkward. Should this be renamed `CommandSectionMixedIndentation`? It's wordy. What would the "everywhere else" name be ruled? 🤔

wdl-lint/tests/lints/mixed-indentation/source.wdl
```diff
@@ -0,0 +1,25 @@
...
+
+version 1.1
+
+task test1 {
+    command <<<
+        this line is prefixed with ~{"spaces and has tailing mixed indentation"}  		
+		this line is prefixed with ~{"tabs"}
+      	this line is prefixed with mixed indentation
+        this line is prefixed with spaces and has trailing mixed indentation  		
+			   >>>
```
a-frantz, line 11: I think your code should properly handle this case, but I want to see an explicit test for this case
```
        this line is getting ~{
            if foo
            then "split across multiple lines"
            else "by a placeholder"
        } but is all one literal line in the command text
```

a-frantz: Follow-up Q: Does escaping a newline do anything to change the parsing? I _think_ it would just be treated as any other literal text, without any special handling. Can we also add a test case for that?
```
        this line won't receive \
 \t         special handling and should be a warning
```

wdl-lint/src/v1/mixed_indentation.rs
a-frantz: > I would favor a single rule that reports slightly different messages. Within the `command` block a `warning` and outside it a `note`.

I was thinking the same until I tried to come up with a decent `explanation`. Imagine someone gets the `note` version due to their meta section. (Somehow) it's not clear to them what's wrong so they follow the `sprocket explain MixedIndentation` instruction. Anything about the command section and its word stripping is irrelevant to that user.

IMO the best user experience will be with 2 different Rules that will be able to communicate the issue most clearly.

> Otherwise, we could do something simple like `NonCommandSectionMixedIndentation` and `CommandSectionMixedIndentation`. It's a bit wordy, but probably the simplest.

Yea I don't love this but IDK any better phrasing for it. In the absence of other suggestions, I think the `ID` for this should become `CommandSectionMixedIndentation`.

wdl-lint/src/v1/mixed_indentation.rs
```diff
@@ -0,0 +1,116 @@
...
+                            match c {
+                                ' ' => spaces = true,
+                                '\t' => tabs = true,
+                                _ => break,
+                            }
+
+                            len += 1;
+                        }
+
+                        // If both spaces and tabs were present, the indentation is mixed
```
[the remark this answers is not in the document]
a-frantz: The language is a bit vague to my reading: https://github.com/openwdl/wdl/blob/main/versions/1.1/SPEC.md#stripping-leading-whitespace

I can understand both @peterhuene 's interpretation and @adthrasher 's. I think it's going to depend on how each engine interprets it? In which case we should err on the side of strictness. So having one line with `spaces` and one with `tabs` should be a `warning`, even if they are consistent _within_ each line.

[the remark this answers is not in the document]
adthrasher: Thanks for linking that section @a-frantz. I don't recall reading it previously, but that explains how mixed indentation can change the interpretation of the command block. Independently of parsing, I think the linter should have a rule (either here or separately) for `Consistency` that flags changes in indentation style.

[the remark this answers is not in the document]
a-frantz: > I think the linter should have a rule (either here or separately) for Consistency that flags changes in indentation style.

My fully detailed thoughts on proper indentation can be found in #11 😅 I devoted a fair amount of text to the topic lmao.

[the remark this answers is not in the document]
a-frantz: Or not actually? I can't find the rule I thought I had written. Well I have thoughts, I just apparently haven't written them down anywhere.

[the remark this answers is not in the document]
adthrasher: I had checked the `indentation` rule but the text there was rather sparse. I was just about to go searching to see if I could find it.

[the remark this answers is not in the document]
a-frantz: I really thought that the specification for how indents should be used was written about in that doc. I think I'm remembering spending a lot of time on `expression_spacing` and confusing that for indentation rules, as `expression_spacing` does cover some indentation specs.

wdl-lint/src/v1/mixed_indentation.rs
peterhuene: Sounds like we should rename this rule to `CommandSectionMixedIndentation` and follow-up later with a `MixedIndentation` rule that looks at the document's whitespace instead (and logs a note diagnostic).

Perhaps we should also tag this rule with correctness to justify it using a warning level whereas the new `MixedIndentation` rule isn't about correctness, just style/spacing, and therefore logs a note?

wdl-lint/tests/lints/mixed-indentation/source.wdl
peterhuene: I'll add a test case for your first comment.

For your second comment, I think that might be a bug; I'll add a test case for it and fix.

wdl-lint/src/v1/mixed_indentation.rs
a-frantz: All sounds good to me @peterhuene . I like calling _the other_ rule just `MixedIndentation`. That's better than `NonCommand...`. Agree to add `Correctness`. And why don't we remove `Style`?

wdl-lint/tests/lints/mixed-indentation/source.wdl
peterhuene: I pushed up a change to the test that includes both test cases and fixes the bug in line continuations.

wdl-lint/src/v1/mixed_indentation.rs
peterhuene: I've added a `Correctness` tag and removed `Style` and renamed the rule to `CommandSectionMixedIndentation`.

We can follow-up with a less-severe `Style` tagged rule named `MixedIndentation` that looks only at document whitespace in a future PR.

wdl-lint/src/v1/mixed_indentation.rs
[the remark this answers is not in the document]
peterhuene: I'll change to a single diagnostic that labels the command section, and perhaps a secondary label for the first mixed whitespace encountered?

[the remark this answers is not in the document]
a-frantz: I'm not sure what a secondary label will render like, but it sounds like a good idea. Maybe the label says `was indented with {spaces/tabs} up until this character`?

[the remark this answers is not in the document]
peterhuene: We can do that.

wdl-lint/tests/lints/command-mixed-indentation/source.wdl
```diff
@@ -0,0 +1,39 @@
...
+
+task test2 {
+    command {
+        this line is ~{
+		    if true
+		    then "split across multiple lines with mixed indentation"
+		    else "by a placeholder"
+	    } but is all one literal line in the command text
+        this line has a continuation \
+ 		   and should not be a warning
```
a-frantz, line 31: I think this _should_ be a warning. I'm taking the perspective of a user. If I write the following WDL:
```
command {
    my-tool \
    arg1 \
    arg2 \
    arg3
}
```
I would expect all those indents of size 4-spaces to be stripped from my command before execution. If I expect them to be stripped, it follows that they can't be mixed.

peterhuene: My understanding is that line continuation stripping is different from command whitespace stripping, though. It should be stripping _all_ leading whitespace on the next line until the first non-whitespace character, so it doesn't matter if it's mixed or not.

Leading whitespace in a command line is stripped by _common_ whitespace, which would then necessitate you don't mix the whitespace characters.

a-frantz: Does that come from the spec? What about the following WDL:
```
command {
    my-tool \
        arg1
}
```
I would (perhaps naively) assume it would execute the following command text:
```

my-tool \
    arg1
```

peterhuene: The spec calls `\<newline>` a continuation, but that's it.

If it's anything like a line continuation in other programming languages, it should treat that as `my-tool arg1` with no newline or leading whitespace.

a-frantz: Here's where I think WDL is different. I don't think the line continuation should be parsed by the WDL engine (or us). It should be passed along as part of the literal command text. If the spec doesn't say anything about how to parse line continuations, I think that's because they aren't meant to be parsed.

Check out this `command` file produced by `miniwdl`:
```
$ cat SJBALL034267_D1.WholeGenome/_LAST/call-multiqc/command

        set -euo pipefail

        # set ENV variables for `multiqc`
        export LC_ALL=C.UTF-8
        export LANG=C.UTF-8

        echo "/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.ValidateSamFile.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.flagstat.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.instrument.tsv
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.readlength.tsv
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.encoding.tsv
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.endedness.tsv
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome_fastqc.zip
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.CollectAlignmentSummaryMetrics.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.CollectInsertSizeMetrics.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.QualityScoreDistribution.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.kraken2.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.whole_genome.mosdepth.summary.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.whole_genome.mosdepth.global.dist.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.global_PHRED_scores.tsv
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.exon.mosdepth.summary.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.CDS.mosdepth.summary.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.UTR.mosdepth.summary.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.exon.mosdepth.region.dist.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.CDS.mosdepth.region.dist.txt
/mnt/miniwdl_task_container/work/_miniwdl_inputs/0/SJBALL034267_D1.WholeGenome.UTR.mosdepth.region.dist.txt" > file_list.txt

        # --strict is too strict. It causes errors due
        # to how our config adds 'custom-content' to the report.
        # Leaving this here as a warning not to try putting it back.
        # --require-logs might be useful at some point, but as of now,
        # it would cause errors. It could replace the check currently
        # run after multiqc is finished.
        # TODO: lots of other options to consider supporting.
        multiqc -v \
            --no-ansi \
            -c /mnt/miniwdl_task_container/work/_miniwdl_inputs/0/multiqc_config_hg38.yaml \
            --file-list file_list.txt \
            -o SJBALL034267_D1.WholeGenome.multiqc

        if [ ! -d SJBALL034267_D1.WholeGenome.multiqc ]; then
            >&2 echo "MultiQC didn't find any valid files!"
            exit 1
        fi

        tar -czf SJBALL034267_D1.WholeGenome.multiqc.tar.gz SJBALL034267_D1.WholeGenome.multiqc

```

a-frantz: Which actually hasn't been properly whitespace stripped. Could be a problem with our code though.

a-frantz: (PSA @adthrasher , it looks like miniwdl is not consistently stripping our whitespace as expected. I haven't identified the exact WDL constructions that cause it, but it's only a problem for some of our commands. Not a big issue as these commands aren't whitespace sensitive, but we should report to miniwdl just to be thorough)

peterhuene: I see. It's really bizarre to me that it would call it out as a line continuation, though, if it didn't want engines to do something with it.

If they _hadn't_ called out `\<newline>` as a line continuation, it would have had the exact same effect: not an escape sequence and therefore a _literal backslash_ at the end of the line to be interpreted by _the shell_ as a line continuation.

At any rate, I'll chalk this up as vague warning in the spec (if it means it to be a line continuation _in the shell script_, it should probably say so) and fix the rule.

🤷

wdl-lint/src/v1/mixed_indentation.rs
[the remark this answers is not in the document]
peterhuene: How's this:

```
warning: mixed indentation within a command [rule: CommandSectionMixedIndentation]
   ┌─ tests/lints/command-mixed-indentation/source.wdl:18:3
   │
 6 │     command <<<
   │     ------- this command section uses both tabs and spaces in leading whitespace
   ·
18 │                >>>
   │         ^^^^ indented with spaces until this tab
   │
   = fix: use the same whitespace character for indentation
```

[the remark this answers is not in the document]
a-frantz: Looks great!

wdl-lint/tests/lints/command-mixed-line-cont/source.wdl
```diff
@@ -0,0 +1,21 @@
...
+ 		   and should be a warning
+    >>>
+
+    runtime {}
+}
+
+task test2 {
+    command {
+        this line has a continuation \
+ 		   and should not be a warning
```
a-frantz, line 17:
```suggestion
 		   and should be a warning
```

a-frantz (approved):

PR #76, opened 2024-06-11
feat: migrate `wdl-gauntlet` to the new parser implementation.

peterhuene:
This commit removes the use of the existing parser implementation in favor of the new implementation in `wdl-gauntlet`.

As a result, `Arena.toml` and `Gauntlet.toml` have been refreshed; in the case of `Arena.toml`, there's a ~10X increase in logged diagnostics due to the new parser implementation successfully parsing files that the existing parser implementation incorrectly parsed with errors.

Also added some wall-clock timing for the time spent analyzing a source file and an aggregate duration displayed at the end of gauntlet's output. From the before and after comparisons, the new parser implementation appears to be at least 20X faster in analyzing files, likely due to using `logos` for lexing and the new parser implementation does not backtrack like `pest` does.

Also added configuration for ignoring specific files so that `task-templates.wdl` can be excluded as it isn't expected to be valid WDL (it has non-WDL placeholders in it to serve as a template).

Once this merges and we're in consensus, removal of the existing parser implementation can begin.

      changes (when appropriate).


Gauntlet.toml
```diff
@@ -46,658 +44,309 @@ commit_hash = "ee08634f28810e5d6fd1a904fc83f4e67821550e"
...
+document = "broadinstitute/warp:/tasks/broad/GermlineVariantDiscovery.wdl"
+message = "/tasks/broad/GermlineVariantDiscovery.wdl:65:32: error: expected string, but found integer"
+
+[[diagnostics]]
+document = "broadinstitute/warp:/tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl"
+message = "/tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl:814:27: error: expected string, but found integer"
+
+[[diagnostics]]
+document = "broadinstitute/warp:/tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl"
+message = "/tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl:866:27: error: expected string, but found integer"
```
a-frantz, line 344: I think this is a bug on our part. Or I'm confused. The below code looks like valid WDL to me.

https://github.com/broadinstitute/warp/blob/develop/tasks/broad/UltimaGenomicsWholeGenomeGermlineTasks.wdl#L866

Arena.toml
a-frantz, line 1: IMO there's too much noise coming from `ENCODE-DCC/chip-seq-pipeline2` to do a good job looking over the Arena results here. I think we should remove that repo. And hopefully replace it with something that emits a more manageable number of diagnostics?

Gauntlet.toml
peterhuene: According to the spec:

> The result of evaluating an expression in a placeholder must ultimately be converted to a string in order to take the place of the placeholder in the command script

...

> Expression placeholder options are `option="value"` pairs that precede the expression within an expression placeholder and customize the interpolation of the WDL value into the containing string expression.

Making it seem that the option's value must be a quoted string as the result of the interpolation will always be a string anyway.

This makes sense for `sep` as it's effectively the separator argument to the `sep` function; it also makes sense for `true` and `false` as they're just replacement strings for those values respectively.

It goes on to say:

> If an expression within a placeholder evaluates to None, then the placeholder is replaced by the empty string.

which makes sense as well (i.e. the default for the `default` placeholder option is `""`).

However, it adds this about the `default` option specifically:

> **The type of the default value must match the type of the expression**

The last one makes absolutely no sense; the default value shouldn't need match the type of the expression; it's simply what is used _instead of the empty string_ when the expression evaluates to `None`. 

Despite the contradictory elements of the spec, I think accepting anything but a string for the `default` option's value should not be legal, otherwise a parser would have to accept, at the minimum, a subset of _literal expression_ at that position, which seems incorrect and definitely not spelled out in the spec like they do for metadata values.

Given that placeholder options are deprecated and there were only a few occurrences of not using a quoted value in the gauntlet run, I was thinking the current behavior and a "c’est la vie" approach would be best here.

Arena.toml
peterhuene: I'd be onboard with that; the existing parser implementation doesn't parse a number of their files correctly, so we weren't getting any value of them in arena until this point anyway.

Gauntlet.toml
peterhuene: In fact, I think we should be linting the use of placeholder options as part of the "deprecation" rule.

peterhuene: FWIW, an official parser does accept a [string or number](https://github.com/openwdl/wdl-parsers/blob/main/antlr4/1.1/WdlV1_1Parser.g4#L50) for the `default` placeholder option; I wonder if that's simply because it was observed in the wild rather than a result of interpreting the spec. If those are the only two accepted literal expressions at that position, the spec should clearly say so.

A literal interpretation of the spec (specifically the part about `the type of the default value must match the type of the expression`) would require `None` to be accepted here too because if the default value's type must match and the type of placeholder expression is required to be *optional*, then the default value's type must also be *optional*, allowing for `None` as a value. Obviously that would be a useless `default` option value.

At any rate, I'm also happy to fix the parser to accept numbers for the default option, despite it being converted to string _anyway_ as the result of the interpolation.

peterhuene: Opened https://github.com/stjude-rust-labs/wdl/issues/77 to track this.

Arena.toml
```diff
@@ -1,358 +1,216 @@
-version = "v1"
```
adthrasher, line 1: Do we need the version identifier? I don't know what the original intent was.

Arena.toml
```diff
@@ -1,358 +1,216 @@
...
 document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
-message = "[v1::W003::[Completeness]::Medium] extraneous parameter meta within task: crai (203:5-203:9)"
+message = "/ww-fastq-to-cram.wdl:202:5: note: task `MergeBamsToCram` has an extraneous parameter metadata key named `cram` [rule: MatchingParameterMeta]"
 
-[[concerns]]
-kind = "LintWarning"
+[[diagnostics]]
 document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
-message = "[v1::W003::[Completeness]::Medium] extraneous parameter meta within task: cram (202:5-202:9)"
+message = "/ww-fastq-to-cram.wdl:203:5: note: task `MergeBamsToCram` has an extraneous parameter metadata key named `crai` [rule: MatchingParameterMeta]"
```
peterhuene, line 32: I'm going to remove the redundant relative path in `message` and just use the file name in the baselines.

Arena.toml
```diff
@@ -1,358 +1,216 @@
...
-message = "[v1::W001::[Spacing, Style]::Low] trailing space (133:46)"
+commit_hash = "a44b18a30fd6192af713fded4bc824f6e77f6ff9"
+filters = ["/template/task-templates.wdl"]
 
-[[concerns]]
-kind = "LintWarning"
+[[diagnostics]]
 document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
-message = "[v1::W001::[Spacing, Style]::Low] trailing space (182:63)"
+message = "/ww-fastq-to-cram.wdl:129:5: note: task `FastqToUnmappedBam` has an extraneous parameter metadata key named `unmapped_bam` [rule: MatchingParameterMeta]"
```
adthrasher: @a-frantz What do you think of having the rule at the end of the line? I don't feel strongly enough to suggest we change it, but it feels more natural at the beginning with the file name and location.

Arena.toml
a-frantz: Nice. I was thinking about asking for that but wasn't sure how big an ask it would be.

peterhuene: Of course, I comment on one where the file is at the root of the repo, so it's not terribly good example 🤦 

Many other files are not at the root, so repeating the path just makes it overly verbose.

Arena.toml
a-frantz: Lol your suggestion is how @peterhuene had it initially and I asked for a change to the back 😂 Though I am far from tied to this message template and would be happy to revisit.

a-frantz: Give me a second to dig up our original thread.

a-frantz: https://github.com/stjude-rust-labs/wdl/pull/68#discussion_r1630375699

Arena.toml
peterhuene: The `version` here wasn't the configuration file format version, but the WDL version to parse with.

Since the new parser implementation responds to what is specified in the document and doesn't require specifying the version upfront, this configuration value is no longer needed.

Arena.toml
adthrasher: Ah, I figured I had missed something with it now at the end of the line. As I said, I don't feel strongly enough to disagree with the placement.

Arena.toml
adthrasher: Thanks for the clarification. I'd rather parse it out of the file themselves, so a clear improvement.

Arena.toml
a-frantz: I unresolved your resolve @adthrasher lol. Now that you've pointed it out I'm second guessing. @peterhuene feel free to weigh in.

My initial concern was that `<severity>[<ID>]` wasn't clear that ID was the "hey look this value up in the documentation". You mentioned it feeling "natural" to have the rule name near the front and I think you're correct.

Current:
```
message = "read_group.wdl:32:12: warning: struct member name `ID` is not snake_case [rule: SnakeCase]"
```
New?
```
message = "read_group.wdl:32:12: warning[SnakeCase]: struct member name `ID` is not snake_case"
```

peterhuene: I'm good with switching it back as I didn't have an opinion one way or the other originally.

I'll revert that and update the baselines.

adthrasher: I like your (@a-frantz) updated format. That feels natural.

peterhuene: I've pushed the change.

a-frantz:
Was kicking tires. Noticed a regression in the `PreambleComments` rule. Alerting repo was `aws-samples/amazon-omics-tutorials`. They use our concept of preamble comments but _after_ the version statement. This results in one diagnostic per line of a long block. One of the changes included in #61 was to report the entire span of problematic comments as one `concern` (as it was in the old parser).

Can we update the `PreambleComments` rule to report a multiline span? That should cut down the number of diagnostics coming out of that AWS repo, making them a candidate to add to `Arena.toml` (maybe, I haven't done a `wc -l` to see how many new ones they would add)

a-frantz:
Another very similar case found in this files preamble: https://github.com/biowdl/tasks/blob/develop/deconstructsigs.wdl

We should similarly be reporting a multiline span instead of one per line

peterhuene:
@a-frantz good catch. That should be easily fixed. Do you want me to include the fix in this PR?

a-frantz:
Sure, if you say it's an easy one!

peterhuene:
Question:

```wdl
# BAD
# BAD



# BAD
```

Is that one diagnostic about the incorrect comments or two? Put another way, are the comments consecutive if they have only a single newline in separating whitespace or any number of blank lines inbetween?

a-frantz:
> Question:
> 
> ```wdl
> # BAD
> # BAD
> 
> 
> 
> # BAD
> ```
> 
> Is that one diagnostic about the incorrect comments or two? Put another way, are the comments consecutive if they have only a single newline in separating whitespace or any number of blank lines inbetween?

IMO one per "chunk" of trivia. Does that make sense?

a-frantz:
Last comment I'll make 😅 
Can we add `getwilds/ww-vc-trio` and `getwilds/ww-star-deseq2` to Arena? Both are very small repos and add a limited number of diagnostics. Gives us a little more to work with without being overwhelming. And `ww-vs-trio` has one of the preamble cases I pointed out, so a good test case.

peterhuene:
@a-frantz pushed up both changes.

a-frantz (approved):
Awesome work!

PR #89, opened 2024-06-17
feat: implement the `ImportPlacement` rule.

peterhuene:
**Rule Name**: `ImportPlacement`

This commit implements the `ImportPlacement` rule which warns if an import comes after a non-import top-level item in the document.

Additionally, this fixes the `PreambleWhitespace` rule to check for a blank line following the version statement.

Also fixed:

* Last whitespace in the file was attaching as a child to the last node in the document instead of as top-level whitespace.
* The CST for an "unparsed" file (i.e. missing version statement or unsupported version) was missing trivia before the version statement and the unparsed token span was incorrect.

Rule specific checks:

- [x] You have added the rule to the `rules()` function in `wdl-lint/src/v1.rs`.
      is implemented.
      (`Arena.toml`).

### DIFF
--- a/wdl-lint/src/v1/import_placement.rs
+++ b/wdl-lint/src/v1/import_placement.rs
@@ -24,7 +24,10 @@ fn misplaced_import(span: Span) -> Diagnostic {
     Diagnostic::warning("misplaced import")
         .with_rule(ID)
         .with_highlight(span)
-        .with_fix("move this import so that it comes directly after the version statement")
+        .with_fix(
+            "move this import so that it comes after the version statement but before any \
+             document items",
+        )
 }
 
 /// Detects incorrect import placements.
@@ -37,7 +40,7 @@ impl Rule for ImportPlacementRule {
     }
 
     fn description(&self) -> &'static str {
-        "Ensures that imports are placed immediately following the version statement."
+        "Ensures that imports are placed between the version statement and any document items."
     }
 
     fn explanation(&self) -> &'static str {
--- a/wdl-lint/src/v1/import_placement.rs
+++ b/wdl-lint/src/v1/import_placement.rs
@@ -1,0 +1,118 @@
+//! A lint rule for import placements.
+
+use wdl_ast::v1::ImportStatement;
+use wdl_ast::v1::StructDefinition;
+use wdl_ast::v1::TaskDefinition;
+use wdl_ast::v1::Visitor;
+use wdl_ast::v1::WorkflowDefinition;
+use wdl_ast::AstNode;
+use wdl_ast::Diagnostic;
+use wdl_ast::Diagnostics;
+use wdl_ast::Span;
+use wdl_ast::ToSpan;
+use wdl_ast::VisitReason;
+
+use super::Rule;
+use crate::Tag;
+use crate::TagSet;
+
+/// The identifier for the import placement rule.
+const ID: &str = "ImportPlacement";
+
+/// Creates a "misplaced import" diagnostic.
+fn misplaced_import(span: Span) -> Diagnostic {
+    Diagnostic::warning("misplaced import")
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_fix("move this import so that it comes directly after the version statement")
+}
+
+/// Detects incorrect import placements.
+#[derive(Debug, Clone, Copy)]
+pub struct ImportPlacementRule;
+
+impl Rule for ImportPlacementRule {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Ensures that imports are placed immediately following the version statement."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "All import statements should follow the WDL version declaration with one empty line \
+         between the version and the first import statement."
+    }
+
+    fn tags(&self) -> TagSet {
+        TagSet::new(&[Tag::Clarity])
+    }
+
+    fn visitor(&self) -> Box<dyn Visitor<State = Diagnostics>> {
+        Box::new(ImportPlacementVisitor::default())
+    }
+}
+
+/// Implements the visitor for the import placement rule.
+#[derive(Default)]
+struct ImportPlacementVisitor {
+    /// Whether or not an import statement is considered invalid.
+    invalid: bool,
+}
+
+impl Visitor for ImportPlacementVisitor {
+    type State = Diagnostics;
+
+    fn import_statement(
+        &mut self,
+        state: &mut Self::State,
+        reason: VisitReason,
+        stmt: &ImportStatement,
+    ) {
+        if reason == VisitReason::Exit {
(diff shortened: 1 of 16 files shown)


wdl-lint/src/v1/import_placement.rs
```diff
@@ -0,0 +1,118 @@
...
+
+/// The identifier for the import placement rule.
+const ID: &str = "ImportPlacement";
+
+/// Creates a "misplaced import" diagnostic.
+fn misplaced_import(span: Span) -> Diagnostic {
+    Diagnostic::warning("misplaced import")
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_fix("move this import so that it comes directly after the version statement")
```
a-frantz, line 27: Does `directly after` contradict the requirement for a blank line? When I read this "fix" message, I think the solution would be:
```
version 1.0
import "foo"
```

wdl-lint/src/v1/import_placement.rs
```diff
@@ -0,0 +1,118 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct ImportPlacementRule;
+
+impl Rule for ImportPlacementRule {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Ensures that imports are placed immediately following the version statement."
```
a-frantz, line 40: Same question for `immediately` as I had for `directly`.

wdl-lint/src/v1/preamble_whitespace.rs
a-frantz, line 1: IMO linting whitespace _after_ the preamble is OOS for the "PreambleWhitespace" rule. In the `explanation` of this rule we explicitly define what part of the document is considered the preamble, it seems odd to then create diagnostics _from the preamble rule_ for WDL which isn't part of the preamble.

Hate asking this after the implementation is complete, but could this WS check be moved to the `ImportPlacement` rule? IMO it fits better there, where the `explanation` can say something like `Imports should be separated from the version statement by a blank line`.

wdl-lint/tests/lints/import-placements/source.wdl
```diff
@@ -0,0 +1,17 @@
+## This is a test of import placements.
+
+version 1.1
+
+import "good"
+
+import "good"
+
+import "good"
```
a-frantz, line 9: Not everyone on the team agrees with me on this (specifically I think Clay registered a disagreement), but IMO there shouldn't be blank lines _between_ import statements. I would think that check would belong within the rule titled "ImportPlacement". Are we planning on add that check either here or as part of another rule?

wdl-lint/src/v1/import_placement.rs
peterhuene: Agreed, I'll remove the "directly" from the message.

wdl-lint/src/v1/import_placement.rs
peterhuene: How about `Ensures that imports are placed between the version statement and any document items`?

wdl-lint/src/v1/import_placement.rs
peterhuene: How about:

```suggestion
        .with_fix("move this import so that it comes after the version statement but before any document items")
```

wdl-lint/src/v1/import_placement.rs
a-frantz: That sounds good 👍

wdl-lint/src/v1/import_placement.rs
a-frantz: LGTM

wdl-lint/src/v1/preamble_whitespace.rs
peterhuene: I personally think the "a blank line must come after the preamble" requirement should be part of the preamble whitespace rule, even if the blank line is _technically_ not part of the preamble itself; the rule is still concerned with _separating_ the preamble from the rest of the document, which feels like it's in scope to me.

It should be enforced even if there are no imports in the document too, so putting it in a "import placement" or "import whitespace" rule seems odd to me too.

wdl-lint/tests/lints/import-placements/source.wdl
peterhuene: I was thinking it'd be its own rule; I can see wanting imports to be at the top of the document, but perhaps allowing whitespace between them.

I assume such a rule would allow comments too, such as:

```wdl
version 1.1

# Foo
import "good"
import "good"
# Bar
import "good"

import "above whitespace is bad"
```

a-frantz: Yes that's exactly how I envisioned the rule. I think Clay's case of wanting XYZ imports separate from ABC imports could be accomplished with a comment labelling the two groups rather than whitespace. I mainly want to avoid:
```
version 1.0

import "fine"

import "too_much_WS"

import "no_excuse_for_this"

import "this_is_bad"
```

wdl-lint/src/v1/preamble_whitespace.rs
a-frantz: > I personally think the "a blank line must come after the preamble" requirement should be part of the preamble whitespace rule, even if the blank line is technically not part of the preamble itself; the rule is still concerned with separating the preamble from the rest of the document, which feels like it's in scope to me.

That does make sense when it's explained like that. I think this is fine then, but the `explanation()` text should be updated to say `a blank line must come after the preamble`.

wdl-lint/tests/lints/import-placements/source.wdl
peterhuene: Great, I think having a `ImportWhitespace` (or similarly named) rule would make sense as a follow-up PR. Does that work for you?

a-frantz: Sure, I think it's fine to be separate rules. Could see it either way, but don't have a preference

wdl-lint/tests/lints/missing-blank-line-after-version/source.wdl
```diff
@@ -0,0 +1,6 @@
+## This is a test of a missing blank line following the version statement.
+
+version 1.1
```
a-frantz, line 3: Maybe this is OOS, but where would the following case be caught?
```wdl
version 1.1




import "foo"
```

peterhuene: It's caught by the existing `Whitespace` rule:

```
warning[Whitespace]: more than one blank line in a row
>  ┌─ tests/lints/import-placements/source.wdl:5:1
>  │  
>5 │ ╭ 
>6 │ │ 
>7 │ │ 
>8 │ │ 
>9 │ │ import "good"
>  │ ╰^
>  │  
>  = fix: remove the unnecessary blank lines

a-frantz (approved):

PR #91, opened 2024-06-18
feat: implement unique name validation.

peterhuene:
This commit implements a validation visitor for ensuring that certain names are unique within certain scopes.

Ensures that:

* Import namespaces are unique.
* Workflow names are unique.
* Task names are unique.
* Struct names are unique between declarations and import aliases.
* Struct member names are unique.
* Declaration names are unique within certain scopes.

      changes (when appropriate).


a-frantz (commented):
Two thoughts:
1. Eventually we want a _lint_ rule that will cover cases _not_ covered by this rule. I write up an example in #11 , but the idea is that some repetitive names are permitted syntactically but should be discouraged as they can be confusing. Just pointing this out as it's related. Not really a question.
2. This will have to be revisited and updated once we implement imports, correct?

peterhuene:
> This will have to be revisited and updated once we implement imports, correct?

That's correct; this will not catch conflicts with struct names that are implicitly introduced with an `import`, only those explicitly aliased for now.

When we do "see through the import", we can also have better diagnostics around conflicts with imports to suggest aliasing the names.

wdl-ast/src/v1/validation/names.rs
```diff
@@ -0,0 +1,237 @@
...
+///
+/// Ensures that the following names are unique:
+///
+/// * Task names.
+/// * Struct names from struct declarations and import aliases.
+/// * Struct member names.
+/// * Declarations and scatter variable names.
+///
+/// Note that it does not check for duplicate workflow names as it's already a
+/// validation error to have more than one workflow in a document.
```
a-frantz, line 69: Isn't it a problem to have a task and workflow with the same name? e.g.
```wdl
version 1.1
task foo {}
workflow foo {}
```
If that's in `foo.wdl` and in `bar.wdl` I have
```wdl
version 1.1
import foo.wdl

workflow bar { call foo.foo }
```
Does the second `foo` call the workflow or the task?

wdl-ast/tests/validation/duplicate-decls/source.errors
```diff
@@ -0,0 +1,108 @@
...
+
+error: conflicting declaration name `nested`
+   ┌─ tests/validation/duplicate-decls/source.wdl:64:13
+   │
+57 │                 Int nested = 0
+   │                     ------ first declaration with this name is here
+   ·
+64 │         Int nested = 0
+   │             ^^^^^^ this conflicts with a previous declaration of the same name
+
```
a-frantz, line 108: I understand what's happening here, but it's a confusing situation. The `first declaration` with the name `nested` changes. Is there some way we can rephrase this so that these 2 errors back-to-back aren't contradicting each other?

wdl-ast/src/v1/validation/names.rs
peterhuene: Given the example to the `name_collision` lint in the baseline lint rules discussion, it sounded like the names only had to be unique for their "kind"; the spec also isn't very clear on it either.

For what it's worth, `miniwdl` does check that the name is unique for a top-level item in the document (including struct names); I personally think that what `miniwdl` is checking makes the most sense and I was quite confused when I first read the description of the proposed `name_collision` rule.

Certainly the example you have there does make it seem like that would be ambiguous to an execution engine and require that the task and workflow have unique names.

wdl-ast/tests/validation/duplicate-decls/source.errors
peterhuene: I can probably change the diagnostic for when it conflicts with a scatter variable to distinguish them.

wdl-ast/src/v1/validation/names.rs
peterhuene: Also, I realized that we should be validating the namespace names being introduced by import statements as being unique as well (either the implicit or explicit name).

If there's no explicit name and it conflicts, then we can suggest using an `as` clause on the import in the diagnostic.

a-frantz: > Given the example to the name_collision lint in the baseline lint rules discussion, it sounded like the names only had to be unique for their "kind"; the spec also isn't very clear on it either.

Yes, I just reviewed the example I gave in that document and it is incorrect. Whoops 😬. 

> For what it's worth, miniwdl does check that the name is unique for a top-level item in the document (including struct names);

So as we established above, a declared workflow must have a unique name amongst any declared tasks. But I think `miniwdl` is going a step beyond what is required by the SPEC. I think a struct name should be able to overlap with a workflow or task. (It should definitely be advised against, but should be syntactically clear to my understanding).

> Also, I realized that we should be validating the namespace names being introduced by import statements as being unique as well (either the implicit or explicit name).
>
> If there's no explicit name and it conflicts, then we can suggest using an as clause on the import in the diagnostic.

Sounds good 👍

peterhuene: So having a task and workflow with the same name does occur in the wild:

https://github.com/broadinstitute/gatk-tool-wdls/blob/master/wdls/ApplyBQSR.wdl

peterhuene:
While working on the code review feedback, I realized as I printed a diagnostic containing a label pointing at an import URI that the span of the URI's node included the whitespace between the `import` keyword and the URI.

To fix this, I changed how `start` works on the parser: if it's not the root node, we `peek` to ensure any trivia events are added to the event list _before_ starting the new node; thus the whitespace now becomes a sibling of the URI's node, so the span no longer contains it.

As a result of this change, a number of the CST baselines in the parser tests had to be updated.

Sorry for effectively reimplementing the PR after your previous review.

a-frantz:
> While working on the code review feedback, I realized as I printed a diagnostic containing a label pointing at an import URI that the span of the URI's node included the whitespace between the `import` keyword and the URI.
> 
> To fix this, I changed how `start` works on the parser: if it's not the root node, we `peek` to ensure any trivia events are added to the event list _before_ starting the new node; thus the whitespace now becomes a sibling of the URI's node, so the span no longer contains the leading whitespace.
> 
> As a result of this change, a number of the CST baselines in the parser tests had to be updated.
> 
> Sorry for effectively reimplementing the PR after your previous review.

No worries, it happens!

wdl-ast/src/v1/validation/names.rs
a-frantz: Unresolving because I have realized _another_ issue with my prior example (😬 I guess it was a _really_ bad example). Rewriting here for ease of reference:

`foo.wdl`
```wdl
version 1.1
task foo {}
workflow foo {}
```
`bar.wdl`
```wdl
version 1.1
import foo.wdl
workflow bar { call foo.foo }
```
So we've talked about the name collision on `foo` _within_ `foo.wdl` (exemplified by the ambiguous `call` in `bar.wdl`), and you've addressed that and the change looks great 👍 .

BUT, there is another name collision! In `bar.wdl` L2 introduces the namespace-`foo` into the scoped-namespace. Then in L3 `call foo.foo` tries to introduce the call-name-`foo` into the same scoped-namespace! This would have to be fixed in one of two ways (let's pretend `foo` resolves unambiguously):
```wdl
version 1.1
import foo.wdl as something_else
workflow bar { call something_else.foo }
```
OR
```wdl
version 1.1
import foo.wdl
workflow bar { call foo.foo as something_else }
```
Note that we should prefer the first option (it's easier to track aliases when they're in the import block than when they're in a workflow body).

a-frantz: > Unresolving because I have realized _another_ issue with my prior example (😬 I guess it was a _really_ bad example). Rewriting here for ease of reference:
> 
> `foo.wdl`
> 
> ```wdl
> version 1.1
> task foo {}
> workflow foo {}
> ```
> 
> `bar.wdl`
> 
> ```wdl
> version 1.1
> import foo.wdl
> workflow bar { call foo.foo }
> ```
> 
> So we've talked about the name collision on `foo` _within_ `foo.wdl` (exemplified by the ambiguous `call` in `bar.wdl`), and you've addressed that and the change looks great 👍 .
> 
> BUT, there is another name collision! In `bar.wdl` L2 introduces the namespace-`foo` into the scoped-namespace. Then in L3 `call foo.foo` tries to introduce the call-name-`foo` into the same scoped-namespace! This would have to be fixed in one of two ways (let's pretend `foo` resolves unambiguously):
> 
> ```wdl
> version 1.1
> import foo.wdl as foo_namespace
> workflow bar { call foo_namespace.foo }
> ```
> 
> OR
> 
> ```wdl
> version 1.1
> import foo.wdl
> workflow bar { call foo.foo as foo_call }
> ```
> 
> Note that we should prefer the first option (it's easier to track aliases when they're in the import block than when they're in a workflow body).

I asked @adthrasher to double check because I didn't feel confident about this, and I am wrong yet again 😬 Apparently I am really bad with conceptualizing namespaces lol.
`bar.wdl` is fine, it does have multiple things in scope named `foo` but they are accessed in different contexts and therefore they can be referred to unambiguously.
* `foo`: introduced by `import` is a task/workflow namespace
* `foo` introduced by `call foo.foo` is a declaration namespace

Those name collisions should probably be lint warnings, but they are ok in the context of this PR (validation-wise).

adthrasher: @a-frantz and I discussed this this morning on Slack and walked through some examples. My opinion is that we should be overzealous in using lint warnings to discourage name conflicts (even if permissible in the spec/execution engine). After all, we are attempting to create rules to encourage well-formed, readable WDL files. Name conflicts are difficult for a human reader to decipher.

wdl-ast/tests/validation/conflicting-call-names/source.errors
```diff
@@ -0,0 +1,103 @@
+error: conflicting call name `my_int`
+  ┌─ tests/validation/conflicting-call-names/source.wdl:7:10
+  │
+6 │     Int my_int = 0      # FIRST
+  │         ------ the declaration with the conflicting name is here
+7 │     call my_int         # OK
+  │          ^^^^^^ this conflicts with a declaration of the same name
+  │
+  = fix: add an `as` clause to the call to specify a different name
```
a-frantz, line 9: I keep getting this namespace stuff wrong, so I'm tagging @adthrasher as a double check.

But I think this validation check needs to be relaxed. This example here: any downstream usage of `my_int` will be unambiguous to the engine (but very confusing to readers).
- If the full identifier is `my_int` we are referring to the `Int`
- If the full identifier is `my_int.*` we are referring to the `call`

There's never an ambiguous situation arising from this "conflict", so I think it needs to be permitted.

adthrasher: Agreed, I think this is allowable. So it should be permitted by this validation check (though maybe flagged by a lint rule?).

peterhuene: Also `my_int.x` is not an identifier, it's an expression:

```
AccessExprNode
  NameRefNode
    Ident "my_int"
  Dot "."
  Ident "x"
```

An evaluator resolves the name reference to whatever value has that name; in a well-specified language, the `call` statement produces a value of a particular type and shouldn't be treated different from any other declaration.

a-frantz: @peterhuene that all makes sense to me. Is there anything resulting from this popping up in Gauntlet?

adthrasher: That's an excellent point @peterhuene. We should collect these examples and bring them up with the WDL maintainers. If our behavior matches `miniwdl`'s, there probably will be little resistance to making that behavior explicit in the spec.

peterhuene: > Is there anything resulting from this popping up in Gauntlet?

The change didn't require an update to `Gauntlet.toml`, so I don't think anyone is running into this validation error currently.

a-frantz (approved):

PR #92, opened 2024-06-20
feat(gauntlet): permalinks for each diagnostic

a-frantz:
Note: these permalinks could contain a line number for an even better user experience, but I couldn't figure that out 😅 

      changes (when appropriate).


Arena.toml
```diff
@@ -18,1099 +18,1374 @@ filters = ["/template/task-templates.wdl"]
 [[diagnostics]]
 document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
 message = "ww-fastq-to-cram.wdl:129:5: note[MatchingParameterMeta]: task `FastqToUnmappedBam` has an extraneous parameter metadata key named `unmapped_bam`"
+permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/7dd6c4c7c8c9fc6dcd4eb5ff8f3262cd26d5e7cc/ww-fastq-to-cram.wdl"
```
adthrasher, line 21: Should the line information be embedded in the permalink? I would think so.

a-frantz: From the PR description:

> Note: these permalinks could contain a line number for an even better user experience, but I couldn't figure that out 😅

So the answer is yes, I was just stumped on how to do that

peterhuene: An `#L<NUM>` anchor can be used on the URL to bring you to the line.

And I think we can utilize [line_index](https://docs.rs/codespan-reporting/latest/codespan_reporting/files/trait.Files.html#tymethod.line_index) to go from span start offset to line.

a-frantz: The problem I'm having is that all the `Diagnostics` get converted into plain `String`s [here](https://github.com/stjude-rust-labs/wdl/blob/feat/gauntlet-urls/wdl-gauntlet/src/lib.rs#L225). I would either have to retrieve the line numbers from the raw string (possible but messy) or rewrite significant portions of this code to pass around `Diagnostics` instead of `Strings`.

wdl-gauntlet/src/lib.rs
```diff
@@ -241,14 +241,14 @@ pub async fn gauntlet(args: Args) -> Result<()> {
             // As the list of diagnostics has been sorted by document identifier, do
             // a binary search and collect the matching messages
             let diagnostics = config.inner().diagnostics();
-            let doc_id_str = document_identifier.to_string();
+            let doc_id = document_identifier.to_string();
             let expected: IndexSet<String> = diagnostics
-                .binary_search_by_key(&doc_id_str.as_str(), |d| d.document())
+                .binary_search_by_key(&doc_id, |d| d.document().to_string())
```
peterhuene, line 246: Does `doc_id` need to be a string now that the field is an `Identifier`?

Could we not just pass `&document_identifier` and do `d.document()` in the closure?

wdl-gauntlet/src/lib.rs
```diff
@@ -337,8 +337,16 @@ pub async fn gauntlet(args: Args) -> Result<()> {
 
         for message in messages {
             diagnostics.push(config::inner::Diagnostic::new(
-                identifier.to_string(),
+                identifier.clone(),
                 message,
+                config
+                    .inner()
+                    .repositories()
+                    .get(&identifier.repository().clone())
```
peterhuene, line 345: Is the `clone` call here necessary?

peterhuene (commented):
Looks pretty good, sans any change to add line numbers to permalink.

Just a few questions.

wdl-gauntlet/src/document/identifier.rs
```diff
@@ -119,3 +115,26 @@ impl std::str::FromStr for Identifier {
         Ok(Self { repository, path })
     }
 }
+
+impl Serialize for Identifier {
```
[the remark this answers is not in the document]
a-frantz: Not a great answer, but my compiler was yelling at me so I did this and it worked? 😂 I'll revisit this to see if it was necessary.

wdl-gauntlet/src/lib.rs
a-frantz: Is there a relevant difference between doing `*d.document() == document_identifier` and `d.document() == &document_identifier`?

Regardless, good call that makes this section easier to read.

peterhuene: It makes no different between the two.

wdl-gauntlet/src/lib.rs
peterhuene: Oh sorry, repository already returns a reference, so perhaps just `.get(identifier.repository())`?

a-frantz: That seems to work! Thanks.

wdl-gauntlet/src/report.rs
```diff
@@ -16,14 +16,14 @@ pub struct UnmatchedStatus {
...
+    pub missing: IndexSet<(Option<usize>, String)>,
     /// The unexpected set of diagnostics.
     ///
     /// These are the diagnostics that were not in the configuration but
     /// were emitted by the parser/validator.
-    pub unexpected: IndexSet<String>,
+    pub unexpected: IndexSet<(Option<usize>, String)>,
     /// The set of all diagnostics that were emitted.
-    pub all: IndexSet<String>,
+    pub all: IndexSet<(Option<usize>, String)>,
```
peterhuene, line 26: Maybe only `all` needs to be `IndexMap<String, Option<usize>>` and `missing` and `unexpected` can remain as they are?

That means we need to manually produce `missing` and `unexpected` rather than using the `-` operator on the sets, but that shouldn't be too tricky.

wdl-gauntlet/src/report.rs
```diff
@@ -36,7 +36,7 @@ pub enum Status {
     /// The document had diagnostics, but the diagnostics exactly matched what
     /// was already expected in the configuration, meaning that this
     /// [`Status`] is considered successful.
-    DiagnosticsMatched(IndexSet<String>),
+    DiagnosticsMatched(IndexSet<(Option<usize>, String)>),
```
peterhuene, line 39: Likewise, this could probably just mirror the `all` above by being an `IndexMap<String, Option<usize>>`.

wdl-gauntlet/src/lib.rs
```diff
@@ -228,27 +228,27 @@ pub async fn gauntlet(args: Args) -> Result<()> {
                         .context("failed to write diagnostic")?;
 
                     assert!(
-                        actual.insert(
+                        actual.insert((
+                            diagnostic.labels().next().map(|l| l.span().start()),
```
peterhuene, line 232: We'll still need to use `line_index` from the above `SimpleFile` to get the line number as `start` is returning the byte offset into the file.

wdl-gauntlet/src/document/identifier.rs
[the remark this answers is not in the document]
a-frantz: Ok I have a good answer now! The derived `Serialize` creates this TOML:
```TOML
[[diagnostics]]
message = "ww-vc-trio.wdl:246:17: warning[SnakeCase]: output name `Mutect_AnnotatedVcf` is not snake_case"
permalink = "https://github.com/getwilds/ww-vc-trio/blob/ee08634f28810e5d6fd1a904fc83f4e67821550e/ww-vc-trio.wdl/#L246"

[diagnostics.document]
repository = "getwilds/ww-vc-trio"
path = "/ww-vc-trio.wdl"
```

And the derived `Deserialize` fails to go from the TOML String to a `repository::Identifier` and a [sub]`String`.

wdl-gauntlet/src/lib.rs
```diff
@@ -227,28 +228,35 @@ pub async fn gauntlet(args: Args) -> Result<()> {
...
                 }
             }
 
             // As the list of diagnostics has been sorted by document identifier, do
             // a binary search and collect the matching messages
             let diagnostics = config.inner().diagnostics();
-            let doc_id_str = document_identifier.to_string();
-            let expected: IndexSet<String> = diagnostics
-                .binary_search_by_key(&doc_id_str.as_str(), |d| d.document())
+            let expected: IndexSet<(String, usize)> = diagnostics
```
peterhuene, line 253: Does this need to be `IndexSet<(String, usize)>`? Could it not just be `IndexSet<String>` containing only the messages?

That way below we don't need `expected_messages` either.

wdl-gauntlet/src/config/inner.rs
```diff
@@ -9,32 +9,73 @@ use serde::Deserialize;
...
     }
+
+    /// Gets the permalink to the source of the diagnostic.
+    pub fn permalink(&self) -> &str {
+        &self.permalink
+    }
+
+    /// Gets the line number of the diagnostic.
+    /// If the line number can't be parsed, it defaults to 1.
+    pub fn line_no(&self) -> usize {
```
peterhuene, line 71: See previous comment, but I don't think we need this method.

wdl-gauntlet/src/config/inner.rs
```diff
@@ -9,32 +9,73 @@ use serde::Deserialize;
...
 }
 
 impl Diagnostic {
     /// Creates a new diagnostic for the given document identifier and message.
-    pub fn new(document: String, message: String) -> Self {
-        Self { document, message }
+    pub fn new(
+        document: document::Identifier,
+        message: String,
+        hash: RawHash,
```
peterhuene, line 33: Can we take this by reference?

wdl-gauntlet/src/lib.rs
```diff
@@ -335,10 +345,20 @@ pub async fn gauntlet(args: Args) -> Result<()> {
...
         }
 
-        for message in messages {
+        let hash = config
+            .inner()
+            .repositories()
+            .get(identifier.repository())
+            .unwrap()
+            .commit_hash()
+            .clone()
```
peterhuene, line 354: See previous comment, but if we take the hash by reference, we shouldn't need to clone it here or below in the loop.

peterhuene (commented):
Looks great, just a few minor comments.

peterhuene (approved):

PR #93, opened 2024-06-20
feat: MissingMetasRule and MissingOutputRule

a-frantz:
This pull request adds a new rule to `wdl-lint`.

- **Rule Name**: `MissingMetas` and `MissingOutput`

_Describe the rules you have implemented and link to any relevant issues._

Before submitting this PR, please make sure:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.

Rule specific checks:

- [x] You have added the rule as an entry within `RULES.md`.
- [x] You have added the rule to the `rules()` function in `wdl-lint/src/v1.rs`.
- [x] You have added a test cases in `wdl-lint/tests/lints` that covers every 
      possible diagnostic emitted for the rule within the file where the rule  
      is implemented.
- [x] You have run `wdl-gauntlet --refresh` to ensure that there are no 
      unintended changes to the baseline configuration file (`Gauntlet.toml`).
- [x] You have run `wdl-gauntlet --refresh --arena` to ensure that all of the 
      rules added/removed are now reflected in the baseline configuration file 
      (`Arena.toml`).

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.0.0/

PR #95, opened 2024-06-21
refactor: make `Visitor` and implementations version agnostic.

peterhuene:
Currently, AST visitation uses `SyntaxKind` to determine which methods on a visitor to call. As `SyntaxKind` is intentionally the union of all possible WDL versions, there's no reason for the `Visitor` trait to be in a `v1` module.

This commit moves the `Visitor` trait to the root of the crate and also moves any implementations of `Visitor` out of a `v1` module as well. For validation, they now go in a `validation` module. For lint rules, they go in a `rules` module.

For the lint rules, it means that if a lint rule works only on certain nodes or tokens (e.g. whitepsace) that aren't tied to a particular version of WDL or the syntax they operate on doesn't change from version to version, the rule will continue to operate on newer versions of WDL without changes.

As new syntax is introduced over time, rules that do need to operate on the newer syntax will need to implement additional `Visitor` trait methods to visit the new syntax nodes.

If for whatever reason a rule needs to be specific to a particular version of WDL, it can implement the `document` visitor method and check the version of the document; if it's not the appropriate version, the rule can disable itself.

      changes (when appropriate).


peterhuene:
I'll wait until #93 is merged and then rebase this on top of that.

a-frantz (approved):
Good stuff 👍

PR #97, opened 2024-06-21
feat: ImportWhitespace rule

a-frantz:

- **Rule Name**: `ImportWhitespace`

Rule specific checks:

- [x] You have added the rule to the `rules()` function in `wdl-lint/src/v1.rs`.
      is implemented.
      (`Arena.toml`).

### DIFF
--- a/wdl-lint/src/v1/import_whitespace.rs
+++ b/wdl-lint/src/v1/import_whitespace.rs
@@ -83,32 +83,38 @@ impl Visitor for ImportWhitespaceVisitor {
             return;
         }
 
-        let last_whitespace = stmt
+        let mut prev_token = stmt
             .syntax()
             .prev_sibling_or_token()
             .and_then(SyntaxElement::into_token);
 
-        if let Some(token) = last_whitespace {
+        while let Some(token) = prev_token {
             if token.kind() == SyntaxKind::Whitespace {
-                let mut too_many_lines = false;
+                let mut should_warn = false;
                 let mut second_line_start = None;
                 for (i, (_, _, next)) in lines_with_offset(token.text()).enumerate() {
                     if i == 0 {
                         second_line_start = Some(next);
-                    } else if i > 0 {
-                        too_many_lines = true;
+                    } else if i == 1 {
+                        should_warn = true;
+                    } else if i == 2 {
+                        should_warn = false;
                         break;
                     }
                 }
 
-                if too_many_lines {
+                if should_warn {
                     let span = token.text_range().to_span();
                     state.add(bad_import_whitespace(Span::new(
                         span.start() + second_line_start.expect("should have a second line start"),
                         span.len() - second_line_start.expect("should have a second line start"),
                     )));
                 }
+            } else if token.kind() != SyntaxKind::Comment {
+                // We've backed into non-trivia, so we're done.
+                break;
             }
-        };
+            prev_token = token.prev_token();
+        }
     }
 }
--- a/wdl-lint/src/v1/import_whitespace.rs
+++ b/wdl-lint/src/v1/import_whitespace.rs
@@ -82,22 +82,20 @@ impl Visitor for ImportWhitespaceVisitor {
             return;
         }
 
-        stmt.syntax()
-            .prev_sibling_or_token()
-            .unwrap()
-            .into_token()
-            .map(|token| {
-                if token.kind() == SyntaxKind::Whitespace {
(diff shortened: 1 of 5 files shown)


wdl-lint/src/v1/import_whitespace.rs
```diff
@@ -0,0 +1,101 @@
...
+
+        if let Some(token) = last_whitespace {
+            if token.kind() == SyntaxKind::Whitespace {
+                let mut num_lines = 0;
+                lines_with_offset(token.text()).for_each(|_| {
+                    num_lines += 1;
+                });
+
+                if num_lines > 1 {
+                    let span = token.text_range().to_span();
```
peterhuene, line 95: We should probably adjust the span here so that it starts at the beginning of the second line.

We also don't have to count all the lines, just know there's more than one.

Something like:

```rust
let mut num_lines = 0;
let mut second_line_start = None;
for (_, _, next) in lines_with_offset(token.text()) {
    num_lines += 1;

    if num_lines == 1 {
        second_line_start = Some(next);
    } else if num_lines > 1 {
        break;
    }
}

if num_lines > 1 {
    let span = token.span();
    state.add(bad_import_whitespace(Span::new(span.start() + second_line_start.expect("should have a start"), span.len() - second_line_start));
}
```

wdl-lint/src/v1/import_whitespace.rs
```diff
@@ -0,0 +1,101 @@
...
+            .syntax()
+            .prev_sibling()
+            .map(|s| s.kind() == SyntaxKind::ImportStatementNode)
+            .unwrap_or(false);
+        if !between_imports {
+            // Another rule will catch any whitespace here.
+            return;
+        }
+
+        let last_whitespace = stmt.syntax().prev_sibling_or_token().unwrap().into_token();
```
peterhuene, line 85: While it should never occur that an import statement is parsed before a version statement (and thus can't be the first child of the root node), could we change this to:

```suggestion
        let last_whitespace = stmt.syntax().prev_sibling_or_token().and_then(SyntaxElement::into_token);
```

wdl-lint/src/v1/import_whitespace.rs
```diff
@@ -0,0 +1,101 @@
...
+use crate::util::lines_with_offset;
+use crate::Tag;
+use crate::TagSet;
+
+/// The identifier for the import sort rule.
+const ID: &str = "ImportWhitespace";
+
+/// Creates a bad import whitespace diagnostic.
+fn bad_import_whitespace(span: Span) -> Diagnostic {
+    Diagnostic::note("No blank lines allowed between imports")
```
peterhuene, line 23: For consistency with other diagnostics:

```suggestion
    Diagnostic::note("blank lines are not allowed between imports")
```

wdl-lint/src/v1/import_whitespace.rs
```diff
@@ -0,0 +1,101 @@
...
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_fix("Delete this whitespace.")
```
peterhuene, line 26: I only like to use "this" in labels as they directly point at source and labels should basically say "this is the offending thing"; as that's not the case here, I think this wording would work a little better as a "fix":

```suggestion
        .with_fix("remove any blank lines between imports")
```

peterhuene (commented):
Looks good, just a few comments.

wdl-lint/tests/lints/import-whitespace/source.wdl
```diff
@@ -0,0 +1,21 @@
...
+
+import "foo.wdl"    # OK
+import "bar.wdl"    # OK
+
+import "baz.wdl"    # BAD
+
+import "huh.wdl"    # BAD
+# but a comment makes it OK
+import "vom.wdl"   # OK
+
```
peterhuene, line 13: I just noticed that this blank line isn't getting a diagnostic and I think that's because of the comment.

I think we may need to change `last_whitespace` to instead look back through _all_ whitespace tokens until we hit the previous import node and count across all whitespace tokens encountered.

Also, now that I think about it, we probably only want to report this diagnostic if there's exactly two lines; any more than that and we'll be triggering the more general `Whitespace` that warns on more than one blank line in a row.

a-frantz: Good eyes! I think I addressed all this in what I'm about to commit.

peterhuene (approved):

PR #99, opened 2024-06-24
feat: LineWidth rule

a-frantz:

- **Rule Name**: `LineWidth`

Rule specific checks:

      is implemented.
      (`Arena.toml`).


adthrasher (approved):

wdl-lint/tests/lints/matching-param-meta/source.errors
```diff
@@ -1,3 +1,11 @@
+note[LineWidth]: line exceeds maximum width of 90
+  ┌─ tests/lints/matching-param-meta/source.wdl:1:1
+  │
+1 │ ## This is a test for checking for missing and extraneous entries in a `parameter_meta` section.
```
peterhuene, line 4: Since this one is in a comment, can we split the comment into two lines so that the diagnostic doesn't fire?

wdl-lint/tests/lints/double-quotes/source.errors
```diff
@@ -6,6 +6,14 @@ warning[DoubleQuotes]: string defined with single quotes
   │
   = fix: change the single quotes to double quotes
 
+note[LineWidth]: line exceeds maximum width of 90
+   ┌─ tests/lints/double-quotes/source.wdl:11:1
+   │
+11 │         "this string is ok ~{'but this is not and ~{"while this one is okay ~{'this one is not'}"}'}!"
```
peterhuene, line 12: I see what you mean about long strings.

I can't think of a good way of working around this then perhaps checking at the newline point if the immediate previous sibling was a string node, and if so, don't include it in the line length total.

So this would fire (the comment is treated as the "end" of the line for the width):

```wdl
"this string is ok ~{'but this is not and ~{"while this one is okay ~{'this one is not'}"}'}!" # FOO
```

And this would (the whitespace _before_ the string is too long):

```wdl
                                                                                                                                  "string"
```

But this would not (the whitespace before the string is fine):

```wdl
      "this string is ok ~{'but this is not and ~{"while this one is okay ~{'this one is not'}"}'}!"
```

wdl-lint/src/rules/line_width.rs
```diff
@@ -0,0 +1,131 @@
...
+use wdl_ast::Diagnostics;
+use wdl_ast::Span;
+use wdl_ast::Visitor;
+
+use crate::Rule;
+use crate::Tag;
+use crate::TagSet;
+
+/// The maximum width of a line.
+const MAX_WIDTH: usize = 90;
```
peterhuene, line 14: I wonder if we should make this just a field on rule and the visitor (e.g. `LineWidthRule::new(80)`) with `LineWidthRule::default` being implemented that uses 90?

peterhuene (commented):
Just a few comments.

wdl-lint/tests/lints/double-quotes/source.errors
a-frantz: I'm going to implement that idea and I'll report back with how it changes `Arena.toml`. I'm having a hard time judging the concept without some tangible results to look at. But in the meantime, I'll point out that we could rewrite this example string so that it would be compliant:
```
        "this string is ok ~{
            'but this is not and ~{
                "while this one is okay ~{
                    'this one is not'
                }"
            }'
        }!"
```
(also pinging @adthrasher )

a-frantz: @peterhuene definitely an interesting idea that was worth testing, but in the wild it eliminated a single diagnostic: https://github.com/stjudecloud/workflows/blob/46a77b33c99c1952396fcaaf0845f1fb5a015987/tools/star.wdl/#L701

It's a hard exception to explain and doesn't seem to actually impact "wild WDL" very often.

peterhuene: I'm fine with leaving as-is then and see if anyone else runs into problems with the rule firing on strings in the future; they could always do an interpolation hack to work around it too.

peterhuene: I'll add that we should apply your suggested workaround above to format the interpolations such that it won't trigger the rule.

peterhuene (approved):

adthrasher (approved):

PR #104, opened 2024-06-26
feat: add inconsistent_newlines rule

adthrasher:

- **Rule Name**: `InconsistentNewlines`

This rule checks for usage of both "\r\n" and "\n".

Rule specific checks:

      is implemented.
      (`Arena.toml`).


wdl-lint/src/rules/inconsistent_newlines.rs
```diff
@@ -0,0 +1,83 @@
...
+
+    fn document(&mut self, state: &mut Self::State, reason: VisitReason, doc: &wdl_ast::Document) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        let mut newline = 0;
+        let mut carriage_return = 0;
+
+        doc.syntax()
```
peterhuene, line 68: So this walks only the children of the root, not all the whitespace tokens in the document.

I would suggest implementing the `whitespace` trait method to do the actual detection, setting fields on the visitor to denote if we've detected `\r\n` and `\n`, then in the `document` trait method add the diagnostic on exit.

The `whitespace` method will be called for every whitespace token in the file, regardless of where they appear.

wdl-lint/src/rules/inconsistent_newlines.rs
```diff
@@ -0,0 +1,83 @@
...
+            .children_with_tokens()
+            .filter(|c| c.kind() == SyntaxKind::Whitespace)
+            .for_each(|w| {
+                if w.to_string().contains("\r\n") {
```
peterhuene, line 72:
```suggestion
                if w.as_str().contains("\r\n") {
```

wdl-lint/src/rules/inconsistent_newlines.rs
```diff
@@ -0,0 +1,83 @@
...
+                    carriage_return += 1;
+                } else if w.to_string().contains('\n') {
```
peterhuene, line 74:
```suggestion
                } else if w.as_str().contains('\n') {
```

wdl-lint/src/rules/inconsistent_newlines.rs
```diff
@@ -0,0 +1,83 @@
...
+                    newline += 1;
+                }
+            });
+
+        if newline > 0 && carriage_return > 0 {
+            state.add(inconsistent_newlines(doc.syntax().text_range().to_span()));
```
peterhuene, line 80: This will highlight the entire document, which is probably not what we want to do.

If we do want to highlight or label anything, I think we would report the span of the first mismatched newline that was detected in the file.

wdl-lint/src/rules/inconsistent_newlines.rs
```diff
@@ -0,0 +1,83 @@
...
+use crate::TagSet;
+
+/// The identifier for the inconsistent newlines rule.
+const ID: &str = "InconsistentNewlines";
+
+/// Creates an inconsistent newlines diagnostic.
+fn inconsistent_newlines(span: Span) -> Diagnostic {
+    Diagnostic::note("inconsistent newlines detected")
+        .with_rule(ID)
+        .with_highlight(span)
```
peterhuene, line 23: If we change `span` to be the span of the first mismatched newline, we should probably use a label to indicate that we're only showing the first such occurrence in the file:

```suggestion
        .with_label("the first mismatched newline is here", span)
```

peterhuene (commented):
Looks good, just a few feedback comments to resolve.

wdl-lint/src/rules/inconsistent_newlines.rs
adthrasher: I've done this, but when is a token not a child of the document root?

peterhuene: Whitespace tokens are only (immediate) children of the root if the whitespace appears outside other node children of the root.

For example:

```wdl
version 1.1

# The above and below whitespace are children of the root and siblings to the workflow definition node.

workflow foo {

    # The whitespace in here is a child of the workflow definition node, however.

}
```

While `descendants_with_tokens` could have been used instead to process each descendant whitespace token, that would amount to another full walk of the AST, which the validator is doing already when calling each visitor.

wdl-lint/src/rules/inconsistent_newlines.rs
```diff
@@ -0,0 +1,96 @@
...
+        if reason == VisitReason::Exit && self.newline > 0 && self.carriage_return > 0 {
+            state.add(inconsistent_newlines(self.first_inconsistent.unwrap()));
+        }
+    }
+
+    fn whitespace(&mut self, _state: &mut Self::State, whitespace: &Whitespace) {
+        if whitespace.as_str().contains("\r\n") {
+            self.carriage_return += 1;
+            if self.newline > 0 && self.first_inconsistent.is_none() {
+                self.first_inconsistent = Some(whitespace.span());
```
peterhuene, line 87: If we change the `contains` above to a `find` to find the offset to the `\r\n`, then we could calculate the span of _only_ the offending `\r\n`, like so:

```rust
self.first_inconsistent = Span::new(whitespace.span().start() + pos, 2);
```

and similarly for the `\n` check below (except with a length of `1`).

peterhuene: Doing it this way would have the label only point to the newline and not any of the surrounding continuous whitespace.

wdl-lint/src/rules/inconsistent_newlines.rs
adthrasher: Ah, that's it `children_with_tokens()` is the _immediate_ children of the `root` not a recursive set of children.

peterhuene (approved):
Looks great!

PR #108, opened 2024-07-03
feat: new wdl-format crate

a-frantz:

Adds a `wdl-format` crate.

      changes (when appropriate).


claymcleod:
This is a great start in terms of getting all of the elements to format correctly. Looking at the results in https://github.com/stjudecloud/workflows/pull/165, they look quite good (with a few nitpicks).

I think there is more to consider regarding the architecture—in particular, how can we more tightly couple the formatting capabilities with the current structure. Ideally, when you implement a new WDL element, you're statically compelled to think about and define how that element should be formatted.

In my mind, this behavior should be compelled by the requirement of a trait implementation dedicated to that function for each AST node. We could simply rely on something like `std::fmt::Display`, or we could introduce something like `trait Formatable`. Further, it's not clear if this should live in `wdl-ast` or be something that exists as a layer on top within `wdl-format`.

wdl-format/src/format.rs
```diff
@@ -0,0 +1,912 @@
...
+                result.push_str(&format_struct_definition(&struct_def));
+            }
+        };
+    });
+
+    Ok(result)
+}
+
+#[cfg(test)]
+mod tests {
```
claymcleod, line 592: These tests also need something more scalable—akin to what we do in the other packages where files are more easily observed and reasoned about (rather than embedding the text right in the rust source).

a-frantz: Totally agree. That was on my agenda to deal with tomorrow.

a-frantz:
History got all weird after a botched rebase+force push. Continued in #133

PR #110, opened 2024-07-08
feat: implement multi-file analysis.

peterhuene:
This commit implements multi-file analysis by introducing the `wdl-analysis` crate.

Analysis is performed via an `AnalysisEngine` that is async and uses `rayon` for a thread pool.

Analysis builds a document graph by resolving imports from both local files and remote files over HTTP.

Files are parsed in parallel; once parsing completes and the document graph is constructed, analysis proceeds in topological order.

Currently analysis is limited to building a `DocumentScope` for each document that contains information about namespaces introduced by imports, struct definitions, and scopes for both tasks and workflows.

The information could be used to quickly identify the names available at any given position in the file.

Note that several "single file" validations that existed previously (chiefly the "names" validation) has moved into analysis and enhanced now that imports are being resolved.

An `analyze` subcommand was added to the `wdl` CLI; this performs analysis and currently just prints out the debug representation of the results. The output might change in the future.

The `wdl check` command now performs analysis only and exits with a zero status if there are no diagnostics printed.

There has also been a breaking change to the `wdl_ast::Document` API: the `Parse` type was removed in favor of having `Document::parse` return simply a tuple of the document and any diagnostics as a `Vec<Diagnostic>`; this allows analysis to append diagnostics to the same collection rather than having to store multiple immutable collections of diagnostics.

Also changed:

* the `binaries` feature was renamed to simply `cli` as now the repo only builds one binary (excluding gauntlet).
* the type representation in the AST now performs WDL type equality rather than node equality.
* the `PartialOrd` and `Ord` implementation of the type representation in the AST has moved into the lint rule relating to sorting of declarations.
* AST visitors are now required to implement the `document` method and it is expected that the visitor will reset its internal state upon entering a document; this allows analysis to reuse a visitor between validations of different documents.
* Certain diagnostics now point at the end of the input instead of having no relation to the file that caused the diagnostic; this is needed now as analysis may generate diagnostics for different files in its output.
* Fixed a bug in the preamble whitespace rule to not fire if there's a single line following a version to end the file.

      changes (when appropriate).


wdl-analysis/tests/analysis/import-incompatible-versions/source.diagnostics
```diff
@@ -0,0 +1,12 @@
+error: unsupported WDL version `2.0`
+  ┌─ tests/analysis/import-incompatible-versions/foo.wdl:1:9
+  │
+1 │ version 2.0
+  │         ^^^ this version of WDL is not supported
+
+error: imported document has incompatible version `2.0`
```
claymcleod, line 7: It would be a nice to have to put the original WDL version here as well to make it clear what version the import is being called from.

claymcleod (approved):
Looks great.

README.md
```diff
@@ -90,7 +90,7 @@ the `wdl` family of crates.
 The `wdl` CLI tool can be run with the following command:
 
 ```bash
-cargo run --bin wdl --features binaries -- $ARGS
+cargo run --bin wdl --features cli -- $ARGS
```
claymcleod, line 93: I believe we should also add a bullet below for the `analysis` subcommand.

a-frantz (approved):
Looks great Peter 👍

PR #113, opened 2024-07-09
feat: description missing lint rule

adthrasher:

- **Rule Name**: `DescriptionMissing`

This rule ensures that each meta section includes a `description` keyword.

Rule specific checks:

      is implemented.
      (`Arena.toml`).


wdl-lint/src/rules/description_missing.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    fn description(&self) -> &'static str {
+        "Ensures that a description is present for each meta section."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "Each task or workflow should have a description in the meta section. This description \
+         should be an explanation of the task or workflow. This description should be written in \
+         active voice and complete sentences. More detailed information can be included in the \
+         `help` key."
```
peterhuene, line 45: Nit: for some reason two `this` sentences in a row reads weird to me. If others agree, maybe:

```suggestion
        "Each task or workflow should have a description in the meta section. The description \
         should be an explanation of the task or workflow. The description should be written in \
         active voice and complete sentences. More detailed information can be included in the \
         `help` key."
```

wdl-lint/tests/lints/between-import-whitespace/source.errors
```diff
@@ -51,3 +51,11 @@ warning[Whitespace]: more than one blank line in a row
...
    = fix: remove the unnecessary blank lines
 
+note[DescriptionMissing]: description key missing in meta section
+   ┌─ tests/lints/between-import-whitespace/source.wdl:22:5
+   │
+22 │     meta {}
+   │     ^^^^
+   │
+   = fix: add a description to the meta section
+
```
peterhuene, line 61: Could we add an `#@ except: DescriptionMissing` to the existing files to suppress this lint and also add a test dedicated to intentionally missing a description?

wdl-lint/src/rules/description_missing.rs
```diff
@@ -0,0 +1,81 @@
+//! A lint rule for to ensure description is included in meta sections.
```
claymcleod, line 1:
```suggestion
//! A lint rule to ensure a description is included in meta sections.
```

wdl-lint/src/rules/description_missing.rs
claymcleod: That about "This" and then "The"? I agree that would should avoid two "This" sentences, and I agree that's just a slight preference.

wdl-lint/tests/lints/between-import-whitespace/source.errors
```diff
@@ -51,3 +51,11 @@ warning[Whitespace]: more than one blank line in a row
    │  
    = fix: remove the unnecessary blank lines
 
+note[DescriptionMissing]: description key missing in meta section
+   ┌─ tests/lints/between-import-whitespace/source.wdl:22:5
+   │
+22 │     meta {}
```
claymcleod, line 57: Same question here as your last PR: should the reported span encapsulate the whole meta object or just this keyword?

wdl-lint/src/rules/description_missing.rs
```diff
@@ -0,0 +1,81 @@
...
+
+/// The identifier for the description missing rule.
+const ID: &str = "DescriptionMissing";
+
+/// Creates a description missing diagnostic.
+fn description_missing(span: Span) -> Diagnostic {
+    Diagnostic::note("description key missing in meta section")
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_fix("add a description to the meta section")
```
claymcleod, line 25:
```suggestion
        .with_fix("add a `description` key to the meta section")
```

I might consider splitting this into two different diagnostics: one for tasks and one for workflows:

- "add a `description` key to the meta section describing the task"
- "add a `description` key to the meta section describing the workflow"

wdl-lint/tests/lints/command-mixed-line-cont/source.errors
```diff
@@ -1,3 +1,11 @@
+note[DescriptionMissing]: description key missing in meta section
+  ┌─ tests/lints/command-mixed-line-cont/source.wdl:7:5
+  │
+7 │     meta {}
```
claymcleod, line 4: For all of these, I would add `#@ except: DescriptionMissing` to the top of the document so that these errors don't show up for non-relevant tests.

wdl-lint/tests/lints/between-import-whitespace/source.errors
peterhuene: I personally think it's find to point at the `meta` keyword to say "this section is missing a `description` key".

Highlighting the whole section feels like it would call attention to mostly valid things.

wdl-lint/src/rules/description_missing.rs
```diff
@@ -0,0 +1,91 @@
...
+use crate::Rule;
+use crate::Tag;
+use crate::TagSet;
+
+/// The identifier for the description missing rule.
+const ID: &str = "DescriptionMissing";
+
+/// Creates a description missing diagnostic.
+fn description_missing(span: Span) -> Diagnostic {
+    Diagnostic::note("this section is missing a description key")
```
a-frantz, line 23: Potential edit:
```suggestion
    Diagnostic::note("<task or workflow `name`> missing a description meta key")
```
I'm thinking of how it will appear in short-form. `this` loses context when all you have is the line+col number. The name of the task or workflow would be nice to have in a context like `Arena`.

a-frantz (commented):
So this lint doesn't fire if there is no meta section, right? Are we sure that's the right behavior? Presumably sometime soon we're going to implement a lint for "must have a meta section". Should that be handled all at once?

wdl-lint/src/rules/description_missing.rs
adthrasher: It has a `highlight` which tags the relevant `meta` keyword.

adthrasher:
> So this lint doesn't fire if there is no meta section, right? Are we sure that's the right behavior? Presumably sometime soon we're going to implement a lint for "must have a meta section". Should that be handled all at once?

I think you're already handling that case here: https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/rules/missing_metas.rs

wdl-lint/src/rules/description_missing.rs
adthrasher: Nevermind @a-frantz , I see what you mean. Now I'm reporting like this: 
```
note[DescriptionMissing]: task `foo` is missing a description key
  ┌─ tests/lints/description-missing/source.wdl:6:5
  │
6 │     meta {}
  │     ^^^^
  │
  = fix: add a `description` key to this meta section
```

a-frantz: Nice, LGTM 👍

a-frantz:
> > So this lint doesn't fire if there is no meta section, right? Are we sure that's the right behavior? Presumably sometime soon we're going to implement a lint for "must have a meta section". Should that be handled all at once?
> 
> @a-frantz - I think you're already handling that case here: https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/rules/missing_metas.rs

Right, I forgot about that 🤦 . But should we consolidate these rules somehow? Maybe I'm over complicating things, but I'm concerned with the case that a user gets the `MissingMetas` lint, thinks they've "fixed" their code by adding metas, and then they receive your new lint because they did it "wrong". Is that two step process annoying?

adthrasher:
> > > So this lint doesn't fire if there is no meta section, right? Are we sure that's the right behavior? Presumably sometime soon we're going to implement a lint for "must have a meta section". Should that be handled all at once?
> > 
> > 
> > @a-frantz - I think you're already handling that case here: https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/rules/missing_metas.rs
> 
> Right, I forgot about that 🤦 . But should we consolidate these rules somehow? Maybe I'm over complicating things, but I'm concerned with the case that a user gets the `MissingMetas` lint, thinks they've "fixed" their code by adding metas, and then they receive your new lint because they did it "wrong". Is that two step process annoying?

I'm not sure it matters. If we fold it into the `MissingMetas` rule, presumably you wouldn't flag the missing `description` as well. We've been avoiding giving repetitive warnings. Then once `meta` exists, you'd get the missing `description` warning.

adthrasher:
> > > > So this lint doesn't fire if there is no meta section, right? Are we sure that's the right behavior? Presumably sometime soon we're going to implement a lint for "must have a meta section". Should that be handled all at once?
> > > 
> > > 
> > > @a-frantz - I think you're already handling that case here: https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/rules/missing_metas.rs
> > 
> > 
> > Right, I forgot about that 🤦 . But should we consolidate these rules somehow? Maybe I'm over complicating things, but I'm concerned with the case that a user gets the `MissingMetas` lint, thinks they've "fixed" their code by adding metas, and then they receive your new lint because they did it "wrong". Is that two step process annoying?
> 
> I'm not sure it matters. If we fold it into the `MissingMetas` rule, presumably you wouldn't flag the missing `description` as well. We've been avoiding giving repetitive warnings. Then once `meta` exists, you'd get the missing `description` warning.

Maybe `MissingMetas` could have `.fix()` updated to clarify to add a `meta` with a `description` field.

a-frantz:
Yea good points. OOS for this PR, but another idea: can we link rules somehow? Like in the output of `sprocket explain MissingMetas` can we say "Also see `DescriptionMissing`"?

adthrasher:
> Yea good points. OOS for this PR, but another idea: can we link rules somehow? Like in the output of `sprocket explain MissingMetas` can we say "Also see `DescriptionMissing`"?

Just added #116 to track that issue.

a-frantz (approved):
Alright this should be good to merge then 🚀

PR #114, opened 2024-07-09
feat: add non-matching output lint rule

adthrasher:

- **Rule Name**: `NonmatchingOutput`

This lint rule ensures that if `output` is populated, then there is an `outputs` key in the `meta` section. It also ensures that each `output` has a corresponding entry in `outputs`.

Rule specific checks:

      is implemented.
      (`Arena.toml`).


a-frantz:
Maybe an annoying question to ask: does it make more sense to have the key named the same as the section? i.e. both of them `output` without the 's'.

adthrasher:
> Maybe an annoying question to ask: does it make more sense to have the key named the same as the section? i.e. both of them `output` without the 's'.

@a-frantz - That's a good question. We were inconsistent in the `workflows` repo. I suppose `output` isn't a keyword when in a `meta` section.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,180 @@
...
+        "Ensures that each output field is documented in the meta section."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "The meta section should have an output key and keys with descriptions for each output of \
+         the task/workflow. These must match exactly. i.e. for each named output of a task or \
+         workflow, there should be an entry under meta.output with that same name. Additionally, \
+         these entries should be in the same order (that order is up to the developer to decide). \
+         No extraneous output entries are allowed. There should not be any blank lines inside the \
+         entire meta section."
```
a-frantz, line 76:
```suggestion
         No extraneous output entries are allowed."
```
As this sentence isn't being enforced by this rule, it shouldn't be mentioned here. (Although maybe once that lint rule gets written it could be "linked" (per #116 )).

adthrasher: I typically just copy/paste the description from #11. So this isn't accurate at all. I'm not enforcing `No extraneous output entries are allowed` either. 

The spacing statement seems like it should be in a different rule anyway.

adthrasher: I'm also not enforcing that the orders are the same.

a-frantz: IMO at least the "no extraneous entries" check belongs within this rule. Probably the order as well. We've discussed (without deciding anything) that maybe `input` and `parameter_meta` sections could be in different orders. There's no rule in #11 about ordering the output section. Seeing as how that's completely "customizable" I think enforcing the same order in the `meta.output` makes sense.

wdl-lint/tests/lints/missing-preamble-ws/source.wdl
```diff
@@ -1,6 +1,5 @@
 #@ except: DescriptionMissing
 ## This is a test of missing preamble whitespace.
-
```
claymcleod, line 3: This looks like an unintended addition.

wdl-lint/tests/lints/missing-preamble-ws/source.errors
```diff
@@ -0,0 +1,10 @@
+note[PreambleWhitespace]: expected a blank line before the version statement
+  ┌─ tests/lints/missing-preamble-ws/source.wdl:2:50
```
claymcleod, line 2: As well as this.

wdl-lint/tests/lints/missing-preamble-ws/source.wdl
adthrasher: This reverts a mistaken change in a prior PR. There should be no whitespace there, so that the preamble whitespace rule fires.

adthrasher: Here: https://github.com/stjude-rust-labs/wdl/commit/1e289b967673bad25504e33f35733d907a2d6467#diff-df8d360942f292702a6c79b38bf6e6af9db95c7dd8d4879eb58ae349b8110b87

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: I ended up checking both lists and reporting extraneous entries in either list. Then if no extra/missing entries are found, I compare the order. It's a little cumbersome because if something is missing and the `outputs` key is out-of-order, you won't get the out-of-order message until you fix the first one. I toyed with trying to do something to remove the extra/missing entries and check the resulting ordering, but I think that's going to be cumbersome and also potentially confusing to the users (and possibly error-prone).

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+        TaskOrWorkflow::Workflow(w) => ("workflow", w.name().as_str().to_string()),
+    };
+
+    Diagnostic::warning(format!(
+        "output `{name}` is missing from `meta` section in {ty} `{item_name}`"
+    ))
+    .with_rule(ID)
+    .with_highlight(span)
+    .with_fix(format!(
+        "add output (`{name}`) key to `outputs` documentation in `meta` section"
```
claymcleod, line 39:
```suggestion
        "add a description of this output at `meta.outputs.{name}`"
```

claymcleod: I'm not 100% sold that this verbiage is better, but wanted to see what you thought.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+
+/// Creates a missing outputs in meta diagnostic.
+fn missing_outputs_in_meta(span: Span, context: &TaskOrWorkflow) -> Diagnostic {
+    let (ty, name) = match context {
+        TaskOrWorkflow::Task(t) => ("task", t.name().as_str().to_string()),
+        TaskOrWorkflow::Workflow(w) => ("workflow", w.name().as_str().to_string()),
+    };
+
+    Diagnostic::warning(format!(
+        "`outputs` key missing in `meta` section in {ty} `{name}`"
```
claymcleod, line 51:
```suggestion
        "`outputs` key missing in `meta` section for the {ty} `{name}`"
```

Maybe not two "in" clauses in a row?

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+    ))
+    .with_rule(ID)
+    .with_highlight(span)
+    .with_fix("add `outputs` key to `meta` section")
```
claymcleod, line 55:
```suggestion
    .with_fix("add an `outputs` key to `meta` section describing the outputs")
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+}
+
+/// Creates a diagnostic for extra `meta` entries.
```
claymcleod, line 58:
```suggestion
/// Creates a diagnostic for extra `meta.outputs` entries.
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+fn extra_output_in_meta(span: Span, name: &str, context: &TaskOrWorkflow) -> Diagnostic {
+    let (ty, item_name) = match context {
+        TaskOrWorkflow::Task(t) => ("task", t.name().as_str().to_string()),
+        TaskOrWorkflow::Workflow(w) => ("workflow", w.name().as_str().to_string()),
+    };
+
+    Diagnostic::warning(format!(
+        "`{name}` appears in `outputs` section of {ty} `{item_name}` but is not a declared \
```
claymcleod, line 66:
```suggestion
        "`{name}` appears in `outputs` section of the {ty} `{item_name}` but is not a declared \
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+         `output`"
+    ))
+    .with_rule(ID)
+    .with_highlight(span)
+    .with_fix(format!(
+        "remove `{name}` from `outputs` key in `meta` section"
```
claymcleod, line 72:
```suggestion
        "ensure the output exists or remove the `{name}` key from `meta.outputs`"
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+        TaskOrWorkflow::Workflow(w) => ("workflow", w.name().as_str().to_string()),
+    };
+
+    Diagnostic::warning(format!(
+        "`outputs` section of `meta` for {ty} `{item_name}` is out of order"
+    ))
+    .with_rule(ID)
+    .with_highlight(span)
+    .with_highlight(output_span)
+    .with_fix("the `outputs` keys of `meta` section must match the order in `output`")
```
claymcleod, line 89:
```suggestion
    .with_fix("ensure the keys within `meta.outputs` have the same order as they appear in `outputs`")
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+#[derive(Default, Debug, Clone, Copy)]
+pub struct NonmatchingOutputRule;
+
+impl Rule for NonmatchingOutputRule {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Ensures that each output field is documented in the meta section."
```
claymcleod, line 102:
```suggestion
        "Ensures that each output field is documented in the meta section under `meta.outputs`."
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    fn explanation(&self) -> &'static str {
+        "The meta section should have an output key and keys with descriptions for each output of \
```
claymcleod, line 106:
```suggestion
        "The meta section should have an `outputs` key and keys with descriptions for each output of \
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+         these entries should be in the same order (that order is up to the developer to decide). \
+         No extraneous output entries are allowed."
+    }
+
+    fn tags(&self) -> TagSet {
+        TagSet::new(&[Tag::Completeness])
+    }
+}
+
+/// Check each output key exists in the outputs key within the `meta` section.
```
claymcleod, line 118:
```suggestion
/// Check each output key exists in the `outputs` key within the `meta` section.
```

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: Agreed. I have tried to avoid that repetition, but I missed it here.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+fn check_output_meta(
```
claymcleod, line 119: One thing that occurs to me here is that, since you only implemented the callback for the `meta` section, it will be impossible for users to `#@ except` this rule for individual outputs. To achieve that, you'd have to base the rule off of, say, a `meta_item` callback. I'm not sure it's worth the effort to change until someone complains, but that's definitely something to be aware of.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+    state: &mut Diagnostics,
+    meta: &MetadataSection,
+    outputs: &OutputSection,
+    context: TaskOrWorkflow,
+) {
+    // Get the output section from the meta section.
```
claymcleod, line 125:
```suggestion
    // Get the `outputs` entries from the `meta` section.
```

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: Maybe this?

```suggestion
        "add a description of output `{name}` to documentation in `meta.outputs`"
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+    if let Some(meta_outputs_key) = meta
+        .items()
+        .find(|entry| entry.name().syntax().to_string() == "outputs")
+    {
+        let actual: IndexMap<_, _> = meta_outputs_key
+            .value()
+            .unwrap_object()
```
claymcleod, line 132: This will panic if the user sets the `meta` key to anything that isn't an object, which I don't think is what we want. It may be worth adding a diagnostic in the case where the `meta.outputs` key is anything but an object.

wdl-lint/src/rules/nonmatching_output.rs
claymcleod: Yeah I'm good with that.

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: While it was not an intentional design, we've been taking an opinionated stance on things. Not allowing users to skip documenting an individual output feels like a reasonable thing to prevent.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+            .items()
+            .map(|entry| {
+                (
+                    entry.name().syntax().to_string(),
+                    entry.syntax().text_range().to_span(),
+                )
+            })
+            .collect();
+
+        // Get the declared outputs.
```
claymcleod, line 142: I'd just delete this comment and call the variables `expected_outputs` and `actual_outputs`. Partially because this comment I think should be moved to where you get the actual outputs, partially because I feel the variable names are sufficient to describe the functionality here without a comment.

wdl-lint/src/rules/nonmatching_output.rs
claymcleod: I think it would only be possible to except extraneous items in `meta.outputs`. But, that said, I am fine with that magnitude of opinion being enforced for this particular rule.

FWIW, I did allow that flexibility in my runtime keys rule. Perhaps it's a bit more reasonable to allow there.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+                (
+                    d.name().syntax().to_string(),
+                    d.syntax().text_range().to_span(),
+                )
+            })
+            .collect();
+
+        let mut extra_found = false;
+
+        // Check for entries missing from `meta`.
```
claymcleod, line 155:
```suggestion
        // Check for entries missing from `meta.outputs`.
```

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+        for (name, span) in &expected {
+            if !actual.contains_key(name) {
+                if !extra_found {
```
claymcleod, line 158: Should this be `missing_found`?

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+                    extra_found = true;
+                }
+                state.add(nonmatching_output(*span, name, &context));
+            }
+        }
+
+        // Check for extra entries in `meta`.
```
claymcleod, line 165:
```suggestion
        // Check for extra entries in `meta.outputs`.
```

wdl-lint/src/rules/nonmatching_output.rs
claymcleod: Or maybe the variable represents something broader, such as `exact_match`?

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+        if !extra_found && !actual.keys().eq(expected.keys()) {
+            state.add(out_of_order(
+                meta_outputs_key.syntax().text_range().to_span(),
+                outputs.syntax().text_range().to_span(),
+                &context,
+            ));
+        }
+    } else {
+        state.add(missing_outputs_in_meta(
+            meta.syntax().first_token().unwrap().text_range().to_span(),
```
claymcleod, line 185: Personally, I think the span here should be the full meta section (`span_of(meta)`), but I know we've taken a different approach on other rules so I am fine if you want to keep it this way.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+        task: &TaskDefinition,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        // If the `meta` section is missing, the MissingMeta rule will handle it.
+        if let Some(meta) = task.metadata().next() {
+            // If the `output` section is missing, the MissingOutput rule will handle it.
+            if let Some(output) = task.outputs().next() {
```
claymcleod, line 242:
```suggestion
            if let Some(outputs) = task.outputs().next() {
```

to match the variable name in `check_output_meta`.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,249 @@
...
+        workflow: &WorkflowDefinition,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        // If the `meta` section is missing, the MissingMeta rule will handle it.
+        if let Some(meta) = workflow.metadata().next() {
+            // If the `output` section is missing, the MissingOutput rule will handle it.
+            if let Some(output) = workflow.outputs().next() {
```
claymcleod, line 216:
```suggestion
            if let Some(outputs) = workflow.outputs().next() {
```

to match the variable name in `check_output_meta`.

wdl-lint/tests/lints/missing-preamble-ws/source.wdl
claymcleod: Oh okay, sounds good.

wdl-lint/tests/lints/nonmatching-output/source.wdl
```diff
@@ -0,0 +1,65 @@
+#@ except: DescriptionMissing, MissingRuntime
```
claymcleod, line 1: We should test that:

- no diagnostics are output for an empty `output` section.
- what happens when the `meta.outputs` key is not an object.
- what happens when the user excepts the rule?

claymcleod (changes requested):
Overall looks good.

For the wording nitpicks, it was just an attempt to improve things, but I'm not confident they are the right direction: just for your consideration, feel free to take of leave what you want.

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: I'm torn. This is an `OutputSection`, but is accessed with the `outputs()` method. I'm inclined to make the opposite update and make it consistently `output` when referring to the `OutputSection`.

a-frantz: Throwing my 2 pennies in the ring: I like universally saying `output` (no 's')

wdl-lint/tests/lints/nonmatching-output/source.wdl
adthrasher: I added exceptions with comments to indicate which work and which don't.

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: Another question, what does an `#@ except` on a single entry mean when you're looking at the sorting?

wdl-lint/tests/lints/nonmatching-output/source.wdl
a-frantz, line 1: What does and does not work for the `#@ except` comments doesn't seem intuitive to me. I think a lint directive above `meta {}` and also above `output {}` should work for disabling this rule. Those are both the most intuitive (to me) places to put this lint directive. As a user I wouldn't immediately think to put it above the task definition (though that does make sense).

I think it's ok to say lint directives _within_ the meta/output section don't work. Maybe I'd call that support a "nice to have" but not a "need".

adthrasher: I'm look at it now. It's going to require a number of new visitors. So it's not a small change.

adthrasher: What is the expected output for this? Currently, `s` and `t` are flagged in `meta.outputs` as extraneous, since the `outputs` section is not processed.

```
task w {
    meta {
        outputs: {
            s: "s",
            t: "t",
        }
    }
    command <<< >>>

    #@ except: NonmatchingOutput
    output {
        String s = "hello"
        String t = "world"
        # Also here
        String v = "!"
    }
}
```

a-frantz: My thinking is that an `except` lint directive above either `meta` or `output` is functionally equivalent to turning off the lint for the entire task.

Does anyone have a different idea?

adthrasher: In an attempt to handle @claymcleod 's suggestion, I'm looking for the `output` section in a `output_section` callback and then looking at the declarations in a `bound_decl` callback. This is so that you can skip individual outputs from the check. So when you put the `#@ except` above `output`, I never get the `output_section` callback and can't store the output declarations. So it triggers a warning for everything in `meta.outputs`. I'm open to alternative ideas on how to make this work.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -75,30 +59,20 @@ fn extra_output_in_meta(span: Span, name: &str, context: &TaskOrWorkflow) -> Dia
...
 /// Creates a diagnostic for non-object `meta.outputs` entries.
-fn non_object_meta_outputs(span: Span, context: &TaskOrWorkflow) -> Diagnostic {
-    let (ty, item_name) = match context {
-        TaskOrWorkflow::Task(t) => ("task", t.name().as_str().to_string()),
-        TaskOrWorkflow::Workflow(w) => ("workflow", w.name().as_str().to_string()),
-    };
-
+fn non_object_meta_outputs(span: Span, item_name: String, ty: &str) -> Diagnostic {
     Diagnostic::warning(format!(
         "`outputs` key in `meta` section for the {ty} `{item_name}` is not an object"
```
a-frantz, line 77: This strikes me as potentially confusing for a user. The most likely scenario for this diagnostic (in my estimation) is that the user is using the `outputs` key with some different convention than what we expect. They look at this message here and it wouldn't make sense to them. I'd suggest changing to something along the lines of:
```suggestion
        "`outputs` key in `meta` section  is reserved for an object with keys corresponding to declared `output` values"
```
And as an aside, isn't `item_name` always going to be `outputs`?

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,351 @@
...
+    fn workflow_definition(
+        &mut self,
+        state: &mut Self::State,
+        reason: VisitReason,
+        workflow: &WorkflowDefinition,
+    ) {
+        match reason {
+            VisitReason::Enter => {
+                self.name = Some(workflow.name().as_str().to_string());
+                self.ty = Some("workflow");
```
a-frantz, line 198: Could we follow the example of other lint rules and wrap this in an `enum Context` instead of passing around `str`?

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: `item_name` has either the `task` or `workflow` name.

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: We probably could, but that seems a heavy solution to ultimately print `task` or `workflow`. Before going that route, I'd actually prefer if we could store a `TaskOrWorkflow` instance and use that instead, rather than creating a new `enum` to handle it.

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: @a-frantz I merged the new and old language. It's a bit wordy in the lint messages, but I think it's clear now.

a-frantz (approved):

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,349 @@
...
+                    state.add(missing_outputs_in_meta(
+                        self.current_meta_span.expect("should have a `meta` span"),
+                        self.name.as_deref().expect("should have a name"),
+                        self.ty.expect("should have a type"),
+                    ));
+                } else {
+                    check_matching(state, self);
+                }
+
+                self.name = None;
```
claymcleod, line 212: Since this is common with the code in the `task_definition` callback, I might pull it out into its own function.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,349 @@
...
+        &mut self,
+        _state: &mut Self::State,
+        reason: VisitReason,
+        section: &MetadataSection,
+    ) {
+        match reason {
+            VisitReason::Enter => {
+                self.current_meta_span = Some(section.syntax().text_range().to_span());
+            }
+            VisitReason::Exit => {}
```
claymcleod, line 266: Just to be explicit and avoid future issues, would this be a place to set `self.current_meta_span = None`? It's probably inconsequential unless something is terribly wrong, but it's worth considering being explicit.

If the answer above is "no", then I feel this should at least be an if statement like in `output_section`.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,349 @@
...
+        }
+    }
+
+    fn output_section(
+        &mut self,
+        _state: &mut Self::State,
+        reason: VisitReason,
+        section: &OutputSection,
+    ) {
+        if reason == VisitReason::Enter {
```
claymcleod, line 276: Same consideration here.

wdl-lint/src/rules/nonmatching_output.rs
```diff
@@ -0,0 +1,349 @@
...
+    fn bound_decl(
+        &mut self,
+        _state: &mut Self::State,
+        reason: VisitReason,
+        decl: &wdl_ast::v1::BoundDecl,
+    ) {
+        if reason == VisitReason::Enter {
+            if let Some(output) = &self.current_output_span {
+                let decl_span = decl.syntax().text_range().to_span();
+                if decl_span.start() > output.start() && decl_span.end() < output.end() {
```
claymcleod, line 290: That's one way to check if the decl is within the output 😄.

Based on my last comment, you might could check if `self.current_output_span` is `None` if you took that approach.

claymcleod (approved):

wdl-lint/src/rules/nonmatching_output.rs
adthrasher: If so, it would require a refactor of the `Exit` logic for both `Task` and `Workflow`.

wdl-lint/src/lib.rs
```diff
@@ -95,6 +95,7 @@ pub fn rules() -> Vec<Box<dyn Rule>> {
         Box::<rules::DeprecatedPlaceholderOptionRule>::default(),
         Box::<rules::RuntimeSectionKeysRule>::default(),
         Box::<rules::TodoRule>::default(),
+        Box::new(rules::NonmatchingOutputRule::default()),
```
a-frantz, line 98:
```suggestion
        Box::<rules::NonmatchingOutputRule>::default(),
```

PR #119, opened 2024-07-10
feat: implements the `runtime_section_keys` lint rule

claymcleod:
This pull request adds two new rules to `wdl-lint`.

- **Rule Name**: `DeprecatedPlaceholderOption`
- **Rule Name**: `RuntimeSectionKeys`

These rules are part of the [baseline lint rules](https://github.com/stjude-rust-labs/wdl/discussions/11). There were other changes that were introduced that were complementary to adding this lint, and all of those have been called out in my notes below.

The biggest changes were:

- I pulled out the `Version` from a private struct within the AST validation to a public struct (named `SupportedVersion` to avoid confusion with the `Version` AST node). This provides a unified way to parse a WDL version so that it can be used across the libraries (and, in particular, the lint rules).
- I now sort the diagnostics that are output by the tests to ensure the order is deterministic.

Before submitting this PR, please make sure:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.

Rule specific checks:

- [x] You have added the rule as an entry within `RULES.md`.
- [x] You have added the rule to the `rules()` function in `wdl-lint/src/lib.rs`.
- [x] You have added a test case in `wdl-lint/tests/lints` that covers every
      possible diagnostic emitted for the rule within the file where the rule
      is implemented.
- [x] You have run `wdl-gauntlet --refresh` to ensure that there are no 
      unintended changes to the baseline configuration file (`Gauntlet.toml`).
- [x] You have run `wdl-gauntlet --refresh --arena` to ensure that all of the 
      rules added/removed are now reflected in the baseline configuration file 
      (`Arena.toml`).

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.0.0/

PR #115, opened 2024-07-09
feat: implements the `deprecated_placeholder_option` lint rule

claymcleod:
This pull request adds a new rule to `wdl-lint`.

- **Rule Name**: `DeprecatedPlaceholderOption`

This rule is part of the https://github.com/stjude-rust-labs/wdl/discussions/11.

Before submitting this PR, please make sure:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [ ] Your code builds clean without any errors or warnings.
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.

Rule specific checks:

- [x] You have added the rule as an entry within `RULES.md`.
- [x] You have added the rule to the `rules()` function in `wdl-lint/src/lib.rs`.
- [ ] You have added a test cases in `wdl-lint/tests/lints` that covers every 
      possible diagnostic emitted for the rule within the file where the rule  
      is implemented.
- [x] You have run `wdl-gauntlet --refresh` to ensure that there are no 
      unintended changes to the baseline configuration file (`Gauntlet.toml`).
- [x] You have run `wdl-gauntlet --refresh --arena` to ensure that all of the 
      rules added/removed are now reflected in the baseline configuration file 
      (`Arena.toml`).

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.0.0/

PR #120, opened 2024-07-10
 feat: implements multiple lint rules

claymcleod:
This pull request adds two new rules to `wdl-lint`.

- **Rule Name**: `DeprecatedPlaceholderOption`
- **Rule Name**: `RuntimeSectionKeys`

These rules are part of the [baseline lint rules](https://github.com/stjude-rust-labs/wdl/discussions/11). There were other changes that were introduced that were complementary to adding this lint, and all of those have been called out in my notes below.

The biggest changes were:

- I pulled out the `Version` from a private struct within the AST validation to a public struct (named `SupportedVersion` to avoid confusion with the `Version` AST node). This provides a unified way to parse a WDL version so that it can be used across the libraries (and, in particular, the lint rules).
- I now sort the diagnostics that are output by the tests to ensure the order is deterministic.

Rule specific checks:

      is implemented.
      (`Arena.toml`).


claymcleod:
Apologies for the duplicates PRs (#115 and #119)—GitHub and I had a moment. I'll apply my previous comments here.

.github/pull_request_template.md
```diff
@@ -28,6 +28,11 @@ Rule specific checks:
 - [ ] You have added a test case in `wdl-lint/tests/lints` that covers every
       possible diagnostic emitted for the rule within the file where the rule
       is implemented.
+- [ ] If you have implemented a new `Visitor` callback, you have also
```
claymcleod, line 31: This was added after discussion with @peterhuene as its a common misstep when introducing a new visitor method.

wdl-ast/src/lib.rs
```diff
@@ -55,11 +55,13 @@ pub use wdl_grammar::ToSpan;
 pub use wdl_grammar::WorkflowDescriptionLanguage;
 
 pub mod v1;
+pub mod version;
 
 mod validation;
 mod visitor;
 
 pub use validation::*;
+pub use version::SupportedVersion;
```
claymcleod, line 64: This is where I pulled the `Version` out of validation and called it `SupportedVersion`.

wdl-ast/src/lib.rs
```diff
@@ -55,11 +55,13 @@ pub use wdl_grammar::ToSpan;
 pub use wdl_grammar::WorkflowDescriptionLanguage;
 
 pub mod v1;
+pub mod version;
```
claymcleod, line 58: The `V1` enum is in here now available at the top level under `version`.

wdl-grammar/src/diagnostic.rs
```diff
@@ -83,6 +84,38 @@ pub struct Diagnostic {
     labels: Vec<Label>,
 }
 
+impl Ord for Diagnostic {
```
claymcleod, line 87: Importantly, this PR includes some unexpected-but-important changes regarding the ordering of diagnostics to make the output of the tests deterministic. This will change the order of many other unrelated tests.

wdl-lint/tests/lints.rs
```diff
@@ -126,7 +126,13 @@ fn run_test(test: &Path, ntests: &AtomicUsize) -> Result<(), String> {
         )
     })?;
 
-    let (document, diagnostics) = Document::parse(&source);
+    let (document, mut diagnostics) = Document::parse(&source);
+
+    // NOTE: sorting of diagnostics is required to ensure a
```
claymcleod, line 131: This is where I added the deterministic sorting of errors.

wdl-lint/tests/lints.rs
```diff
@@ -138,7 +144,13 @@ fn run_test(test: &Path, ntests: &AtomicUsize) -> Result<(), String> {
         validator.add_visitor(LintVisitor::default());
         let errors = match validator.validate(&document) {
             Ok(()) => String::new(),
-            Err(diagnostics) => format_diagnostics(&diagnostics, &path, &source),
+            Err(mut diagnostics) => {
+                // NOTE: sorting of diagnostics is required to ensure a
```
claymcleod, line 148: And, subsequently, the sorting of lint warnings.

wdl-lint/tests/lints/command-mixed-line-cont/source.wdl
```diff
@@ -1,4 +1,4 @@
-#@ except: NoCurlyCommands, DescriptionMissing
+#@ except: NoCurlyCommands, DescriptionMissing, RuntimeSectionKeys
```
claymcleod, line 1: From here on down, it's mostly excepting the new rule and changes to the ordering of non-related tests. You can take a look at `runtime-keys-wdl-1.0` and `runtime-keys-wdl-1.1` for the most important new tests.

wdl-lint/RULES.md
```diff
@@ -7,10 +7,11 @@ be out of sync with released packages.
 ## Lint Rules
 
 | Name                             | Tags                          | Description                                                                           |
-|:---------------------------------|:------------------------------|:--------------------------------------------------------------------------------------|
+| :------------------------------- | :---------------------------- | :------------------------------------------------------------------------------------ |
```
claymcleod, line 10: This is due to the Markdown All-in-One extension formatting from within VSCode. I suppose it's fine unless anyone has any issues—that makes it much easier to keep the spacing consistent (for me at least).

claymcleod:
@peterhuene and @adthrasher, though this thread is on the now-closed PR, [this](https://github.com/stjude-rust-labs/wdl/pull/119#discussion_r1673140726) still needs an opinion from you.

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,355 @@
...
+        "include an entry for the `{key}` key in the `runtime` section"
+    ))
+}
+
+/// Inspects a runtime key for issues within v1.1 WDL documents.
+fn inspect_v1_1_key(state: &mut Diagnostics, key: &str, span: Span, specification: &str) {
+    let entries = V1_1_KEYS
+        .iter()
+        .filter(|entry| entry.name() == key)
+        .collect::<Vec<_>>();
```
peterhuene, line 158: As `inspect_v1_1_key` is called once per runtime item, is there a way we could avoid the `collect` here?

Such as:

```rust
let mut entries = V1_1_KEYS.iter().filter(|e| e.name() == key);
let entry = entries.next();
assert!(entries.next().is_none(), "more than one key matched; duplicate entry in static key map");
match entry {
...
}
```

Also, it seems like the key sets could instead be a `HashMap<&'static str, Key>` where `Key` would be:

```rust
enum Key {
    /// A key that is deprecated in favor of another key.
    /// The value is the replacement key.
    Deprecated(&'static str),
    /// A runtime key that is recommended to be included.
    Recommended,
    /// A runtime key that has a reserved meaning in the specification but which
    /// execution engines are _not_ required to support. These are also called
    /// "hints" in WDL parlance.
    Hint,
    /// A runtime key that has a reserved meaning in the specification and which
    /// execution engines are required to support (but don't necessarily have to
    /// be present in WDL documents).
    ReservedMandatory,
}
```

That way we don't need to iterate the collection at all.

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,355 @@
...
+                let runtime_span = match self.current_runtime_span {
+                    Some(span) => span,
+                    None => return,
+                };
+
+                if let Some(SupportedVersion::V1(minor_version)) = self.version {
+                    let recommended_keys = match minor_version {
+                        V1::Zero => recommended_keys(&V1_0_KEYS.clone()),
+                        V1::One => recommended_keys(&V1_1_KEYS.clone()),
+                        _ => Default::default(),
```
peterhuene, line 283: This seems like a lot of allocations, a `clone` here, and then both a `clone` and `collect` in `recommended_keys`.

An alternative might be:

```rust
let empty = Default::default();
let recommended_keys = match minor_version {
    V1::Zero => &V1_0_KEYS,
    V1::One => &V1_1_KEYS,
    _ => &empty,
};

for key in recommended_keys.iter().filter(|key| key.is_recommended())  {
    ...
}
```

wdl-lint/tests/lints/deprecated-placeholder-options-v1.1/source.errors
```diff
@@ -0,0 +1,48 @@
+warning[DeprecatedPlaceholderOption]: use of the deprecated `sep` placeholder option
+   ┌─ tests/lints/deprecated-placeholder-options-v1.1/source.wdl:10:32
+   │
+10 │     String bad_sep_option = "~{sep="," numbers}"
+   │                                ^^^^^^^
```
peterhuene, line 5: I think the span should probably be just the identifier of the node (e.g. `sep`), rather than the entire node.

We can probably add a `name(&self) -> Ident` method to each of the options to facilitate this.

peterhuene (commented):
This looks great; a few comments to go over.

wdl-lint/tests/lints/deprecated-placeholder-options-v1.1/source.errors
peterhuene: On second thought, given that the remediation for the diagnostic is to remove the entire option, having the highlight be the entire option makes sense to me. 

Resolving this comment.

Cargo.toml
```diff
@@ -28,6 +28,7 @@ serde = { version = "1", features = ["derive"] }
 serde_with = "3.8.1"
 tokio = { version = "1.38.0", features = ["full"] }
 toml = "0.8.14"
+lazy_static = "1.5"
```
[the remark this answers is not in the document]
claymcleod: Oh, I was not aware of that. Okay, thanks for the tip—I'll try that instead.

wdl-lint/src/rules/runtime_section_keys.rs
claymcleod: So I originally had it the way you describe above and ended up flipping back and forth. There were some pros I liked in terms of the clarity of the code with the `HashSet` option (from the top of my head, duplicates keys would be able to be detected rather than re-inserting over an existing key + pulling out the `DeprecatedKey` made it possible to statically verify only deprecated keys could be passed to `deprecated_runtime_key` for future implementers).

All that said, I made those decisions because I felt the performance hit was small compared to a small bump up in clarity of the code but, as you and I discussed, we can take the philosophy of eeking out every bit of performance here.

wdl-lint/src/rules/runtime_section_keys.rs
claymcleod: Yeah, this is a case of not coming back to fix this: there were some quirks of `lazy_static` that I was looking at trying to avoid this, but I see now that I got a working solution and never actually came back to it. With `std::sync::OnceLock`, this appears to be much more straightforward.

Arena.toml
```diff
@@ -1130,6 +1155,31 @@ document = "stjudecloud/workflows:/data_structures/read_group.wdl"
...
 permalink = "https://github.com/stjudecloud/workflows/blob/efdca837bc35fe5647de6aa95989652a5a9648dc/data_structures/read_group.wdl/#L13"
 
+[[diagnostics]]
+document = "stjudecloud/workflows:/data_structures/read_group.wdl"
+message = "read_group.wdl:140:5: note[RuntimeSectionKeys]: the `cpu` runtime key is recommended by the WDL v1.1 specification"
+permalink = "https://github.com/stjudecloud/workflows/blob/efdca837bc35fe5647de6aa95989652a5a9648dc/data_structures/read_group.wdl/#L140"
+
+[[diagnostics]]
+document = "stjudecloud/workflows:/data_structures/read_group.wdl"
+message = "read_group.wdl:140:5: note[RuntimeSectionKeys]: the `disks` runtime key is recommended by the WDL v1.1 specification"
```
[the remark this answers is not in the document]
claymcleod: Yeah I recall this quirk (maybe Cromwell uses `disk` but WDL settled on `disks`)? Should this special case mention Cromwell? I need some input from you and @a-frantz on the particulars around this case.

wdl-lint/src/rules/runtime_section_keys.rs
peterhuene: If detecting duplicates on init is desired, I think we could structure initialization to look like:

```rust
V1_1_KEYS.get_or_init(|| {
    const KEYS: &[(&str, Key)] = [("container", Key::ReservedMandatory), ...];

    let mut keys = HashMap::with_capacity(KEYS.len());
    for (k, v) in KEYS {
        let prev = keys.insert(k, v);
        assert!(prev.is_none(), "duplicate key `{k}`");
    }
    
    keys
});
```

Arena.toml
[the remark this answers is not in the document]
a-frantz: It's an archaic Cromwell quirk that has persisted in our repo and I'm not sure it it's present anywhere else. [The current state of affairs](https://cromwell.readthedocs.io/en/stable/RuntimeAttributes/#disk) is that even Cromwell has dropped support for `disk` (mostly). I don't remember offhand the exact history, but I'm not sure it matters at this point.

I'm about to look through Arena to see if this is present in any repo other than ours. I suspect it won't be. I'm not certain this needs any special handling. We just need to fix it in `workflows`, and then hopefully we won't see this "in the wild" anywhere else.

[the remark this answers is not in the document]
a-frantz: Yea that's what I thought. I expanded Arena to look through every Gauntlet repo, and then `grep`ed the resulting Arena.toml for this diagnostic. It only appears in `workflows`. I still think the current diagnostics emmitted by `RuntimeSectionKeys` needs to be consolidated, but no special handling for `disk`/`disks` is needed IMO.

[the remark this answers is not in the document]
claymcleod: Okay, that sounds good, I will roll with Andrew F.'s suggestion if that's good with you @adthrasher

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,355 @@
...
+    /// A runtime key that is recommended to be included.
+    Recommended(&'static str),
+    /// A runtime key that has a reserved meaning in the specification but which
+    /// execution engines are _not_ required to support. These are also called
+    /// "hints" in WDL parlance.
+    Hint(&'static str),
+    /// A runtime key that has a reserved meaning in the specification and which
+    /// execution engines are required to support (but don't necessarily have to
+    /// be present in WDL documents).
+    ReservedMandatory(&'static str),
```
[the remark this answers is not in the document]
claymcleod: It's a good question and I considered that briefly too. To me, since this will be hidden as a behind-the-scenes implementation detail combined with the fact that the specification refers to them as "reserved, mandatory keys", I felt it was okay to go with this.

claymcleod:
I've now hammered out a draft of all of the changes that were requested on #120—there may be some sloppiness, such as unnecessary allocations that can be cleaned up, but I'm going to push it up now so others can review the overall output with the new implementation. I'll take a look for any optimizations tomorrow.

wdl-analysis/tests/analysis.rs
```diff
@@ -110,7 +110,7 @@ fn compare_results(test: &Path, results: Vec<AnalysisResult>) -> Result<()> {
         // Attempt to strip the CWD from the result path
         let path: Cow<'_, str> = match &path {
             // Strip the CWD from the path
-            Some(path) => path.strip_prefix(&cwd).unwrap_or(&path).to_string_lossy(),
+            Some(path) => path.strip_prefix(&cwd).unwrap_or(path).to_string_lossy(),
```
claymcleod, line 113: FWIW, this was a stray lint warning that clippy found and I just corrected here.

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,501 @@
...
+            // that we would hit this condition for a task with two `runtime`
+            // sections. Here, we simply ignore any `runtime` section but the
+            // first one, as multiple `runtime` sections will be flagged as an
+            // error by the validation.
+            Some(_) => {}
+            None => self.current_runtime_span = Some(span_of(section)),
+        }
+    }
+
+    fn runtime_item(&mut self, state: &mut Self::State, reason: VisitReason, item: &RuntimeItem) {
```
adthrasher, line 432: What happens if someone does this?
```
task foo {
  .... 
  runtime {
    cromwell: {
      # cromwell specific 
      ...
    }
    miniwdl: {
      # miniwdl specific
      ...
    }
  }
}
```
That's an example in the spec (https://github.com/openwdl/wdl/blob/1.1.1/SPEC.md#conventions-and-best-practices).

claymcleod: That's a good question. The specification is clear that any non-reserved keys are deprecated:

> 🗑 Additional arbitrary attributes may be specified in the runtime section, but these may be ignored by the execution engine. These non-standard attributes are called "hints". The use of hint attributes in the runtime section is deprecated; a later version of WDL will introduce a new hints section for arbitrary attributes and disallow non-standard attributes in the runtime section.

That said, the example does still exist in the specification. My feeling is that you should be able to `#@ except` those entries.

a-frantz: > That said, the example does still exist in the specification. My feeling is that you should be able to `#@ except` those entries.

Sounds fine to me. I ofc have not memorized the docs for all the execution engines out there, but I don't recall ever actually seeing anyone adopt this convention. I think it's a weird hypothetical that most likely doesn't belong in the spec in the first place.

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,502 @@
...
+        // WDL version and will be reported by another validation rule. As such,
+        // in that case, we can skip reporting diagnostics within this rule.
+        if let Some(SupportedVersion::V1(minor_version)) =
+            self.version.expect("the document version should be known")
+        {
+            // The only keys that actually need to be individually inspected are
+            // WDL v1.1 keys because,
+            //
+            // - WDL v1.0 contains no deprecated keys: the only issue that can occur is when
+            //   one of the two recommendeds key is omitted, and that is handled at the end
```
peterhuene, line 454:
```suggestion
            //   one of the two recommended keys is omitted, and that is handled at the end
```

wdl-lint/src/rules/deprecated_placeholder_option.rs
```diff
@@ -0,0 +1,162 @@
...
+#[derive(Debug, Default, Clone, Copy)]
+pub struct DeprecatedPlaceholderOptionRule {
+    /// The detected version of the current document.
+    ///
+    /// The first [`Option`] represents whether the version has been hooked or
+    /// not by this rule.
+    ///
+    /// The second [`Option`] represents whether or not a valid version was able
+    /// to be parsed from the version string.
+    version: Option<Option<SupportedVersion>>,
```
peterhuene, line 69: I'm not quite sure why this is `Option<Option<SupportedVersion>>`. Rules may expect that `document` is called before any other callbacks as it is the root of the walk, so should this not be `Option<SupportedVersion>`?

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,502 @@
...
+    fn task_definition(
+        &mut self,
+        state: &mut Self::State,
+        reason: VisitReason,
+        _: &TaskDefinition,
+    ) {
+        match reason {
+            VisitReason::Enter => {
+                self.current_runtime_span = None;
+                self.encountered_keys = Default::default();
```
peterhuene, line 360: The fix for the issue you were seeing:

```suggestion
                self.encountered_keys.clear();
                self.non_reserved_keys.clear();
```

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,502 @@
...
+            return;
+        }
+
+        match self.current_runtime_span {
+            // Since validation and linting occur simultaneously, it is possible
+            // that we would hit this condition for a task with two `runtime`
+            // sections. Here, we simply ignore any `runtime` section but the
+            // first one, as multiple `runtime` sections will be flagged as an
+            // error by the validation.
+            Some(_) => {}
```
peterhuene, line 428: So I think this isn't quite correct, as we'll still process the items of the additional runtime sections.

I think we should have a flag, reset on task entry, which indicates that we've exited from a runtime section; if that flag is set, have future calls to `runtime_section` and `runtime_item` return early.

peterhuene (commented):
Looks great, just four more additional comments.

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,485 @@
...
+
+    fn description(&self) -> &'static str {
+        "Ensures that `runtime` sections have the appropriate keys."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "The behavior of this rule is different depending on the WDL version:
+        
+        For WDL v1.0 documents, the `docker` and `memory` keys are recommended, but the inclusion \
+         of any number of other keys is encouraged.
```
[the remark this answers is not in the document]
claymcleod: I ended up changing the language before I saw this comment to exactly what you say here.

wdl-lint/src/rules/deprecated_placeholder_option.rs
claymcleod: You're correct that this assumption _should_ always remain true. The virtue that lead me to split into two `Option`s is simply to be able to handle the two different meanings of the previously overloaded `None` from the rule's perspective. Explicitly, collapsing the meanings and simply having the rule ignore things if the value is `None` is _okay_, but I felt it was a bit better to panic from the rule if the assumption it has is ever broken.

[`serde_with` uses a similar concept which I like quite a lot](https://docs.rs/serde_with/latest/serde_with/rust/double_option).

Though it introduces a small bit of complexity, it also makes things more explicit and provides an additional fail-safe if that assumption is every broken, so I'd argue to keep it (though I could be persuaded not to if the team felt strongly otherwise).

peterhuene: I think by the nature of the CST having a root node and the root node being the "document", this invariant is very unlikely to ever be violated unless something is catastrophically wrong.

If we change this to `Option<SupportedVersion>`, then if somehow we violate the invariant, the rule won't fire a diagnostic rather than panicking like it does now with the double option; either way we'll observe the problem in our test coverage.

To me it just complicates the code (albeit it only slightly) and I certainly do a double take when I see a double option.

wdl-lint/src/rules/runtime_section_keys.rs
peterhuene: There is a distinction between assigning `Default::default()` and calling `clear()`: the latter will preserve the capacity. 

While it's a minor thing, I think reducing allocations when it's trivial to do so would be ideal. That said, I don't think it's worth holding up the PR unless you'd like to fix.

wdl-lint/src/rules/deprecated_placeholder_option.rs
claymcleod: Okay, I've finally found some time to dig into the details of the packages and how they work together, and now I have a better idea of the inner mechanisms here. But first, since this may come up as we continue to jointly develop out the package as a group, I should briefly share my perspective on the first part of your comment.

> I think by the nature of the CST having a root node and the root node being the "document", this invariant is very unlikely to ever be violated unless something is catastrophically wrong.
>
> If we change this to Option<SupportedVersion> and somehow we violate the invariant, the rule won't fire a diagnostic rather than panicking like it does now with the double option; either way we'll observe the problem in our test coverage.

To my eye, the difference in approach boils down to whether the benefit of being explicit (both in terms of the code as well as the error messages themselves) is worth the cost of introducing complexity or ugliness into the code.

As you say, it is quite unlikely that the assumption you laid out will ever be broken. However, in the case where it _is_ broken, I will always prefer to immediately get back an error telling me exactly what assumption was violated rather spending time tracking down why the majority of the tests are suddenly failing. Further, when there are two different reasons a `version` could be missing (in this case, because a catastrophic failure occurred in the parsing versus whether the version exists but failed to be parsed to a supporting version), I always prefer to make that immediately obvious and wrestle with those cases out in the open: the practical implications may be the same such that they could be collapsed, but the context with which I collapsed them is far more likely to be missed the next time I visit that code. You could encode those assumptions in comments, but I prefer for the code to yell at me when it can (😄).

As you might intuit from this perspective, I'm quite a defensive software engineer, as I've been bitten in big ways more than once by not taking this approach. However, I'm happy to be reigned in and accept the risks associated with not being explicit, as the likelihood that either approach will be consequential in the long term is small.

---

> Alternatively, we could also unwrap a Option<SupportedVersion> and continue to panic as visitors should not be receiving this callback on an unsupported document version (we skip parsing the contents of the file).

Now that I have a better understanding of what's going on underneath the hood, this comment spurred me to think that there is a better solution hidden behind a more fundamental question: do we want to allow `Visitor`s to traverse documents that aren't a supported WDL version?

To illustrate what happens today, take this document:

```
version 2.0

task {}
workflow {}
```

The root node and preamble are parsed, and the rest of the document is left as an unparsed node:

```
RootNode@0..33
  VersionStatementNode@0..11
    VersionKeyword@0..7 "version"
    Whitespace@7..8 " "
    Version@8..11 "2.0"
  Unparsed@11..33 "\n\ntask {}\nworkflow {}\n"
```

On its face, there is no reason not to allow this AST to be visited by the validator: documents with unsupported WDL versions produce an AST just the same as any supported versions. The big difference is that these ASTs only contain `RootNode` and the various version nodes, so only the `document()` and `version_statement()` callbacks will be executed.

That said, after giving it some deeper thought, I think there is very little benefit to allowing this behavior.

- I think there are very few pros to be had. Basically nothing can be confidently asserted about unsupported versions besides the fact that they are unsupported—any rules that do manage to execute could produce false positive or false negative results in a "silent" manner. Practically speaking, no callbacks besides `document()` and `version_statement_node()` would be run on this kind of AST, so none of the current rules will execute properly.
- The cost of allowing it means that every rule needs to now consider a greater surface area of whether the rule is valid for non-supported versions (a strange concept in general to me—the versions are unsupported by definition). This will lead to boilerplate that is needed in any version-dependent rule, which will become tiring.

Today, we almost always manually check if the parser produced any diagnostics before moving forward with validation/linting. For example, in the `wdl lint` command:

```
 async fn exec(self) -> Result<()> {
        let source = read_source(&self.path)?;
        let (document, diagnostics) = Document::parse(&source);

        // HERE, WE CHECK TO MAKE SURE NO DIAGNOSTICS OCCURRED WHILE PARSING
        // AND RETURN IF ANY ARE ENCOUNTERED.
        if !diagnostics.is_empty() {
            emit_diagnostics(&self.path.to_string_lossy(), &source, &diagnostics)?;

            bail!(
                "aborting due to previous {count} diagnostic{s}",
                count = diagnostics.len(),
                s = if diagnostics.len() == 1 { "" } else { "s" }
            );
        }

      
        let mut validator = Validator::default();
        validator.add_visitor(LintVisitor::default());
        if let Err(diagnostics) = validator.validate(&document) {
            emit_diagnostics(&self.path.to_string_lossy(), &source, &diagnostics)?;

            bail!(
                "aborting due to previous {count} diagnostic{s}",
                count = diagnostics.len(),
                s = if diagnostics.len() == 1 { "" } else { "s" }
            );
        }

        Ok(())
    }
```

There's only one place I spotted in the current implementation where we don't take this approach, and it's in [the engine code](https://github.com/stjude-rust-labs/wdl/blob/main/wdl-analysis/src/engine.rs#L552). When I was designing a solution that followed @peterhuene's suggestion to `unwrap()` the version, this was the cause of inconsistent behavior depending on what happened when I used `wdl lint` to test my solution versus when the engine took over during unit tests.

I'm still open to saying that we want to take the philosophy of allowing unsupported WDL documents to be visited, but, if we do that, we should probably go the full mile and implemented the `supported(version: Option<SupportedVersion>) -> bool` function to make this dance a little more graceful.

peterhuene: Agreed, I think having `supported` on the rules / visitors makes a lot of sense.

I'm in support of resolving this issue as you see fit as I know completely understand why it's a double option 👍 .

Regarding the way that analysis is behaving, we could certainly make that consistent with everything else and only extend the diagnostics for a validation visit if parsing didn't have any documents.

claymcleod: My preference is to roll with exactly what is implemented right now and merge it in so it makes the Sprocket Docket on Monday. My gut is that the `supported` callback is a bigger change, and this PR accomplishes enough such that we should kick that can down the road.

@a-frantz and @adthrasher, I plan to merge this Monday if there are no other objections, as Peter and I appear to land in the same place on this.

wdl-lint/src/rules/runtime_section_keys.rs
claymcleod: Good catch here, I corrected this.

peterhuene (approved):

wdl-lint/src/rules/runtime_section_keys.rs
```diff
@@ -0,0 +1,500 @@
...
+        "The behavior of this rule is different depending on the WDL version:
+        
+        For WDL v1.0 documents, the `docker` and `memory` keys are recommended, but the inclusion \
+         of any number of other keys is permitted.
+        
+        For WDL v1.1 documents, 
+        
+        - A list of mandatory, reserved keywords will be recommended for inclusion if they are not \
+         present. Here, 'mandatory' refers to the requirement that all execution engines support \
+         this key—not that the key must be present in the `runtime` section.
```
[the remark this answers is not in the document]
peterhuene: Accordingly, the only value without a default is `container` (`docker`):

> The user is strongly suggested to specify a container for every task. There is no default value for container. If container is not specified, the execution behavior is determined by the execution engine. Typically, the task is simply executed in the host environment.

a-frantz (approved):

PR #136, opened 2024-07-19
feat: adds a comment whitespace rule

adthrasher:

- **Rule Name**: `CommentWhitespace`

This ensures that comments are properly spaced. That is a single space follows the comment delimiter. Two spaces precede the delimiter for in-line comments. Also check the indentation level is appropriate for non-inline comments.

Rule specific checks:

      is implemented.
- [ ] If you have implemented a new `Visitor` callback, you have also
      (`Arena.toml`).


wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -0,0 +1,182 @@
...
+        // Reset the visitor upon document entry
+        *self = Default::default();
+    }
+
+    fn comment(&mut self, state: &mut Self::State, comment: &Comment) {
+        let re = Regex::new(r"^#+@?").unwrap();
+        if is_inline_comment(comment) {
+            // check preceding whitespace for two spaces
+            if let Some(prior) = comment.syntax().prev_sibling_or_token() {
+                if prior.kind() == SyntaxKind::Whitespace && prior.to_string() != "  " {
```
a-frantz, line 114: I think this will miss the case when an inline comment isn't preceded by whitespace. e.g.
```
meta {# this is a problematic yet valid comment
}
```

wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -0,0 +1,182 @@
...
+            unreachable!("A comment must start with a #")
+        }
+    }
+}
+
+/// Detect is a comment is in-line or not by looking for `\n` in the prior
+/// whitespace.
+fn is_inline_comment(token: &Comment) -> bool {
+    if let Some(prior) = token.syntax().prev_sibling_or_token() {
+        return prior.kind() == SyntaxKind::Whitespace && !prior.to_string().contains('\n')
```
a-frantz, line 179: Same problem here
```
workflow# a problematic inline comment
{}
```

wdl-lint/src/rules/comment_whitespace.rs
adthrasher: Good catch. That should be.

```suggestion
                if prior.kind() != SyntaxKind::Whitespace || prior.to_string() != "  " {
```

wdl-lint/src/rules/comment_whitespace.rs
adthrasher:
```suggestion
        return prior.kind() != SyntaxKind::Whitespace || !prior.to_string().contains('\n')
```

Arena.toml
```diff
@@ -140,6 +200,16 @@ document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
...
 
+[[diagnostics]]
+document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
+message = "ww-fastq-to-cram.wdl:62:5: note[CommentWhitespace]: in-line comments should be preceded by two spaces"
+permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/2d1e1989a57402642c06d15f4b623ac66fd9ed7d/ww-fastq-to-cram.wdl/#L62"
+
+[[diagnostics]]
+document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
+message = "ww-fastq-to-cram.wdl:64:3: note[CommentWhitespace]: comment not sufficiently indented"
+permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/2d1e1989a57402642c06d15f4b623ac66fd9ed7d/ww-fastq-to-cram.wdl/#L64"
```
a-frantz, line 211: using 2 spaces per indent level

Arena.toml
```diff
@@ -1110,16 +1325,156 @@ document = "getwilds/ww-vc-trio:/ww-vc-trio.wdl"
...
+
 [[diagnostics]]
 document = "getwilds/ww-vc-trio:/ww-vc-trio.wdl"
 message = "ww-vc-trio.wdl:98:29: note[CallInputSpacing]: call input not properly spaced"
 permalink = "https://github.com/getwilds/ww-vc-trio/blob/c2c13e85efda8ac7aca7f8765fbaa7c2cacb08b2/ww-vc-trio.wdl/#L98"
 
+[[diagnostics]]
+document = "stjudecloud/workflows:/data_structures/flag_filter.wdl"
+message = "flag_filter.wdl:30:1: note[CommentWhitespace]: comment delimiter should be followed by a single space"
+permalink = "https://github.com/stjudecloud/workflows/blob/129060215443bbd1b67c4d3d149d48acdab90ef2/data_structures/flag_filter.wdl/#L30"
```
a-frantz, line 1341: We need an exception for preamble comments.

Arena.toml
```diff
@@ -1194,3 +1549,63 @@ permalink = "https://github.com/stjudecloud/workflows/blob/129060215443bbd1b67c4
 document = "stjudecloud/workflows:/workflows/qc/quality-check-standard.wdl"
 message = "quality-check-standard.wdl:111:1: note[LineWidth]: line exceeds maximum width of 90"
 permalink = "https://github.com/stjudecloud/workflows/blob/129060215443bbd1b67c4d3d149d48acdab90ef2/workflows/qc/quality-check-standard.wdl/#L111"
+
+[[diagnostics]]
+document = "stjudecloud/workflows:/workflows/qc/quality-check-standard.wdl"
+message = "quality-check-standard.wdl:124:13: note[CommentWhitespace]: comment not sufficiently indented"
+permalink = "https://github.com/stjudecloud/workflows/blob/129060215443bbd1b67c4d3d149d48acdab90ef2/workflows/qc/quality-check-standard.wdl/#L124"
```
a-frantz, line 1556: This seems fine to me. What's triggering this one?

wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -0,0 +1,178 @@
...
+            if let Some(prior) = comment.syntax().prev_sibling_or_token() {
+                if prior.kind() != SyntaxKind::Whitespace || prior.to_string() != "  " {
+                    // Report a diagnostic if there are not two spaces before the comment delimiter
+                    state.add(inline_preceding_whitespace(comment.span()))
+                }
+            }
+        } else {
+            // Not an in-line comment, so check indentation level
+            let ancestors = comment.syntax().parent_ancestors().collect::<Vec<_>>();
+            let expected_indentation = INDENT.repeat(ancestors.len() - 1);
```
a-frantz, line 122: Wow. This is _so_ much simpler than what I'm doing in `wdl-format`. And according to the `Arena` results, it seems pretty spot on. i found one potential false positive, but that is a weirdly placed comment. This approach seems solid and is a lot easier to think about than what I'm doing. You've given me some food for thought here.

wdl-lint/tests/lints/comment-whitespace/source.wdl
a-frantz, line 1: Another test case I would like to see. Is this properly handled?
```
String foo = "bar"       # too much space for an inline comment
```

Arena.toml
adthrasher:
```
note[CommentWhitespace]: comment not sufficiently indented
    ┌─ ../workflows/workflows/qc/quality-check-standard.wdl:124:13
    │
124 │             # 0x4 (unmapped) || 0x100 (secondary) || 0x800 (supplementary)
    │             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    │
    = fix: this comment has 3 levels of indentation. It should have 4 levels of indentation.
```

Arena.toml
adthrasher: This probably needs something at a higher level to detect the indentation "type".

Arena.toml
adthrasher: This comment is a child of a `LiteralMapNode` since it is part of a `struct` declaration.

a-frantz: I think you would agree the "fix" here is wrong:
```
        FlagFilter comparative_filter = {
            "include_if_all": "0x0",
                # 0x4 (unmapped) || 0x100 (secondary) || 0x800 (supplementary)
            "exclude_if_any": "0x904",
            "include_if_any": "0x0",
            "exclude_if_all": "0x0",
        }
```
What should we do?

wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -0,0 +1,178 @@
...
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        // Reset the visitor upon document entry
+        *self = Default::default();
+    }
+
+    fn comment(&mut self, state: &mut Self::State, comment: &Comment) {
+        let re = Regex::new(r"^#+@?").unwrap();
```
claymcleod, line 110:
```suggestion
        let comment_start = Regex::new(r"^#+@?").unwrap();
```

wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -0,0 +1,178 @@
...
+            }
+        } else {
+            unreachable!("A comment must start with a #")
+        }
+    }
+}
+
+/// Detect is a comment is in-line or not by looking for `\n` in the prior
+/// whitespace.
+fn is_inline_comment(token: &Comment) -> bool {
```
claymcleod, line 173: Should probably have a small unit test to make sure this is working appropriately.

claymcleod: Like this: https://github.com/stjude-rust-labs/wdl/pull/138/files#diff-4406a4f641401d5eb6bd18bb1aaba2229755b82832da763fd464782f34e33d78R175

wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -0,0 +1,178 @@
...
+            // check preceding whitespace for two spaces
+            if let Some(prior) = comment.syntax().prev_sibling_or_token() {
+                if prior.kind() != SyntaxKind::Whitespace || prior.to_string() != "  " {
+                    // Report a diagnostic if there are not two spaces before the comment delimiter
+                    state.add(inline_preceding_whitespace(comment.span()))
+                }
+            }
+        } else {
+            // Not an in-line comment, so check indentation level
+            let ancestors = comment.syntax().parent_ancestors().collect::<Vec<_>>();
```
claymcleod, line 121: You just want to use the `count()` method on the iterator here, not cause an allocation by collecting into a `Vec`.

wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -0,0 +1,178 @@
...
+                    }
+                }
+            } else {
+                // If there is no prior whitespace, this comment must be at the
+                // start of the file.
+            }
+        }
+
+        // check the comment for one space following the comment delimiter
+        if let Some(delimiter) = re.captures(comment.as_str()) {
```
claymcleod, line 153: I feel like using a regex here is a bit heavy handed. What about this?

```
let mut comment_chars = comment
            .as_str()
            .trim_start_matches('#')
            .chars()
            .skip_while(|c| *c == '#')
            .peekable();

        if let Some('@') = comment_chars.peek() {
            comment_chars.next();
        }

        let n_whitespace = comment_chars.by_ref().take_while(|c| *c == ' ').count();

        if comment_chars.skip(n_whitespace).count() > 0 && n_whitespace != 1 {
            state.add(following_whitespace(Span::new(
                comment.span().start(),
                n_whitespace,
            )));
        }
```

Arena.toml
adthrasher: To me, the fix would actually be:
```
        FlagFilter comparative_filter = {
                "include_if_all": "0x0",
                # 0x4 (unmapped) || 0x100 (secondary) || 0x800 (supplementary)
                "exclude_if_any": "0x904",
                "include_if_any": "0x0",
                "exclude_if_all": "0x0",
        }
```

We can skip certain ancestors, but that's going to be more complicated. I suppose you could do something like checking each ancestor to see if it is separated from the prior ancestor by a newline. That might be sufficient.

a-frantz: I have a strong dislike for changing indent levels in increments >1. So I don't like that first suggestion on principle lol. So I pointed this out as being problematic over Slack, but this declaration is actually using map->struct coercion which IMO should not be permitted. Does the "correct" implementation have the same issue?
```
     FlagFilter comparative_filter = FlagFilter {
            include_if_all: "0x0",
            # 0x4 (unmapped) || 0x100 (secondary) || 0x800 (supplementary)
            exclude_if_any: "0x904",
            include_if_any: "0x0",
            exclude_if_all: "0x0",
    }
```
Can you check?

Arena.toml
adthrasher: Done. Preamble comments are now only required to have at least 1 space following the delimiter.

Arena.toml
adthrasher: Do we have a method for a user to pass a value during linting?

wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -0,0 +1,251 @@
...
+                                if next_next.kind() == SyntaxKind::Whitespace
+                                    && next_next.to_string().contains('\n')
+                                {
+                                    return true;
+                                }
+                            }
+                        }
+                    }
+                    false
+                })
```
a-frantz, line 142: Can you please refactor this chunk into its own function? `filter_parent_ancestors(a: SyntaxNode) -> bool` I think would be the signature. Asking so that it can be unit tested. I have a hunch that L131 needs to include `SyntaxKind::OpenBracket` and `OpenBrace` as well as `OpenParen` but I'm not totally sure.

a-frantz (commented):
Other than my one comment, this looks really good

wdl-lint/src/rules/comment_whitespace.rs
adthrasher: I pulled that logic out and put in a number of test cases. I updated the logic as well. I'm fairly certain that the special case only applies to `ParenthesizedExpressionNode`s because those are the only ones that can be part of a arithmetic expression, where a `\n` placed before the operator token appears inside the `ExpressionNode` itself. That seems to be the tricky case.

a-frantz: What if an expression contains an array, map, or object? All those are possible, and wouldn't they contribute a level to indentation?

adthrasher: None of those are permissible on the RHS of an expression (other than `=`, `==`, and `!=`) right?

adthrasher: @a-frantz  - I added test cases with array and map/object expressions. I think everything is working as expected.

a-frantz (approved):

PR #137, opened 2024-07-19
feat: trailing comma lint rule

adthrasher:

- **Rule Name**: `TrailingComma`

This rule enforces that the last item in a MetadataObject or MetadataArray has a trailing comma. This only applies if there is > 1 child.

Rule specific checks:

      is implemented.
      (`Arena.toml`).

### DIFF
--- a/Arena.toml
+++ b/Arena.toml
@@ -162,6 +162,11 @@ permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/2d1e1989a57402642
 
 [[diagnostics]]
 document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
+message = "ww-fastq-to-cram.wdl:44:11: note[TrailingComma]: item missing trailing comma"
+permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/2d1e1989a57402642c06d15f4b623ac66fd9ed7d/ww-fastq-to-cram.wdl/#L44"
+
+[[diagnostics]]
+document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
 message = "ww-fastq-to-cram.wdl:46:7: note[CommentWhitespace]: in-line comments should be preceded by two spaces"
 permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/2d1e1989a57402642c06d15f4b623ac66fd9ed7d/ww-fastq-to-cram.wdl/#L46"
 
@@ -182,6 +187,11 @@ permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/2d1e1989a57402642
 
 [[diagnostics]]
 document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
+message = "ww-fastq-to-cram.wdl:53:9: note[TrailingComma]: item missing trailing comma"
+permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/2d1e1989a57402642c06d15f4b623ac66fd9ed7d/ww-fastq-to-cram.wdl/#L53"
+
+[[diagnostics]]
+document = "getwilds/ww-fastq-to-cram:/ww-fastq-to-cram.wdl"
 message = "ww-fastq-to-cram.wdl:56:25: note[CommentWhitespace]: in-line comments should be preceded by two spaces"
 permalink = "https://github.com/getwilds/ww-fastq-to-cram/blob/2d1e1989a57402642c06d15f4b623ac66fd9ed7d/ww-fastq-to-cram.wdl/#L56"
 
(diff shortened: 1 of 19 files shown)


wdl-lint/src/rules/trailing_comma.rs
```diff
@@ -0,0 +1,132 @@
...
+            let last_child = item.items().last();
+            if let Some(last_child) = last_child {
+                if let Some(last_child_comma) = last_child.syntax().next_sibling_or_token() {
+                    if last_child_comma.kind() != wdl_ast::SyntaxKind::Comma {
+                        state.add(missing_trailing_comma(
+                            last_child.syntax().text_range().to_span(),
+                        ));
+                    }
+                } else {
+                    // no next means no comma
```
claymcleod, line 94: Won't there always at least be a closing brace? So I don't know that this will ever execute, right? You can probably `unwrap()` the result of `next_sibling_or_token()` with a `// SAFETY` comment.

wdl-lint/src/rules/trailing_comma.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        // Check if object is multi-line
+        if item.syntax().to_string().contains('\n') && item.items().count() > 1 {
+            let last_child = item.items().last();
+            if let Some(last_child) = last_child {
+                if let Some(last_child_comma) = last_child.syntax().next_sibling_or_token() {
```
claymcleod, line 87: How do you want to handle something like this?

```
>note[TrailingComma]: item missing trailing comma
>   ┌─ tests/lints/trailing-comma/source.wdl:55:13
>   │
>55 │             baz: "quux" ,
>   │             ^^^^^^^^^^^
>   │
>   = fix: add a comma after this element
>
```

In this case, there is a whitespace before the comma—technically it ends with a comma, but the code indicates that it doesn't. You're probably better off consuming the trivia tokens with their own error (extraneous whitespace or similar) until you get to a non-trivia node.

claymcleod: I'd add a test for it too.

wdl-lint/tests/lints/trailing-comma/source.wdl
```diff
@@ -0,0 +1,47 @@
+#@ except: DescriptionMissing, MatchingParameterMeta
+
+version 1.2
+
+task foo {
+    meta {
+        description: {
+            help: "test"
```
claymcleod, line 8: Maybe add comments saying what is okay and what isn't? It was clear to me from the code alone.

claymcleod (changes requested):
Here's my initial thoughts.

Arena.toml
a-frantz, line 1: Have you reviewed each of these? There's a huge number of new ones which makes me suspicious. Are these all true positives?

adthrasher: I'll double-check, but there should be a ton. See the number of changes in: https://github.com/stjudecloud/workflows/pull/175/files

a-frantz: Good point. Let me review that quickly, merge, and then we can check whatever remains instead of going through the current Arena.

wdl-lint/src/rules/trailing_comma.rs
```diff
@@ -0,0 +1,169 @@
...
+impl Rule for TrailingCommaRule {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Ensures that lists and objects have a trailing comma."
+    }
+
+    fn explanation(&self) -> &'static str {
```
a-frantz, line 50: 1. I don't think this should mention the `key_value_pairs` rule (especially not with that casing). This is another case for that "linked rule" idea I proposed somewhere else, but IMO doesn't belong in the `explanation()` text.
2. There's another case not mentioned in this text or your code, but IMO should be covered here. Any `CallStatement` with more than one input should have trailing commas (and also be kept to one input-per-line).

wdl-lint/tests/lints/trailing-comma/source.wdl
```diff
@@ -0,0 +1,55 @@
...
+        description: {
+            help: "test"  # OK
+        }
+        help: {
+            name: "something",
+            other: "another"  # missing comma
+        }
+        foo: {
+            bar: "baz",
+            baz: "quux" ,  # misplaced comma
```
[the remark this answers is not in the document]
adthrasher: I added that case and it works as expected. The `find_next_comma` function skips anything tokens when looking for the comma. So any amount of trivia should be fine.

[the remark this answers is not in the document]
a-frantz: Nice! LGTM

Arena.toml
adthrasher: Based on our Slack conversation, these are all real issues. They're fixed in the `workflows` PR (https://github.com/stjudecloud/workflows/pull/175), but that needs resolution of a `miniwdl` issue (https://github.com/chanzuckerberg/miniwdl/issues/708) to be resolved to handle trailing commas in `array` types in metadata sections.

wdl-lint/src/rules/trailing_comma.rs
adthrasher: I removed #1. For #2, I think that belongs in the `CallInputSpacing` rule, personally.

adthrasher: Ehh, maybe not for (2), since that will likely need the `find_next_comma` function.

a-frantz: It's the exact same logic applied here, isn't it? More-or-less? That suggests to me the enforcement falls here as well.

A rule with `Spacing` in the name shouldn't enforce comma presence IMO.

adthrasher: @a-frantz - OK, I just added the CallStatement check. The spacing is enforced in `CallInputSpacing`. The trailing comma is enforced here. It actually checks each `input` for a trailing comma, even though missing commas before the last element would be a syntax error.

a-frantz (approved):

PR #138, opened 2024-07-19
feat: adds diving facilities for tree traversal

claymcleod:
This pull request introduces facilities for "diving" into the syntax tree.

@peterhuene and I had a discussion this morning regarding how some rules were not necessarily amenable to being written using the visitor pattern—in my mind, (at least) the following cases can fall into this category:

- very simple rules for which keeping track of state through the visitor pattern is arduous, and
- rules that require traversing the tree multiple times.

In these cases, facilities for simply scanning the tree for nodes of interest in, say, the `document()` callback is desired.

Here, I've introduced a `DiveIterator` and corresponding `dive()` and `dive_with_ignore()` methods. You can review the docs for each to see their independent purposes.

      changes (when appropriate).

### DIFF
--- a/wdl-grammar/src/lib.rs
+++ b/wdl-grammar/src/lib.rs
@@ -35,7 +35,7 @@ mod diagnostic;
 pub mod grammar;
 pub mod lexer;
 pub mod parser;
-pub mod tree;
+mod tree;
 pub mod version;
 
 pub use diagnostic::*;
--- a/wdl-grammar/src/lib.rs
+++ b/wdl-grammar/src/lib.rs
@@ -35,7 +35,7 @@ mod diagnostic;
 pub mod grammar;
 pub mod lexer;
 pub mod parser;
-mod tree;
+pub mod tree;
 pub mod version;
 
 pub use diagnostic::*;
--- a/wdl-grammar/src/tree.rs
+++ b/wdl-grammar/src/tree.rs
@@ -2,10 +2,11 @@
 
 pub mod dive;
 
+use std::borrow::Cow;
 use std::fmt;
 
-use rowan::GreenNode;
 use rowan::GreenNodeBuilder;
+use rowan::GreenNodeData;
 
 use super::grammar;
 use super::lexer::Lexer;
@@ -483,10 +484,8 @@ impl SyntaxTree {
     }
 
     /// Gets a copy of the underlying root green node for the tree.
-    pub fn green(&self) -> GreenNode {
-        // NOTE: this does clone the underlying green node, but that is an RC
-        // underneath the hood, so it's a very cheap clone.
-        self.0.green().into_owned()
+    pub fn green(&self) -> Cow<'_, GreenNodeData> {
+        self.0.green()
     }
 
     /// Converts the tree into a syntax node.
--- a/wdl-grammar/src/tree.rs
+++ b/wdl-grammar/src/tree.rs
@@ -1,7 +1,10 @@
 //! Module for the concrete syntax tree (CST) representation.
 
+pub mod dive;
+
 use std::fmt;
 
+use rowan::GreenNode;
 use rowan::GreenNodeBuilder;
 
 use super::grammar;
@@ -477,6 +480,13 @@ impl SyntaxTree {
     /// Gets the root syntax node of the tree.
     pub fn root(&self) -> &SyntaxNode {
         &self.0
+    }
+
+    /// Gets a copy of the underlying root green node for the tree.
+    pub fn green(&self) -> GreenNode {
+        // NOTE: this does clone the underlying green node, but that is an RC
+        // underneath the hood, so it's a very cheap clone.
+        self.0.green().into_owned()
     }
 
     /// Converts the tree into a syntax node.
--- a/wdl-grammar/src/tree/dive.rs
+++ b/wdl-grammar/src/tree/dive.rs
@@ -1,98 +1,76 @@
 //! Utilities for traversing a syntax tree while collecting elements of interest
 //! (i.e., "diving" for elements).
 
-use std::collections::VecDeque;
 use std::iter::FusedIterator;
-use std::marker::PhantomData;
 
+use rowan::api::PreorderWithTokens;
 use rowan::Language;
 
 use crate::SyntaxElement;
 use crate::SyntaxNode;
 use crate::WorkflowDescriptionLanguage;
 
-/// An iterator that explores every node within a
-/// [`SyntaxNode`](rowan::SyntaxNode) yielding all of the elements match the
-/// conditions laid out in `match_predicate`.
-///
-/// There are also facilities to ignore certain subtrees within the tree via the
-/// `ignore_predicate` callback. If the entire tree is desired, one can simply
-/// have the `ignore_predicate` always return `false`.
-#[derive(Debug)]
-pub struct DiveIterator<L, M, I>
+/// An iterator that performs a pre-order traversal of a
+/// [`SyntaxNode`](rowan::SyntaxNode)'s descendants and yields all elements
+/// while ignoring undesirable subtrees.
+#[allow(missing_debug_implementations)]
+pub struct DiveIterator<L, I>
 where
     L: Language,
-    M: Fn(&rowan::SyntaxElement<L>) -> bool,
     I: Fn(&rowan::SyntaxNode<L>) -> bool,
 {
-    /// The queue of nodes to be visited.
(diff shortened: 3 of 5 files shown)


wdl-grammar/src/lib.rs
```diff
@@ -35,7 +35,7 @@ mod diagnostic;
 pub mod grammar;
 pub mod lexer;
 pub mod parser;
-mod tree;
+pub mod tree;
```
claymcleod, line 38: cc: @peterhuene, just making sure you're okay with this

wdl-grammar/src/tree/dive.rs
```diff
@@ -0,0 +1,265 @@
...
+    #[test]
+    fn it_dives_correctly() {
+        let tree = get_syntax_node();
+
+        let mut ignored_idents = tree.dive(|element| element.kind() == SyntaxKind::Ident);
+
+        // NOTE: unlike the visitor, which carries out a preorder traversal, the
+        // current implementation of dive in this create uses a level-order
+        // traversal. Thus, the names of the workflow and task will be seen
+        // before their inner members.
```
adthrasher, line 219: Do you mean the names of **all** workflow and tasks will be seen before their members?

wdl-grammar/src/tree/dive.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    /// Visits every element in a syntax tree and accumulates all
+    /// [`SyntaxElement`](rowan::SyntaxElement)s for which the `match_predicate`
+    /// evaluates to `true`. No children nodes will be searched for
+    /// [`SyntaxNode`](rowan::SyntaxNode)s that are visited and for which the
+    /// `ignore_predicate` evaluates to `true`. In this way, you can ignore
+    /// entire sections of the tree.
+    ///
+    /// No guarantees are made about the order in which elements will be
+    /// traversed or collected.
```
adthrasher, line 126: Can you clarify what you mean here? You're doing a breadth-first traversal. Do you just mean that the `ignore` component breaks the order guarantee? It should still be breadth-first-ish right?

wdl-grammar/src/tree/dive.rs
claymcleod: In this case, yes. The idea is that a breadth-first search is carried out in this iterator, so the `workflow` and `task` nodes will be visited before their children.

wdl-grammar/src/tree/dive.rs
claymcleod: No, it's not that the `ignore` component breaks the guarantee but rather that, by virtue of the way the elements are accumulated and stored, the traversal is different. I would consider this to be fully breadth-first, as every element is inspect in a BFS order (though not every element is returned).

wdl-grammar/src/tree/dive.rs
```diff
@@ -0,0 +1,265 @@
...
+impl<L, M, I> Iterator for DiveIterator<L, M, I>
+where
+    L: Language,
+    M: Fn(&rowan::SyntaxElement<L>) -> bool,
+    I: Fn(&rowan::SyntaxNode<L>) -> bool,
+{
+    type Item = rowan::SyntaxElement<L>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        while let Some(element) = self.nodes.pop_front() {
```
claymcleod, line 71: @adthrasher this is the key: internally, the iterator keeps track of a `VecDeque` with the nodes it needs to visit. Before any particular node is served up by the iterator, all of that node's children are added to the back of the queue. This design is what culminates in the BFS.

wdl-grammar/src/lib.rs
peterhuene: For my understanding, why was this change required? We're already publicly using everything in the tree module at the top-level.

wdl-grammar/src/tree.rs
```diff
@@ -479,6 +482,13 @@ impl SyntaxTree {
         &self.0
     }
 
+    /// Gets a copy of the underlying root green node for the tree.
+    pub fn green(&self) -> GreenNode {
```
peterhuene, line 486: Minor nit: generally I've been returning a reference to the green node and letting the caller decide if it needs to be cloned or not.

wdl-grammar/src/tree/dive.rs
```diff
@@ -0,0 +1,259 @@
...
+
+    /// Visits every element in a syntax tree and accumulates all
+    /// [`SyntaxElement`](rowan::SyntaxElement)s for which the `match_predicate`
+    /// evaluates to `true`. No children nodes will be searched for
+    /// [`SyntaxNode`](rowan::SyntaxNode)s that are visited and for which the
+    /// `ignore_predicate` evaluates to `true`. In this way, you can ignore
+    /// entire sections of the tree.
+    ///
+    /// No guarantees are made about the order in which elements will be
+    /// traversed or collected.
```
peterhuene, line 126: Similar to the above:

```suggestion
    /// Iterates over every element in the tree at the current root and yields the
    /// elements for which the given `match_predicate` evaluates to `true`.
    /// 
    /// If the `ignore_predicate` evaluates to `true`, the subtree at the given node
    /// will not be traversed.
    ///
    /// No guarantees are made about the order in which elements will be traversed.
```

wdl-grammar/src/tree/dive.rs
```diff
@@ -0,0 +1,259 @@
...
+                rowan::SyntaxElement::Node(node) => {
+                    if !(self.ignore_predicate)(node) {
+                        self.nodes.extend(node.children_with_tokens());
+                    }
+                }
+                // NOTE: tokens have no children to explore.
+                rowan::SyntaxElement::Token(_) => {}
+            }
+
+            if (self.match_predicate)(&element) {
```
peterhuene, line 82: At first glance this feels like a job for `Iterator::filter`.

I.e. this iterator always yields every element in the BFS (excluding subtrees that are ignored) and one can adapt the iterator with `filter` to exclude the element from the result.

Related: is the `clone` necessary? Seems like `element` is owned here and could just be returned, but that's just running the source through my brain's borrow-checker.

wdl-grammar/src/tree/dive.rs
```diff
@@ -0,0 +1,259 @@
...
+pub trait Divable<L>
+where
+    L: Language,
+{
+    /// Visits every element in a syntax tree and accumulates all
+    /// [`SyntaxElement`](rowan::SyntaxElement)s for which the `match_predicate`
+    /// evaluates to `true`.
+    ///
+    /// No guarantees are made about the order in which elements will be
+    /// traversed or collected.
```
peterhuene, line 110: As there's no accumulation/collection taking place, I think we could simply say the following:

```suggestion
    /// Iterates over every element in the tree at the current root and yields the
    /// elements for which the given `match_predicate` evaluates to `true`.
    ///
    /// No guarantees are made about the order in which elements will be traversed.
```

peterhuene (commented):
This looks great!

Just a comment or two with some nits.

wdl-grammar/src/tree/dive.rs
peterhuene: I do wonder if these methods should just return `DiveIterator` so we can make a guarantee about the ordering rather than `impl Iterator` (perhaps even renaming `DiveIterator` to `BreadthFirstIterator` to make it even more explicit).

I'm sure you have a reason, though, for wanting the ordering to be unspecified in the `Divable` trait.

wdl-grammar/src/lib.rs
claymcleod: Hm, I'm not sure looking in hindsight. I suppose that was something I changed and forgot to remove this once it was unneeded.

wdl-grammar/src/tree.rs
claymcleod: Fair enough. I introduced it as a convenience method for exactly that purpose but, to your point, the method is named more generally than that. I don't love any of the other method names, so I suppose I'll split the difference and require the caller to call `.into()`.

wdl-grammar/src/tree/dive.rs
claymcleod: Okay, I've addressed both of these.

wdl-grammar/src/tree/dive.rs
claymcleod: @peterhuene and I discussed offline—we decided to wrap the provided `SyntaxNode::preorder_with_tokens` iterator.

a-frantz (approved):

PR #142, opened 2024-07-25
feat: implements the `container_value` rule

claymcleod:

- **Rule Name**: `ContainerValue`

This rule is intended to implement the `mutable_container` and `immutable_container_not_tagged` baseline lint rules ([link](https://github.com/stjude-rust-labs/wdl/discussions/11)). In so doing, it was an exploration in implementing various wrappers within the AST that were built from existing AST nodes (rather than cast from `SyntaxNode`s). I also added in a few checks related to cointainers that seemed logical to me, namely:

- `container` arrays with zero elements are ambiguous and should be replaced with something concrete.
- `container` arrays with one element should be replaced with a single, literal string.
- `container` arrays with the special 'any' container designator (`*`) are ambiguous, as you're both indicating that a set of specific containers should be used _and_ that any container can be used. As such, it should be avoided.

Additionally, I fixed a small bug ("oversight"?) I discovered in the `MissingRuntime` lint while developing this: `MissingRuntime` should only be flagging things for WDL v1.1 or earlier, as the `runtime` section was deprecated in favor of `requirements` in WDL v1.2.

Rule specific checks:

      is implemented.
      (`Arena.toml`).


wdl-lint/tests/lints.rs
```diff
@@ -154,10 +154,10 @@ fn main() {
 
     let ntests = AtomicUsize::new(0);
 
-    fn inner<'a>(test: &'a PathBuf, ntests: &AtomicUsize) -> Option<(&'a str, String)> {
+    fn inner<'a>(test: &'a Path, ntests: &AtomicUsize) -> Option<(&'a str, String)> {
```
claymcleod, line 157: This is a stray clippy lint.

wdl-lint/src/rules/comment_whitespace.rs
```diff
@@ -306,7 +306,7 @@ task foo {
         let ancestors = comment
             .syntax()
             .parent_ancestors()
-            .filter(|a| super::filter_parent_ancestors(a))
+            .filter(super::filter_parent_ancestors)
```
claymcleod, line 309: These are stray clippy lints.

wdl-lint/src/rules/deprecated_object.rs
```diff
@@ -40,12 +40,12 @@ impl Rule for DeprecatedObjectRule {
 
     fn explanation(&self) -> &'static str {
         "WDL `Object` types are officially deprecated and will be removed in the next major WDL release.
-        
+
```
claymcleod, line 43: `cargo +nightly fmt`

wdl-lint/RULES.md
```diff
@@ -6,35 +6,35 @@ be out of sync with released packages.
...
-| `Whitespace`                     | Spacing, Style                | Ensures that a document does not contain undesired whitespace.                        |
-
+| Name                             | Tags                          | Description                                                                                       |
+| :------------------------------- | :---------------------------- | :------------------------------------------------------------------------------------------------ |
+| `CallInputSpacing`               | Style, Clarity, Spacing       | Ensures proper spacing for call inputs                                                            |
+| `CommandSectionMixedIndentation` | Clarity, Correctness, Spacing | Ensures that lines within a command do not mix spaces and tabs.                                   |
+| `CommentWhitespace`              | Spacing                       | Ensures that comments are properly spaced.                                                        |
+| `ContainerValue`                 | Clarity, Portability          | Ensures that the value for `container` keys in `runtime`/`requirements` sections are well-formed. |
```
claymcleod, line 14: The only thing that changed it this line—everything else is just a spacing change for the last column.

wdl-ast/examples/explore_requirements.rs
```diff
@@ -0,0 +1,39 @@
...
+        .ast()
+        // SAFETY: only V1 ASTs are supported for this example at present.
+        .unwrap_v1()
+        .tasks()
+        .next()
+        // SAFETY: we expect the document to contain at least one task for this example
+        // to work correctly.
+        .expect("document to contain at least one task");
+
+    let requirements = task.requirements().next().unwrap_or_else(|| {
```
claymcleod, line 29: @peterhuene, I did not end up making the changes we discussed for rolling the validation back to the CST here. One of us can do that in a future PR.

wdl-ast/src/v1/expr.rs
```diff
@@ -66,6 +66,28 @@ pub enum Expr {
 }
 
 impl Expr {
+    /// Attempts to reference a literal expression.
```
claymcleod, line 69: All of these were not strictly needed for this PR, but I did need one, so I added all of them to be consistent.

wdl-ast/src/v1/expr.rs
```diff
@@ -3698,7 +4507,7 @@ task test {
             r#"
 version 1.1
 
-task test { 
+task test {
```
claymcleod, line 4510: I guess clippy caught this?

wdl-ast/src/v1/task/common/container/value.rs
```diff
@@ -0,0 +1,407 @@
+//! Values for the `container` item within the `runtime` and `requirements`
```
claymcleod, line 1: This gets, admittedly, a bit more gnarly than I would have hoped. I iterated on quite a few designs on how to capture all of the information I wanted. To summarize,

- I landed on calling the entire right side of the `container` assignment as a `Value`. The `Value` can be `Kind::String` (a single, string literal) or `Kind::Array` (an array of string literals).
- Each individual string literal within the value is represented by a `Uri`. The `Uri` can be `Kind::Any` (representing the `*` container URI) or `Kind::Entry` (open to a better name here, but represents any defined container URI that is not `*`).

Both `Value` and `Uri` are dynamically typed because (a) they both have common functionality where they keep track of the backing `AstNode` regardless of their type and (b) it was easier to test the `Kind`s without needing to deal with the fact that both `Value` and `Uri` need backing `AstNode`s.

Arena.toml
adthrasher, line 1: @a-frantz - Are we still convinced that this is a good rule? Particularly, `container URI uses a mutable tag`. Unsurprisingly, it flags what appears to be every `container` specification in our test set.

wdl-lint/src/rules/container_value.rs
```diff
@@ -0,0 +1,199 @@
...
+        &mut self,
+        state: &mut Self::State,
+        reason: VisitReason,
+        section: &RuntimeSection,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        if let Some(container) = section.container() {
```
adthrasher, line 143: Should this lint warn if `container` is omitted?

wdl-lint/src/rules/missing_runtime.rs
```diff
@@ -75,9 +76,13 @@ impl Visitor for MissingRuntimeRule {
...
-            let name = task.name();
-            state.add(missing_runtime_section(name.as_str(), name.span()));
+        // This rule should only be present for WDL v1.1 or earlier, as the
+        // `requirements` section replaces it in WDL v1.2.
+        if let SupportedVersion::V1(minor_version) = self.0.expect("version should exist here") {
+            if minor_version <= V1::One && task.runtimes().next().is_none() {
+                let name = task.name();
+                state.add(missing_runtime_section(name.as_str(), name.span()));
+            }
         }
```
adthrasher, line 86: We need additional checks. One to warn that `runtime` is deprecated in 1.2+ documents. We also need a lint rule to warn on missing `requirements` sections in 1.2+ documents. Another check if `runtime` and `requirements` are specified in 1.2+ documents, though that probably ought to be a validation error if it isn't currently.

Some of that could be implemented here or in new rules, but we at least need to record that these are needed.

Arena.toml
a-frantz: Copying the text from #11:
> All tasks should run in an immutable container. This ensures reproducibility across time and environments. wdl-grammar and wdl-ast will look for a :SHASUM tag in your container declarations and warn if one is missing. A sha digest will always point to the exact same image, whereas tags are mutable. This mutability makes even versioned tags problematic when we want to ensure reproducibility.
>
> While the confidence in persistence gained by using sha digests for pulling container images is valuable, it comes at a cost: lack of human readability. So we enforce following container strings with a comment which gives a human readable name to the image being pulled. It is imperative that the sha digest and the human readable tag are kept in sync. It's extra overhead compared to just using a tag directly, but we consider it well worth it.

The above text is the result of a few rounds of back and forth, and I would argue is the best we could make of the current situation. The main drawback is acknowledged in the text, which is that SHASUMs are not human-readable.

I am tentatively voting for adopting the convention outlined above. I am not a big fan of it, maintaining both an SHASUM tag and a human-readable label comment is error-prone. But I think it's the best we can do with what we're given. I would love to hear alternative ideas.

wdl-ast/src/v1/task/common/container/value.rs
```diff
@@ -0,0 +1,407 @@
...
+                    return Ok(Value {
+                        kind: Kind::Array(uris),
+                        expr,
+                    });
+                }
+                _ => {}
+            }
+        }
+
+        Err(Error::InvalidExpression { expr })
```
a-frantz, line 294: Is this right? I believe that any expression that can be coerced into a `String` or `Array[String]` is valid. Our workflows repo is in the minority by using literal strings for the container. Most other* WDL repos use combinations of placeholders and runtime declarations. I think we can argue that's bad practice, but it should be parsed as valid.

*"most" as defined by my informal survey

wdl-lint/src/rules/container_value.rs
```diff
@@ -0,0 +1,199 @@
...
+/// sections are well-formed.
+#[derive(Default, Debug, Clone, Copy)]
+pub struct ContainerValue;
+
+/// Creates a missing tag diagnostic.
+fn missing_tag(span: Span) -> Diagnostic {
+    Diagnostic::warning(String::from("container URI is missing a tag"))
+        .with_rule(ID)
+        .with_highlight(span)
+        .with_fix("add a tag to the container URI (e.g., `ubuntu:latest` instead of `ubuntu`)")
```
a-frantz, line 35: Shouldn't the fix here use an SHASUM?

wdl-lint/src/rules/container_value.rs
a-frantz: I think "yes". Although I could see implementing that as a separate lint? It should be warned on, either here or elsewhere.

Arena.toml
claymcleod: There's a case to be made that we shouldn't be quite so invasive on some of the rules. I haven't decided how I feel about it yet. That goes for some of the naming rules as well (for example, I saw it being discussed about `id`, and I'm on the fence whether `wdl-lint` should have an opinion on that).

All that being said, I think we're in a place right now of implementing the rules we _think_ could be useful, but I think we are going to eventually have to come back and (a) add in configuration options for all of these rules and (b) also, potentially, turn some of the rules off by default and make them opt-in in some way (that was the idea behind the "pedantic" category of old).

wdl-ast/examples/explore_requirements.rs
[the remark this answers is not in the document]
claymcleod: Yeah that's a fair point. I ended up writing these as I was developing how the interface should work to get a particular `runtime`/`requirements` key. My thought was that these help show how it's intended to get access to the inner fields, and we could add other fields as they are supported (it's going to take some time to write all of that, as it took me quite a long time for just container).

I'd be curious whether others think it's better to not include versus to include a halfway finished example here.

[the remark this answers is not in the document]
peterhuene: Although I don't have a strong opinion relating to examples, I agree with @a-frantz that this particular example seems a little too specific.

Perhaps we could extend the existing "explore" example to print more?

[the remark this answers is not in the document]
adthrasher: I also agree with @a-frantz (and @peterhuene).

Arena.toml
adthrasher: I could certainly see adding a configuration method to turn off some of the checks in each rule. Using the input/output name rules, I could see someone agreeing with most, but not wanting to require 3+ characters. Right now, the whole rule is on or off.

adthrasher: > So we enforce following container strings with a comment which gives a human readable name to the image being pulled.
Right now, I don't think this rule enforces this.

a-frantz: > > So we enforce following container strings with a comment which gives a human readable name to the image being pulled.
> 
> Right now, I don't think this rule enforces this.

Correct, I think that should be added as a separate lint rule (in another PR).

peterhuene: I agree that this seems overly invasive if very few users specify their containers by SHA, despite it being a good practice.

I would be less worried about it we had configuration in place so that users may opt to globally disable it for an entire workspace if so desired.

This particular issue seems like a good feature for an execution engine: producing an "image label lock file" for your image label references in your WDLs and would detect label changes; one could then commit the lock file to your workflow repository.

a-frantz: > This particular issue seems like a good feature for an execution engine: producing an "image tag lock file" for your image tag references in your WDLs and would detect tag changes; one could then commit the lock file to your workflow repository, allowing users to continue to use human-friendly version tag in their container values.

I really like this idea. So the actual WDL file would have a human-readable tag that a separate lockfile associates with an SHASUM. I'm thinking the tag used in the WDL file doesn't even have to correspond to anything in the container repository, so long as the relationship between tag and SHASUM is codified by the lock file. This bypasses the issue of Docker tags being mutable.

Should we share this idea with the OpenWDL slack? I think it has potential.

wdl-ast/examples/explore_requirements.rs
claymcleod: This is now tracked in #149.

Arena.toml
claymcleod: Yeah that is totally the way to do it, great idea @peterhuene.

What does that mean for _this_ lint rule—that we should simply suggest adding a tag and not worry about the SHA256 aspect of it?

a-frantz: > What does that mean for _this_ lint rule—that we should simply suggest adding a tag and not worry about the SHA256 aspect of it?

I think it will probably be a decent amount of time before any "WDL Container Lock File" reaches maturity. So in the meantime, I'm still a (reluctant) advocate for using SHASUMs plus a human-readable label in an inlined comment.

a-frantz: > What does that mean for _this_ lint rule—that we should simply suggest adding a tag and not worry about the SHA256 aspect of it?

I think it will probably be a decent amount of time before any "WDL Container Lock File" reaches maturity. So in the meantime, I'm still a (reluctant) advocate for using SHASUMs plus a human-readable label in an inlined comment.

wdl-lint/src/rules/container_value.rs
claymcleod: I tried this out in this rule and didn't like the affect. I think that all of the logic about which keys should be included/not included should be in the same rule as where we check to make sure certain sections are present _at all_. That way, we're not both flagging sections as being deprecated (such as a `runtime` block in a WDL v1.2 document) while also flagging within the runtime block that the `container` key is missing.

claymcleod: This is tracked here: https://github.com/stjude-rust-labs/wdl/issues/153.

claymcleod (commented):
This should be ready for review again.

wdl-ast/src/v1/task/common/container/value.rs
claymcleod: You're right, but it's going to require type-checking to make that work. So I settled on this for now and think we should change it in the future. If you agree, I'll create an issue to track it.

wdl-lint/src/rules/missing_runtime.rs
claymcleod: I've added a new `MissingRequirements` rule to address this. This rule also flags `runtime` sections used when `requirements` should be. The validation takes care of both of them not being present today.

Arena.toml
claymcleod: Okay, that's good with me I suppose. We'll add configurations such that the rule can be disabled soon anyway.

wdl-ast/src/v1/task/common/container/value.rs
a-frantz: I guess I'm a bit confused about all the plumbing between the pieces here. The class of errors I'm concerned with are expressions with placeholders and name references. I don't see anything in Arena or Gauntlet related to these, but the code _here_ should be firing an error in those cases. Is this error just not exposed anywhere ATM?

There also should be test cases in this file for those styles of expressions, which is maybe contributing to my confusion here.

adthrasher (approved):
Looks good, pending @a-frantz's comments.

wdl-ast/src/v1/task/common/container/value.rs
claymcleod: Oh I see what you're concerned about. This does throw an error but, ultimately, that error gets ignored on line 150 and 167 of `wdl-lint/src/rules/container_value.rs` (because I just check if the result is `Ok()`—I don't do anything if it's an `Err`).

Put another way, the current `Value` for this field doesn't respect coercion rules, but that's because coercion is complicated and needs to wait for the bigger coercion PR to be able actually do the coercion.

I've put a test to demonstrate that the use case you mention above is simply ignored.

a-frantz: That's great @claymcleod . Answered my question and LGTM 👍 
Doing a quick review to make sure I didn't miss anything then hitting approve 🚀

a-frantz (approved):

PR #146, opened 2024-07-30
refactor: refactor the `Analyzer` API and remove cache from `wdl-lsp`.

peterhuene:
Main changes:

* Analysis results now have a unique identifier that changes when a file is reanalyzed.
* The analyzer can now be notified of incremental or full changes to a document to cause it to be reparsed and reanalyzed.
* Removed the `changed` callback that was previously passed to `Analyzer::new`; the analyzer now mutates its state when notified of changes instead.
* Removed `Workspace` from `wdl-lsp` as its functionality as been integrated into `Analyzer`; it also removes an unnecessary cache layer.

Also included:

* Made the file tests for `wdl-analysis` run in parallel, except for the cycle detection test which is sensitive to the ordering in which parsing completes.
* Added a few unit tests for `Analyzer`.

      changes (when appropriate).


wdl-analysis/src/analyzer.rs
```diff
@@ -64,6 +70,11 @@ pub enum ParseResult {
     Error(Arc<anyhow::Error>),
     /// The document was parsed.
     Parsed {
+        /// The version of the parsed document.
```
claymcleod, line 73: I might clarify this is the monotonic version (just to delineate from the WDL version).

wdl-analysis/src/analyzer.rs
```diff
@@ -393,52 +423,79 @@ impl Analyzer {
...
+                    documents.push(path_to_uri(&path).with_context(|| {
+                        format!(
+                            "failed to convert path `{path}` to a URI",
+                            path = path.display()
+                        )
+                    })?);
                     continue;
                 }
 
+                if !path.is_dir() {
```
claymcleod, line 448: What about `PathBuf::is_symlink()`?

wdl-analysis/src/analyzer.rs
```diff
@@ -393,52 +423,79 @@ impl Analyzer {
...
-                        log::error!("error while searching for WDL documents: {e}");
+                let paths = glob::glob_with(&pattern, options).with_context(|| {
+                    format!("failed to glob directory `{path}`", path = path.display())
+                })?;
+                for file in paths {
+                    let file = file.with_context(|| {
+                        format!("failed to read directory `{path}`", path = path.display())
+                    })?;
+
+                    if !file.is_file() {
```
claymcleod, line 467: Should we follow symlinks here? Or at least warn that we aren't.

peterhuene: `PathBuf::is_file` does traverse symlinks.

Still, I'm going to rewrite this to use `walkdir` and not `glob` as the latter doesn't do symlink cycle detection.

wdl-analysis/src/analyzer.rs
peterhuene: Both `is_file` and `is_dir` traverse symlinks so the only case in which both return false is an error to retrieve the filesystem metadata.

I'll also rewrite this to be slightly more efficient as only a single metadata call is needed.

wdl-lsp/src/proto.rs
```diff
@@ -84,3 +100,140 @@ pub fn diagnostic(
...
+            if previous == result.id().as_ref() {
+                log::debug!(
+                    "diagnostics for document `{uri}` have not changed (client has latest)",
+                    uri = result.uri(),
+                );
+
+                items.push(WorkspaceDocumentDiagnosticReport::Unchanged(
+                    WorkspaceUnchangedDocumentDiagnosticReport {
+                        uri: result.uri().as_ref().clone(),
+                        // TODO: FIX ME
```
claymcleod, line 193: Should this be addressed before we merge?

peterhuene: Whoops, I fixed it in one place but not the other. Fixing now.

wdl-lsp/src/server.rs
```diff
@@ -0,0 +1,575 @@
...
+            self.analyzer.notify_incremental_change(
+                params.text_document.uri,
+                IncrementalChange {
+                    version: params.text_document.version,
+                    start: Some(params.text_document.text),
+                    edits: Vec::new(),
+                },
+            );
+
+            // TODO: do something with version
```
claymcleod, line 404: And this?

wdl-analysis/src/analyzer.rs
claymcleod: I see you resolved this, but I didn't see any update to the comment. What's your
strategy on that @peterhuene?

peterhuene: I accidentally pushed up a clarification on the internal representation of this rather than the public API.

I'll fix!

wdl-analysis/src/analyzer.rs
```diff
@@ -242,7 +286,7 @@ impl SourceEdit {
         self.range.start..self.range.end
     }
 
-    /// Applies the edit to the given string.
+    /// Applies the edit to the given string if it's in range.
     pub(crate) fn apply(&self, source: &mut String, lines: &LineIndex) {
```
claymcleod, line 290: Attempting to apply an edit that isn't in range appears like something that should
return a `Result`. When that happens, is that typically going to be an issue on the
_client_ side of things (assuming no bugs in our own code, which we'd clearly want to flag)?

wdl-analysis/src/analyzer.rs
```diff
@@ -285,28 +329,30 @@ impl SourceEdit {
                 .offset(end)
                 .expect("expected a valid end position")
                 .into();
-        source.replace_range(range, &self.text);
+
+        if source.is_char_boundary(range.start) && source.is_char_boundary(range.end) {
+            source.replace_range(range, &self.text);
+        }
```
claymcleod, line 335: If we decide not to return a `Result` here, then failure to apply should at least be
logged somewhere so it can eventually be uncovered.

wdl-analysis/src/analyzer.rs
```diff
@@ -320,26 +366,31 @@ pub enum DocumentChange {
...
 /// Note that dropping the analyzer is a blocking operation as it will wait for
 /// the queue thread to join.
+///
+/// The type parameter is the context type passed to the progress callback.
 #[derive(Debug)]
-pub struct Analyzer {
+pub struct Analyzer<Context> {
     /// The sender for sending analysis requests to the queue.
-    sender: ManuallyDrop<mpsc::UnboundedSender<Request>>,
+    sender: ManuallyDrop<mpsc::UnboundedSender<Request<Context>>>,
```
claymcleod, line 374: Perhaps it is simply that this is confined to the `wdl` crates, but the purpose of
introducing `Context` here is not clear to me. Could you expound briefly on why you
introduced that?

wdl-analysis/src/analyzer.rs
```diff
@@ -393,52 +425,73 @@ impl Analyzer {
...
+        }
+
+        // Send the add request to the queue
+        let (tx, rx) = oneshot::channel();
+        self.sender
+            .send(Request::Add(AddRequest {
+                documents,
+                completed: tx,
+            }))
+            .expect("failed to send remove request");
```
claymcleod, line 491:
```suggestion
            .expect("failed to send add request");
```

wdl-analysis/src/analyzer.rs
```diff
@@ -393,52 +425,73 @@ impl Analyzer {
...
+
+        rx.await.unwrap_or_default();
```
claymcleod, line 493: Should we do this instead? It seems like we'd want to know about a `RecvError`.

```suggestion
        rx.await?;
```

wdl-analysis/src/analyzer.rs
```diff
@@ -447,37 +500,83 @@ impl Analyzer {
     /// the analyzer, those documents will be removed.
     ///
     /// Documents are only removed when not referenced from importing documents.
-    pub async fn remove_documents<'a>(&self, uris: Vec<Url>) {
+    pub async fn remove_documents<'a>(&self, documents: Vec<Url>) {
```
claymcleod, line 503: This should probably return a `Result<()>` same as adding documents.

wdl-analysis/src/analyzer.rs
claymcleod: There is also the alternative strategy that we should `.expect()` this if we expect that
it's always going to eventually resolve. In that case, that strategy applies for all of
the `unwrap_or_defaults()` I flagged here.

wdl-analysis/src/analyzer.rs
```diff
@@ -447,37 +500,83 @@ impl Analyzer {
...
-        context: Option<String>,
-    ) -> Vec<AnalysisResult> {
-        if documents.is_empty() {
-            return Default::default();
-        }
+    /// If the document is up-to-date and was previously analyzed, the current
+    /// analysis result is returned.
+    ///
+    /// Returns an analysis result for each document that was analyzed.
+    pub async fn analyze_document(&self, context: Context, document: Url) -> Vec<AnalysisResult> {
```
claymcleod, line 552: Return a `Result`?

wdl-analysis/src/analyzer.rs
```diff
@@ -501,5 +600,256 @@ impl Drop for Analyzer {
...
+        assert_eq!(results[0].diagnostics().len(), 1);
+        assert_eq!(results[0].diagnostics()[0].rule(), None);
+        assert_eq!(results[0].diagnostics()[0].severity(), Severity::Error);
+        assert_eq!(
+            results[0].diagnostics()[0].message(),
+            "conflicting workflow name `test`"
+        );
+
+        // Analyze again and ensure the analysis result id is unchanged
+        let id = results[0].id().clone();
```
claymcleod, line 723: Not required IMO, but it could be interesting to rewrite the file without the
conflicting name here and check that the diagnostic resolves itself (sort of an analog
to what you do with the incremental test).

wdl-analysis/src/graph.rs
```diff
@@ -105,6 +122,8 @@ impl Analysis {
 pub struct DocumentGraphNode {
     /// The URI of the document.
     uri: Arc<Url>,
+    /// The current incremental change to the document.
+    change: Option<IncrementalChange>,
     /// The parse state of the document.
     parse_state: ParseState,
     /// The analysis of the document.
```
claymcleod, line 129:
```suggestion
    /// The analysis of the document.
    ///
    /// If `None`, an analysis does not exist for the current state of the node.
```

claymcleod: This is not related to the PR changes, but I caught it when I suggested the above change.

wdl-analysis/src/analyzer.rs
peterhuene: This was originally `expect`, but I was running into an issue when the future was dropped (in the case of the LSP, the client canceled the request).

Let me investigate before doing so.

wdl-analysis/src/queue.rs
```diff
@@ -21,146 +21,238 @@ use tokio::sync::oneshot;
...
+                        "received request to add {count} document(s) to the graph",
                         count = documents.len()
                     );
 
-                    self.analyze(documents, context, completed);
+                    self.add_documents(documents);
 
                     log::info!(
-                        "analysis request completed in {elapsed:?}",
+                        "request to add documents completed in {elapsed:?}",
```
claymcleod, line 149: Considering that we're getting to a level of maturity where we want this kind of profiling, I wonder if it makes sense to switch to leveraging `tracing` in these instances. Shouldn't hold up the PR, but we may thank ourselves later rather than having to go back and add it in.

wdl-analysis/src/analyzer.rs
peterhuene: The edit comes from a notification in the LSP, so there's no response we send the client to let them know the range data they gave us was invalid.

The best we could do is fail a subsequent diagnostics request, which is when we're applying the edits anyway.

wdl-analysis/src/queue.rs
```diff
@@ -21,146 +21,238 @@ use tokio::sync::oneshot;
...
     fn analyze(
         &self,
-        documents: Vec<Arc<Url>>,
-        context: Option<String>,
-        completed: oneshot::Sender<Vec<AnalysisResult>>,
-    ) {
+        document: Option<Url>,
+        context: Context,
+        completed: &oneshot::Sender<Vec<AnalysisResult>>,
+    ) -> Vec<AnalysisResult> {
```
claymcleod, line 233: Nit: since this is consumer facing, it should probably be an enum that tells the caller
what happened (`T::Cancelled`, `T::Completed(Vec<AnalysisResult)`, etc).

wdl-analysis/src/analyzer.rs
peterhuene: I felt the `Option<String>` context argument that was used before was too specific to our LSP implementation, so instead changed it to a type parameter so that anyone using `wdl-analysis` can pass whatever they want to their progress callbacks (like in the case of the CLI, we can pass the progress bar in now, which I'll change to do).

wdl-analysis/src/analyzer.rs
peterhuene: What are you thinking would be the error case? It's quite intentional that we ignore documents that weren't previously added as this gets called from a file change notification in the LSP, which may be providing us directories that don't contain any WDL documents.

wdl-analysis/src/queue.rs
peterhuene: I think that would be a great follow-up PR, definitely put in sooner than later.

wdl-analysis/src/analyzer.rs
claymcleod: Okay I see now. Nevermind.

claymcleod (changes requested):
Great work, here's some thoughts.

wdl-analysis/src/analyzer.rs
claymcleod: I think this is wrapped up in the last question of "should this return an error if the queue fails to acknowledge the removal, thus indicating some issue with the queue".

wdl-analysis/src/analyzer.rs
claymcleod: That's what I was afraid of.

In that case, since it would really be the client and not whatever code wrapped our analyzer (`wdl-lsp` or otherwise), I think we should simply emit a warning saying that the edit will be ignored because it was out of range. That way, the indication will be clear to anyone who wants to investigate.

wdl-analysis/src/analyzer.rs
peterhuene: Added and this caught a bug!

wdl-analysis/src/analyzer.rs
peterhuene: I decided to thread `Result` back through on the analysis request, so if the client send us an invalid edit, requests the analyze the file will fail.

wdl-analysis/src/analyzer.rs
peterhuene: We now return this as an error rather than an empty result set.

wdl-analysis/src/analyzer.rs
peterhuene: The `Analyzer` methods now all return `Result`.

claymcleod (approved):
Looks great. Thanks for addressing those.

PR #159, opened 2024-08-13
fix: validate placeholder options.

### LINKED ISSUE #154: Ensure only one expression placeholder option is present
The AST currently has an `options()` method (https://github.com/stjude-rust-labs/wdl/blob/main/wdl-ast/src/v1/expr.rs#L1168) that will return multiple expression placeholder options. This is not allowed per the SPEC: https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#-expression-placeholder-options

> There are three options available. An expression placeholder may have at most one option.

peterhuene:
This commit adds a check to the validation rules for ensuring there is at most one placeholder option in a placeholder.

Consequently, the `options` method on `Placeholder` has been renamed to `option` and now returns `Option<PlaceholderOption>`.

Fixes #154.

      changes (when appropriate).

### DIFF
--- a/Gauntlet.toml
+++ b/Gauntlet.toml
@@ -377,14 +377,104 @@ message = 'umi.wdl:39:60: error: unknown escape sequence `\.`'
 permalink = "https://github.com/biowdl/tasks/blob/2bf875300d90a3c9c8d670b3d99026452d5dbae2/umi.wdl/#L39"
 
 [[diagnostics]]
+document = "broadinstitute/palantir-workflows:/BenchmarkSVs/BenchmarkSVs.wdl"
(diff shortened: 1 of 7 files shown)


Gauntlet.toml
a-frantz, line 1: So looks like cromwell allows `default` and `sep` to be together? That's unexpected, although I guess the logic of how to interpret it is clear.

a-frantz (approved):

Gauntlet.toml
peterhuene: Indeed, the violations were unexpected, but since we treat these as deprecated anyway, I don't feel like we should be making any exceptions even if cromwell accepts it.

PR #163, opened 2024-08-28
feat: type checking in tasks.

peterhuene:
This commit implements full type checking in tasks.

* Detection of cycles in declarations.
* Type calculations of expressions.
* Type checking of string interpolations.

Workflow type checking is forthcoming.

The following have not been implemented yet, which is required for full 1.2 support:

* support `task` name (with "task" hidden type) in commands and output declarations.
* support `input`, `output`, and `hints` hidden types in literal expression evaluation.

Additionally, the following changes are included:

* Refactored `wdl-analysis` to make certain modules public rather than reexporting everything at the root (cleans up the docs a bit).
* Removed DFS space from `DocumentGraph` as it doesn't need to persist after analysis completes.
* Simplified `DocumentScope` to contain only whats needed to express names and their types.
* Removed the restriction on `Map` type representation to allow `Union` as the key type, which allows type checking to work on a literal `{}`, which has calculated type `Map[Union, Union]`.
* Fixed `basename` function to accept a `String` argument.
* Fixed `size` function to accept a `String` argument.
* Removed the `span_of` function in favor of an `AstNode` extension trait.
* Changed the `call_expr` grammar rule to only accept an identifier for the target and not an expression as functions cannot be values in WDL.
* Fixed the CLI tools to not print colors when output is not to a terminal.
* Fixed incorrect parsing of some expressions which required the interior to be non-empty (e.g. `foo[]` is not a valid index expression).

      changes (when appropriate).


peterhuene:
Opened #164 to track the missing 1.2 features.

a-frantz (commented):
2 Qs:
1. does merging this bring us up to parity with `miniwdl check`?
2. I don't see a test for a forward reference, but from my understanding of this code and the spec those should be parseable.
```
task foo {
    input {
        String arg = if bool then "--arg" else ""
        Boolean bool
    }
    command <<<>>>
}
```

wdl-analysis/src/task/v1.rs
```diff
@@ -0,0 +1 @@
+//! Task graph representation for a V1 AST.
```
claymcleod, line 1: Not sure about this as well.

wdl-analysis/src/scope.rs
```diff
@@ -221,28 +163,25 @@ pub struct Struct {
     /// This is either the name of a struct definition (local) or an import's
     /// URI or alias (imported).
     span: Span,
+    /// The offset of the CST node from the start of the document.
+    ///
+    /// This is used to adjust
```
claymcleod, line 168: Used to adjust?

wdl-analysis/src/stdlib.rs
```diff
@@ -1581,6 +1610,16 @@ pub static STDLIB: LazyLock<StandardLibrary> = LazyLock::new(|| {
                             .parameter(PrimitiveTypeKind::String)
                             .ret(PrimitiveTypeKind::String)
                             .build(),
+                        // This overload isn't explicitly specified in the spec, but the spec
```
claymcleod, line 1613: Should we propose updates to the spec for these? Or no?

wdl-analysis/tests/analysis/conflicting-decl-names/source.diagnostics
```diff
@@ -5,7 +5,7 @@ error: conflicting output name `x`
    │             - the input with the conflicting name is here
    ·
 13 │         String x = "x"
-   │                ^ this conflicts with a input of the same name
+   │                ^ this output conflicts with a previously used name
```
claymcleod, line 8: this output name?

wdl-analysis/tests/analysis/name-reference-cycle/source.diagnostics
```diff
@@ -0,0 +1,9 @@
+error: name reference cycle detected
+  ┌─ tests/analysis/name-reference-cycle/source.wdl:6:9
+  │
+6 │     Int a = b
+  │         ^ this name participates in the cycle
+7 │     Int b = c
```
claymcleod, line 6: Should this be flagged as participating in the cycle?

claymcleod (approved):
When those are addressed, looks great! Admittedly, there is a ton here, so I may not have caught everything—I sort of ran with the assumption that the approach you took is solid and we can continue to tweak it in the future if need be (mainly because I don't have any concrete experience with better ways to do it).

wdl-analysis/src/scope/v1.rs
```diff
@@ -175,704 +201,1013 @@ fn call_conflict(name: &Ident, first: NameContext, suggest_fix: bool) -> Diagnos
...
+            DocumentItem::Import(_) | DocumentItem::Struct(_) => continue,
+            DocumentItem::Task(definition) => {
+                if let Some(task) = document.tasks.get(definition.name().as_str()) {
+                    type_check_task(document, &definition, task.scope, diagnostics);
                 }
             }
-            None => {
-                return Err(Some(import_missing_version(span)));
+            DocumentItem::Workflow(_) => {
+                // TODO: implement
```
claymcleod, line 924: Just a note that this is still TODO as well.

peterhuene:
@a-frantz:

> does merging this bring us up to parity with miniwdl check?

It does not yet; once I complete workflow type checking, it will.

> I don't see a test for a forward reference, but from my understanding of this code and the spec those should be parseable.

I added tests for cycle detection, but you're right, it'd be good to have one that passes without errors with a forward reference.  I'll get that added!

wdl-analysis/src/lib.rs
```diff
@@ -10,14 +10,13 @@
 #![warn(rustdoc::broken_intra_doc_links)]
 
 mod analyzer;
+// pub mod evaluation;
```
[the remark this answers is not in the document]
peterhuene: It was not, good catch!

wdl-analysis/src/task/v1.rs
peterhuene: This is also a dead code from very early on which was accidentally committed; I'll remove.

wdl-analysis/src/scope.rs
peterhuene: Whoops, must have context switched away right in the middle of completing that sentence 😅. I'll fix!

wdl-analysis/src/stdlib.rs
peterhuene: It's _mentioned_ in the spec in the same section, but there probably should be explicit overloads for them in the list of signatures. I'll file a spec issue.

wdl-analysis/tests/analysis/name-reference-cycle/source.diagnostics
peterhuene: The DFS stops at the first place it encounters a cyclic reference, so we only report on two locations: the starting point and the node that we've encountered which was previously visited.

If we want, we could walk _a_ path between the two nodes and report on each node we encounter.

wdl-analysis/src/scope/v1.rs
peterhuene: I'll file a tracking issue to complete static analysis by implementing workflow type checks.

wdl-analysis/src/stdlib.rs
peterhuene: Opened https://github.com/openwdl/wdl/issues/683.

wdl-analysis/src/scope/v1.rs
peterhuene: Opened https://github.com/stjude-rust-labs/wdl/issues/167.

a-frantz (approved):

PR #168, opened 2024-08-29
feat: introduce support for the `task` variable from WDL 1.2.

peterhuene:
This commit introduces support for the `task` variable in type checking for WDL 1.2 documents.

It resolves the `task` variable to type `Task`, which has a well-known list of members and their types.

This partially completes #164.

      changes (when appropriate).

### DIFF
--- a/wdl-analysis/src/scope.rs
+++ b/wdl-analysis/src/scope.rs
@@ -12,7 +12,6 @@ use wdl_ast::AstNode;
 use wdl_ast::Diagnostic;
 use wdl_ast::Span;
 use wdl_ast::SupportedVersion;
-use wdl_ast::SyntaxElement;
 use wdl_ast::SyntaxKind;
 use wdl_ast::ToSpan;
 use wdl_ast::WorkflowDescriptionLanguage;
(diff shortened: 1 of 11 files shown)


wdl-analysis/src/scope.rs
```diff
@@ -362,8 +382,12 @@ impl<'a> ScopeRefMut<'a> {
 struct Task {
     /// The span of the task name.
     name_span: Span,
-    /// The scope index for the task.
+    /// The root scope index for the task.
     scope: ScopeIndex,
+    /// The scope index for the outputs.
+    outputs: Option<ScopeIndex>,
+    /// The scope index for the command.
+    command: Option<ScopeIndex>,
```
a-frantz, line 390: If the command block is required, why is this behind an `Option<>`?

wdl-analysis/tests/analysis/task-variable-unsupported/source.diagnostics
a-frantz, line 1: I wonder if these diagnostics should be more informative. Would it be difficult to have these say something like `'task' object only available in WDL >=v1.2`?

wdl-analysis/src/scope.rs
peterhuene: That's a good question.

While the command block is required and we emit a diagnostic for it during validation, analysis is designed in such a way as to still supports analyzing on invalid documents to emit additional diagnostics.

So it's written in places to assume that if we haven't seen something because it's invalid, validation emitted a diagnostic for it and otherwise ignore it.

wdl-analysis/tests/analysis/task-variable-unsupported/source.diagnostics
peterhuene: I can add that!

wdl-analysis/src/scope.rs
a-frantz: Makes sense! Carry on then 👍

wdl-analysis/tests/analysis/task-variable-unsupported/source.diagnostics
peterhuene: How about:

```
the `task` name may only be used within a task command section or task output section using WDL 1.2 or later
```

a-frantz: Seems spot on 💯

claymcleod (approved):

claymcleod (approved):
LGTM.

claymcleod:
FWIW, TIL that in the VSCode extension, it automatically approves a PR (without letting you write a comment with it) before you actually press "approve".

PR #170, opened 2024-09-03
 feat: complete type checking of tasks.

### LINKED ISSUE #166: feat: type check task `runtime`, `requirements`, and `hints` sections
This is required to complete static analysis for tasks from #163 and related to #164, where the `hints` section type check should support `hints`, `input`, and `output` literal expressions for full 1.2 support.

peterhuene:
This commit completes the type checking in tasks to include expressions in
`requirements`, `hints`, and `runtime` sections.

Type checking works off of the "known" keys of those sections and the types
each key supports for an acceptable value.

For the `hints` section, this also includes supporting the hidden `input`,
`output`, and `hints` types and their literal expressions. For `input` and
`output` types, this means ensuring the input and output keys match a declared
input or output and also resolving any dotted form of the names when inputs and
outputs are structs.

Also fixed with these changes:

* Fixed a bug where duplicate tasks were generating duplicate section
  diagnostics instead of being ignored.
* Fixed a bug where duplicate tasks were being type checked with the wrong
  scope instead of being ignored.
* Fixed a failure to detect duplicate aliased items in task `hints` section.

Closes #166 and #164.

      changes (when appropriate).

### DIFF
--- a/wdl-analysis/src/types/v1.rs
+++ b/wdl-analysis/src/types/v1.rs
@@ -75,6 +75,32 @@ pub(crate) fn type_mismatch(
     )
 }
 
+/// Creates a "element type mismatch" diagnostic.
+pub(crate) fn element_type_mismatch(
+    types: &Types,
+    expected: Type,
+    expected_span: Span,
+    actual: Type,
+    actual_span: Span,
+) -> Diagnostic {
+    Diagnostic::error(format!(
+        "type mismatch: expected all elements to be type `{expected}`, but found type `{actual}`",
+        expected = expected.display(types),
+        actual = actual.display(types)
+    ))
+    .with_label(
+        format!("this is type `{actual}`", actual = actual.display(types)),
+        actual_span,
+    )
+    .with_label(
+        format!(
+            "the first element is type `{expected}`",
+            expected = expected.display(types)
+        ),
+        expected_span,
+    )
+}
+
 /// Creates a custom "type mismatch" diagnostic.
 fn type_mismatch_custom(
     types: &Types,
@@ -846,7 +872,7 @@ where
                 for expr in elements {
                     if let Some(actual) = self.evaluate_expr(scope, &expr) {
                         if !actual.is_coercible_to(self.types, &expected) {
-                            self.diagnostics.push(type_mismatch(
+                            self.diagnostics.push(element_type_mismatch(
                                 self.types,
                                 expected,
                                 expected_span,
@@ -919,7 +945,7 @@ where
                             }
 
                             if !actual_value.is_coercible_to(self.types, &expected_value) {
-                                self.diagnostics.push(type_mismatch(
+                                self.diagnostics.push(element_type_mismatch(
                                     self.types,
                                     expected_value,
                                     expected_value_span,
--- a/wdl-analysis/src/types/v1.rs
+++ b/wdl-analysis/src/types/v1.rs
@@ -68,7 +68,7 @@ pub(crate) fn type_mismatch(
     )
     .with_label(
         format!(
-            "this is type `{expected}`",
+            "this expects type `{expected}`",
             expected = expected.display(types)
         ),
         expected_span,
--- a/wdl-analysis/src/types/v1.rs
+++ b/wdl-analysis/src/types/v1.rs
@@ -11,8 +11,11 @@ use wdl_ast::v1::IfExpr;
 use wdl_ast::v1::IndexExpr;
 use wdl_ast::v1::LiteralArray;
 use wdl_ast::v1::LiteralExpr;
+use wdl_ast::v1::LiteralHints;
+use wdl_ast::v1::LiteralInput;
 use wdl_ast::v1::LiteralMap;
 use wdl_ast::v1::LiteralMapItem;
+use wdl_ast::v1::LiteralOutput;
 use wdl_ast::v1::LiteralPair;
 use wdl_ast::v1::LiteralStruct;
 use wdl_ast::v1::LogicalAndExpr;
@@ -22,6 +25,7 @@ use wdl_ast::v1::NegationExpr;
 use wdl_ast::v1::Placeholder;
 use wdl_ast::v1::PlaceholderOption;
 use wdl_ast::v1::StringPart;
+use wdl_ast::version::V1;
 use wdl_ast::AstNodeExt;
 use wdl_ast::AstToken;
 use wdl_ast::Diagnostic;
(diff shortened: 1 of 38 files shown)


peterhuene:
Note: `type_check_task` is getting a bit gnarly. I plan on extracting out a lot of that method's complexity out into an evaluation graph for the task that we can share between task evaluation and task type checking.

wdl-analysis/src/types/v1.rs
```diff
@@ -1004,6 +1046,472 @@ where
...
+
+                true
+            }
+            "memory" => {
+                if !expr_ty.is_coercible_to(self.types, &PrimitiveTypeKind::Integer.into())
+                    && !expr_ty.is_coercible_to(self.types, &PrimitiveTypeKind::String.into())
+                {
+                    self.diagnostics.push(type_mismatch_custom(
+                        self.types,
+                        "type `Int` or type `String`",
```
a-frantz, line 1116: At what point will `String` entries be checked for conforming to the expected format? e.g. they use the proper units etc.

wdl-analysis/src/types/v1.rs
```diff
@@ -1004,6 +1046,472 @@ where
...
+                true
+            }
+            "return_codes" if self.version >= SupportedVersion::V1(V1::Two) => {
+                if !expr_ty.is_coercible_to(self.types, &PrimitiveTypeKind::Integer.into())
+                    && !expr_ty.is_coercible_to(self.types, &PrimitiveTypeKind::String.into())
+                    && !expr_ty.is_coercible_to(self.types, &STDLIB.array_int)
+                {
+                    self.diagnostics.push(type_mismatch_custom(
+                        self.types,
+                        "type `Int`, type `String`, or type `Array[Int]`",
```
a-frantz, line 1200: If this can be a `String` I would also expect `Array[String]` to be supported.

wdl-analysis/tests/analysis/hints-section/source.diagnostics
```diff
@@ -0,0 +1,108 @@
...
+    │         │      
+    │         this expects type `Int` or type `String`
+
+error: type mismatch: expected type `Boolean`, but found type `String`
+    ┌─ tests/analysis/hints-section/source.wdl:119:21
+    │
+119 │         short_task: "false"
+    │         ----------  ^^^^^^^ this is type `String`
+    │         │            
+    │         this is type `Boolean`
```
a-frantz, line 47: I think this would make more sense if:
```suggestion
    │         this expects type `Boolean`
```

wdl-analysis/tests/analysis/hints-section/source.diagnostics
```diff
@@ -0,0 +1,108 @@
...
+128 │             baz.foo.wrong: hints {
+    │                 ^^^
+
+error: type mismatch: expected type `hints`, but found type `String`
+    ┌─ tests/analysis/hints-section/source.wdl:131:18
+    │
+131 │             foo: "wrong"
+    │             ---  ^^^^^^^ this is type `String`
+    │             │     
+    │             this is type `hints`
```
a-frantz, line 81:
```suggestion
    │             this expects type `hints`
```

wdl-analysis/tests/analysis/requirements-section/source.diagnostics
```diff
@@ -0,0 +1,88 @@
...
+   │         │        
+   │         this expects type `Int` or type `String`
+
+error: type mismatch: expected type `Boolean`, but found type `String`
+   ┌─ tests/analysis/requirements-section/source.wdl:65:14
+   │
+65 │         gpu: "false"
+   │         ---  ^^^^^^^ this is type `String`
+   │         │     
+   │         this is type `Boolean`
```
a-frantz, line 55:
```suggestion
   │         this expects type `Boolean`
```

wdl-analysis/src/types/v1.rs
peterhuene: The only supported string value is `*` to indicate any return code is treated as a success.

https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#return_codes

wdl-analysis/src/types/v1.rs
peterhuene: That's a good question; while some of these values could be statically validated, such as having a proper suffix, some of the values will need to be validated by the executor.

I think we could probably do a follow-up commit to add suffix validation for static analysis, at least.

wdl-analysis/src/types/v1.rs
peterhuene: I'll add a check for a value other than `*`.

wdl-analysis/tests/analysis/type-mismatch/source.diagnostics
```diff
@@ -4,69 +4,69 @@ error: type mismatch: expected type `Int`, but found type `String`
...
+   │                      this expects type `Map[Int, String]`
 
 error: type mismatch: expected type `Int`, but found type `String`
    ┌─ tests/analysis/type-mismatch/source.wdl:15:24
    │
 15 │     Array[Int] f = [1, "2", "3"]
    │                     -  ^^^ this is type `String`
    │                     │   
-   │                     this is type `Int`
+   │                     this expects type `Int`
```
a-frantz, line 47: Up until this error, switching `is` -> `expects` makes for clearer diagnostics IMO, but it gets wonky from here down. Can we treat these cases separately?

wdl-analysis/src/types/v1.rs
peterhuene: Oh as these are expressions and may depend on user inputs, there's no way to statically validate these other than when they are specified as string literals; validation of these values should probably be delayed until evaluation.

wdl-analysis/src/types/v1.rs
peterhuene: Actually, that's not something that can be statically validated as I forgot these are expressions and not strictly literals. Validation of the string will need to be done at evaluation time.

wdl-analysis/tests/analysis/type-mismatch/source.diagnostics
peterhuene: I can further specialize this error.

wdl-analysis/src/types/v1.rs
a-frantz: Ok that makes sense 👍

a-frantz (approved):

Issue #174, opened 2024-09-13
Trailing whitespace in comments should be linted


a-frantz:
This check should be added to an existing rule, probably `CommentWhitespace` or maybe just the generic `Whitespace` rule. It should trigger for all comments (lint directives, preamble comments, and "regular" comments).

BradenEverson:
I'd be interested in taking this on! Does this rule live in the `comment_whitespace.rs` file in `wdl-lint`? 

I might need a bit of an explanation of how this is all working, but it looks like the `Visitor` implementation is what's doing the actual checking?

a-frantz:
It could reasonably fit `comment_whitespace.rs`, but I think a slightly better fit would be in https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/rules/whitespace.rs

But yes, the `Visitor` impl is what does the heavy lifting for linting. We (and by we I mean Peter) have architected it so that all the lint rules are enforced with one pass through the syntax tree. The syntax tree is made up of "nodes" and "tokens". Each node will be visited twice, once on `Enter` and once on `Exit`. Each "token" is visited only once.

You'll want to add a comment visitor to `whitespace.rs` that checks for trailing whitespace. Comments are tokens, so are only visited once.

Feel free to reach out either on this issue or over Slack if you need some more direction!

BradenEverson:
Okay makes sense, I'll get started on this all sometime today or tomorrow and let you know if I have any questions over slack, thanks!

PR #177, opened 2024-09-19
feat: Trailing Comment Whitespace Lint

BradenEverson:

This PR adds a new check to the `Whitespace` linting rule that checks for trailing whitespace in all types of comments. It is tested against preamble comments, linting directives and normal comments to ensure this behavior. 

Issue: [174](https://github.com/stjude-rust-labs/wdl/issues/174)

- [X] You have added a few sentences describing the PR here.
- [X] You have added yourself or the appropriate individual as the assignee.
- [X] You have added at least one relevant code reviewer to the PR.

Rule specific checks:

- [X] ~You have added the rule as an entry within `RULES.md`.~
- [X] ~You have added the rule to the `rules()` function in `wdl-lint/src/lib.rs`.~
- [X] You have added a test case in `wdl-lint/tests/lints` that covers every
      is implemented.
- [X] ~If you have implemented a new `Visitor` callback, you have also
      visitor callback will execute.~
- [X] You have run `wdl-gauntlet --refresh` to ensure that there are no 
- [X] You have run `wdl-gauntlet --refresh --arena` to ensure that all of the 
      (`Arena.toml`).

### DIFF
--- a/wdl-lint/CHANGELOG.md
+++ b/wdl-lint/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-* Added Trailing Comments lint rule as an extension of the `Whitespace` rule ([]())
+* Added Trailing Comments lint rule as an extension of the `Whitespace` rule ([177](https://github.com/stjude-rust-labs/wdl/pull/177))
 
 ## 0.6.0 - 09-16-2024
 
--- a/wdl-lint/CHANGELOG.md
+++ b/wdl-lint/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+* Added Trailing Comments lint rule as an extension of the `Whitespace` rule ([]())
+
 ## 0.6.0 - 09-16-2024
 
 ### Fixed
--- a/wdl-lint/CHANGELOG.md
+++ b/wdl-lint/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-* Added Trailing Comments lint rule as an extension of the `Whitespace` rule ([177](https://github.com/stjude-rust-labs/wdl/pull/177))
+* Added comments to the trailing whitespace check of the `Whitespace` rule ([177](https://github.com/stjude-rust-labs/wdl/pull/177))
 
 ## 0.6.0 - 09-16-2024
 
--- a/wdl-lint/CHANGELOG.md
+++ b/wdl-lint/CHANGELOG.md
@@ -8,7 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 
 ### Added
-* Added comments to the trailing whitespace check of the `Whitespace` rule ([177](https://github.com/stjude-rust-labs/wdl/pull/177))
+* Added comments to the trailing whitespace check of the `Whitespace` rule ([#177](https://github.com/stjude-rust-labs/wdl/pull/177))
 
 ## 0.6.0 - 09-16-2024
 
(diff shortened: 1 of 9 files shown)


wdl-lint/src/rules/whitespace.rs
```diff
@@ -81,6 +81,18 @@ impl Rule for WhitespaceRule {
 impl Visitor for WhitespaceRule {
     type State = Diagnostics;
 
+    fn comment(&mut self, state: &mut Self::State, comment: &wdl_ast::Comment) {
+        let comment_str = comment.as_str();
+        if comment_str.ends_with(|char: char| char.is_whitespace()) {
+            // Trailing whitespace
+            state.exceptable_add(
+                trailing_whitespace(comment.span()),
```
a-frantz, line 89: The `Span` passed into `trailing_whitespace()` gets highlighted (or underlined with a squiggle). Here, you are highlighting the entire comment in the diagnostic with `comment.span()`.

It would be preferable if the diagnostic only highlighted the trailing whitespace (and not just the last character. It should should look backwards and highlight any continuous whitespace at the end of the string).

Can you make that change? After you do that, you should update the test you added with a variety of trailing whitespace to make sure it all gets highlighted correctly.

wdl-lint/CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Added
+* Added Trailing Comments lint rule as an extension of the `Whitespace` rule ([177](https://github.com/stjude-rust-labs/wdl/pull/177))
```
a-frantz, line 11:
```suggestion
* Added comments to the trailing whitespace check of the `Whitespace` rule ([177](https://github.com/stjude-rust-labs/wdl/pull/177))
```
just nit picking over language 😄

a-frantz (changes requested):
Great work @BradenEverson ! Two small tweaks I'd like to see, but I think this is in pretty great shape!

wdl-lint/src/rules/whitespace.rs
BradenEverson: Thanks for the explanation, I'll get that fixed up after my classes today :+1:

adthrasher: @a-frantz - The printed diagnostic (which predates this PR) is `remove the trailing whitespace from this line`. Should that be `remove this trailing whitespace`, since we're highlighting just the excess whitespace in our diagnostics?

a-frantz: That's a good catch, I think you're right. @BradenEverson this is a nit over language that predates your PR, but should be fixed while we're working on the file. [Line we're discussing](https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/rules/whitespace.rs#L37). Please make the change to language that @adthrasher pointed out.

a-frantz: @adthrasher similar thought of just cleaning up while we're here: should `trailing whitespace` really be at a `warning` level? That feels more like `note` territory. Not sure how important that distinction is.

adthrasher: After spamming @a-frantz's Slack, I agree that `note` is more appropriate for this lint.

adthrasher: That probably applies to all three diagnostics in this rule.

BradenEverson: The span has been updated to only highlight the trailing whitespace and I updated the language to what you said @adthrasher. Do we want to also do the change from warning to note for all Whitespace diagnostics as a part of this PR?

BradenEverson: > The span has been updated to only highlight the trailing whitespace and I updated the language to what you said @adthrasher. Do we want to also do the change from warning to note for all Whitespace diagnostics as a part of this PR?

I pushed a commit changing all Whitespace diagnostics to notes, if this isn't the way we want to move with this we can revert the commit.

adthrasher (approved):

a-frantz (approved):
LGTM! Nice work @BradenEverson

PR #179, opened 2024-09-20
ci: automation

a-frantz:
First stage of our new release process!

This PR adds a manually triggered workflow which will bump all our crate versions and submit a PR with those changes. That will then be manually reviewed and merged in.

Next (not included in this PR) will be tagging those merge commits and creating releases from them.


.github/workflows/release.yml
```diff
@@ -0,0 +1,30 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      crate:
+        description: Crate to bump (anything which depends on this crate will also be bumped)
+        type: choice
+        default: wdl-grammar
```
peterhuene, line 9: Would the default of `wdl` make sense?

peterhuene (commented):
Looks great! Just the one comment.

.github/workflows/release.yml
a-frantz: The current default (`wdl-grammar`) is the same as saying "bump everything" which I think is our most common use. Specifying `wdl` would only bump that individual crate because it sits at the top of the dependency tree.

peterhuene: Oh right, I was thinking in _publish_ order, so basically the reverse of what we want. Makes sense 👍.

peterhuene (approved):

PR #184, opened 2024-09-24
ci: release

a-frantz:
Part 2 of the new release process!

When there's a tag matching`wdl-v*` pushed to `main`, this kicks off publishing to crates.io and creating a GitHub release. The GitHub release will have auto-generated release notes based on the merged PRs since the last release. Example [here](https://github.com/stjude-rust-labs/wdl/releases/tag/untagged-db7e6b7505c9a55fa426)

This also features a change to the previous `bump` step where CHANGELOGS are now updated as well.

### DIFF
--- a/ci/src/main.rs
+++ b/ci/src/main.rs
@@ -63,7 +63,8 @@ const IGNORE_PATHS: &[&str] = &["target", "tests", "examples", "benches", "book"
 #[derive(Debug, Clone)]
 struct Crate {
     manifest: DocumentMut,
-    path: PathBuf,
+    manifest_path: PathBuf,
+    changelog_path: Option<PathBuf>,
     name: String,
     version: String,
     should_bump: bool,
@@ -157,7 +158,7 @@ fn main() {
             // publish in a loop and we remove crates once they're successfully
             // published. Failed-to-publish crates get enqueued for another try
             // later on.
-            for _ in 0..10 {
+            for _ in 0..3 {
                 all_crates.retain(|krate| !publish(&krate.borrow(), dry_run));
 
                 if all_crates.is_empty() {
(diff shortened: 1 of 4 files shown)


claymcleod:
I don't want to use git commits as the `CHANGELOG.md` or release notes, as I generally agree with the philosophy laid out [here](https://keepachangelog.com/en/1.0.0/).

- Changelogs should remain updated in each commit and should be manually added by the software engineer who made the changes. I don't want any automation editing the changelogs.
- Release notes can be autogenerated from parsing the `CHANGELOG.md`(s) upon release if you'd like, but that will require writing parsing code for the way the changelogs are structured.

a-frantz:
I think we're in _almost_ total agreement here. I want us to continue manually maintaining the 6 different CHANGELOGs we have, and I am not proposing we automate any of that (other than the insertion of version+date headers which will still have to undergo manual review before making it to `main`).

I view the release notes we show on GitHub to be a completely separate from the changelogs we maintain.

> * Release notes can be autogenerated from parsing the `CHANGELOG.md`(s) upon release if you'd like, but that will require writing parsing code for the way the changelogs are structured (and they need to generate the notes according to the process laid out in `RELEASE.md`).

I think this is a bad idea. As the "Keep a changelog" document says, changelogs are designed for humans, not machines. The format is very resistant to _good_ automation, and most anything we write to parse them is going to be error-prone.

So I believe we're in agreement about the changelogs and that can be considered settled. But the release notes issue I see as unsolved. Our constraints when filling out release notes are:
a. continue to manually write a bunch of distinct changelogs
b. bc of the format and manual nature of (1), we can't write [good] automation that can parse those changelogs

I see a few potential solutions:
1. dump git logs using GitHub's automation
    * (which I'll say I'm surprised by how nice these turn out. I didn't expect much from log dumping, but they did pretty good with cleaning it up)
2. Leave release notes empty
3. Populate the release notes with static links to the changelogs
    * i.e. the release notes will be more or less identical for every release (aside for any changes to which CHANGELOGs we point to) 
4. don't automate the release and write it by hand 

I think any of those 4 are reasonable courses of action for this repo, though option (4) is definitely in last place for me personally.

peterhuene:
I personally think it's fine for the _GitHub release_ to be auto-generated from the commit history with the individual `CHANGELOG.md` expected to be maintained with every commit into the repo.

ci/src/main.rs
```diff
@@ -223,18 +233,35 @@ fn bump_version(krate: &Crate, crates: &[Rc<RefCell<Crate>>], patch: bool) {
...
+                &format!(
+                    "## Unreleased\n\n## {} - {}",
+                    bump(&krate.version, patch),
+                    todays_date
+                ),
+            );
+            fs::write(changelog_path, changelog).expect("failed to write changelog");
+        }
+    }
 }
```
peterhuene, line 265: Does this happen even if we fail to publish because the crate has already been published previously?

That is to say, I wouldn't want this to continually insert the `## Unreleased` when we've done so for the previous one.

Do we need to check to see if the file contents contains `format!("## {} - ", bump(&krate.version, patch))`?

peterhuene: Oh never mind, this only happens for the `bump` command.

peterhuene (approved):
LGTM!

Issue #193, opened 2024-09-26
rewrite or remove existing `fix` messages


a-frantz:
Many of our `fix` messages are redundant, obvious, or otherwise unhelpful. We should use the `fix` feature more mindfully. That means supplying the "corrected" code or **adding** something to the diagnostic that isn't already conveyed. And if those can't be accomplished, the `fix` should be omitted.

related to #192

a-frantz:
Copying some thoughts I typed up in slack:

Existing diagnostic:
```
note[BlankLinesBetweenElements]: missing blank line
   ┌─ tests/lints/blank-lines-between-elements/source.wdl:17:5
   │  
17 │ ╭     scatter (i in ["hello", "world"]) {
18 │ │         call bar { input: s = i }
19 │ │     }
   │ ╰─────^
   │  
   = fix: add a blank line before this element
```
We discussed being more intentional with the `fix` messages. So we probably want to drop this one as it doesn't add very much. That leaves:
```
note[BlankLinesBetweenElements]: missing blank line
   ┌─ tests/lints/blank-lines-between-elements/source.wdl:17:5
   │  
17 │ ╭     scatter (i in ["hello", "world"]) {
18 │ │         call bar { input: s = i }
19 │ │     }
   │ ╰─────^
```
Problem here is that now the Span doesn't make sense. Why are lines 18-19 included? Maybe the span's `len` should be `0`? That creates:
```
note[BlankLinesBetweenElements]: missing blank line
   ┌─ tests/lints/blank-lines-between-elements/source.wdl:17:5
   │
17 │     scatter (i in ["hello", "world"]) {
   │     ^
```
Ideally, I want an arrow that points _between lines 16 and 17_ and indicates _a blank goes here_. IDK how to do that in codespan. Can we do that?

Created by hand, but something like:
```note[BlankLinesBetweenElements]: missing blank line
   ┌─ tests/lints/blank-lines-between-elements/source.wdl:17:5
16 │     input {}
---->
17 │     scatter (i in ["hello", "world"]) {
```

(We can't do that in `codespan`)

The moral of this experiment is that reworking the `fix` messages may also require changing up some of the reported Spans to make more sense. That can be a bit tricky depending on how the rule is written. So just something to be mindful of.

a-frantz:
It was decided to reverse the course of this issue and instead ensure that _every_ lint diagnostic has a `fix` included (#260 )

PR #186, opened 2024-09-25
docs: CONTRIBUTING.md

a-frantz:
I opted for an FAQ format. Maybe the file should be called `FAQ.md` instead? But if we make that change, it should probably broaden in scope as all the current questions are very much focused on the development side of things.

Closes #178 and #181

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.0.0/

PR #194, opened 2024-09-26
WIP feat: Malformed Lint Directive Rule 

michaelgattas:
This feature is currently a work in progress, but I'm creating a PR so Andrew can offer some help troubleshooting. 
- **Rule Name**: `MalformedLintDirective`

I am having difficulty unwrapping the `comment.syntax().prev_token()` in `malformed_lint_directive.rs', within the `comment` function.

I have tried using both expect() and unwrap(), but suspect that I'm missing some key error handling in accessing the previous token when there is no previous token to unwrap. I figure there's still some knowledge gaps for me in exactly how data is passed into this Comment function.


wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,139 @@
...
+    fn document(&mut self, _: &mut Self::State, _: VisitReason, _: &Document, _: SupportedVersion) {
+        // This is intentionally empty, as this rule has no state.
+    }
+
+    fn comment(&mut self, state: &mut Self::State, comment: &Comment) {
+        // println!("Comment: {:?}", comment);
+        if let x = comment.syntax().prev_token() {
+            let y = x.unwrap().kind();
+            println!("{:?}", y);
+        }
```
a-frantz, line 94: For production code, we generally prefer using `expect("<the assumption we're making>")` instead of `unwrap()`. This is because `expect()` is self-documenting, where the message communicates the assumption. Whereas `unwrap()` is usually accompanied by a comment explaining the assumption. `expect()` is just a bit nicer than `unwrap()`, but they accomplish the same thing.

But the key for both of them is that we only use them when we are dang-near positive they won't `panic` in production. Anytime they get hit, that's considered an unhandled bug (IMO at least).

Here, a comment may or may not have a `prev_token()` so the `None` case needs to be handled. You probably want something like:
```suggestion
        if let Some(prev) = comment.syntax().prev_token() {
            // handle the case that something occurs before the comment
        } else {
            // handle the case that the comment is the first thing in the document
        };
```

a-frantz:
Don't worry about the existing history, but going forward this repo likes to use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), so check out that link and start using those for you commit messages. Clearly a biased source here, but we all think they make for a much cleaner git history and you should probably use them going forward.

And it's fine to leave as is while your PR is in draft state, but make sure to go back and edit your PR description so that it fits the complete template (if you deleted it, you can check out the source [here](https://github.com/stjude-rust-labs/wdl/blob/main/.github/pull_request_template.md?plain=1))

michaelgattas:
@a-frantz I've updated my branch to implement two Diagnostics in the malformed_lint_directive rule:
1. `inline_lint_directive` - assigned when a comment is recognized as an inline comment and a lint directive
2. `invalid_lint_directive` - assigned when the lint directive keyword is not in a formally declared list of lint directives (current, just "except").

Something is going wrong in trying to call the `.with_fix()` in my `invalid_lint_directive` method, but I can't quite crack it. Please review and let me know if I'm missing anything, and any other sloppy code in the branch that can be improved. 

Thanks!  🚀

wdl-lint/src/util.rs
```diff
@@ -1,5 +1,16 @@
 //! A module for utility functions for the lint rules.
 
+use wdl_ast::{AstToken, Comment, SyntaxKind};
+
+/// Detect is a comment is in-line or not by looking for `\n` in the prior
+/// whitespace. Used in CommentWhitespace and MalformedLintDirective rules.
```
a-frantz, line 6:
```suggestion
/// Detect if a comment is in-line or not by looking for `\n` in the prior
/// whitespace.
```
Tip: when writing comments (or really writing code in general) try and think about whether what you're writing is going to be easy to maintain.

e.g. this sentence I cut out, I removed it because I don't want to keep this updated if I ever start using the function in some third or fourth rule and I also don't want the comment to be incorrect. Best just not to say anything in this case.

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,103 @@
...
+            if is_inline_comment(comment) {
+                state.add(inline_lint_directive(
+                    comment.span(),
+                ));
+            } else if let Some(directive) = comment.as_str().strip_prefix("#@") {
+                if !ACCEPTED_LINT_DIRECTIVES.contains(&directive) {
+                    state.add(invalid_lint_directive(
+                        comment.span(),
+                    ));
+                }
```
a-frantz, line 95: A lint directive can be _both_ an inline comment and invalid, right? I think we would want to log both diagnostics in that case.

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,103 @@
...
+/// Creates an "Invalid Lint Directive" diagnostic.
+fn invalid_lint_directive(
+    span: Span,
+) -> Diagnostic {
+    let accepted_directives = ACCEPTED_LINT_DIRECTIVES.join(", ");
+    Diagnostic::warning(
+        "lint directive not recognized")
+        .with_rule(ID)
+        .with_label("lint directive not recognized", span)
+        .with_fix(format!("remove unrecognized lint directives, use any of the directives included: [{:#?}]", accepted_directives))
```
a-frantz, line 45: Can you respond with the exact error you're seeing? This looks wrong to my eye, but can't put my finger on what the issue is without the error messaging.

wdl-lint/src/util.rs
michaelgattas: Makes sense, and sorry for the typo!

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,103 @@
...
+
+
+    fn tags(&self) -> TagSet {
+        TagSet::new(&[Tag::Clarity, Tag::Correctness])
+    }
+
+    fn exceptable_nodes(&self) -> Option<&'static [wdl_ast::SyntaxKind]> {
+        // Some(&[SyntaxKind::VersionStatementNode])
+        None
+    }
```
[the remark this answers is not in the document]
michaelgattas: I think I might need some more guidance on how `exceptable_nodes` works, but I'll follow your direction in wrapping this issue up nonetheless.

wdl-lint/src/rules/malformed_lint_directive.rs
michaelgattas: Yes, you're right. I'll modify the code to allow for a Diagnostic creation in both cases.

wdl-lint/src/rules/malformed_lint_directive.rs
michaelgattas: The error message I get it: 
```
---- visitor::test::it_supports_reuse stdout ----
thread 'visitor::test::it_supports_reuse' panicked at wdl-lint/src/visitor.rs:411:14:
should not have any diagnostics: [Diagnostic { rule: Some("MalformedLintDirective"), severity: Warning, message: "lint directive not recognized", fix: Some("remove unrecognized lint directives, use any of the directives included: [\"except\"]"), labels: [Label { message: "lint directive not recognized", span: Span { start: 15, end: 53 } }] }]
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    visitor::test::it_supports_reuse
```
Previously, I have encountered this error when I wasn't safely handling an Option, but now I'm a bit turned around on what's causing it. My hunch is that the `invalid_lint_directive()` is causing the error, because commenting out code and re-testing, the problem appears with this code active.

wdl-lint/src/rules/malformed_lint_directive.rs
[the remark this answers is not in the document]
a-frantz: I'll write something up about `exceptable_nodes() and `exceptable_add()` and put it in #186 

I'll ping you once I've pushed that Q+A!

wdl-lint/src/rules/malformed_lint_directive.rs
a-frantz: It's not very obvious what's happening here, but the problem is that you're new rule is adding a diagnostic on the [WDL document created by `it_supports_reuse`](https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/visitor.rs#L392-L399). I'm not sure why that is, probably some buggy logic in your detection of an invalid lint directive (because the lint directive in that document looks fine).

The `it_supports_resuse` test assumes that the WDL snippet I linked is lint-diagnostic-free, and your new rule is inserting one.

a-frantz: (That is a kind of an annoying assumption for the test to make and we should consider finding a different way to write this test)

wdl-lint/src/rules/malformed_lint_directive.rs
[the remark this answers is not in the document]
a-frantz: https://github.com/stjude-rust-labs/wdl/pull/186/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R82

Hope this helps. LMK if it doesn't 😅

wdl-lint/src/rules/malformed_lint_directive.rs
michaelgattas: @a-frantz I hope I'm not misinterpreting the presence of these errors, but where should I go from here?

a-frantz: Lines 90 and 91 aren't doing what you expect them to, so the following `add()` is inserting a diagnostic in `it_supports_reuse` when one isn't expected. Rework those two lines until this test stops failing.

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,103 @@
...
+
+    fn comment(&mut self, state: &mut Self::State, comment: &Comment) {
+        if comment.as_str().starts_with("#@") {
+            if is_inline_comment(comment) {
+                state.add(inline_lint_directive(
+                    comment.span(),
+                ));
+            } else if let Some(directive) = comment.as_str().strip_prefix("#@") {
+                if !ACCEPTED_LINT_DIRECTIVES.contains(&directive) {
```
a-frantz, line 91: Example comment from the `it_supports_reuse` test: `#@ except: MissingMetas, MissingOutput`

After line 90, `directive == " except: MissingMetas, MissingOutput"` which triggers line 91.

michaelgattas: Thank you very much for helping me dig into this bug. How were you able to see the output of the function for that test?

a-frantz: O, I just followed it through in my head 😅 I didn't actually run the code, so what I shared could be wrong lol.

`comment.as_str() == "#@ except: MissingMetas, MissingOutput"` and `strip_prefix("#@")` removes the first two characters, leaving everything else.

Although printing that could be accomplished by writing `dgb!(&directive);` between lines 90 and 91 and then running the tests.

michaelgattas: You were 100% right, so I've got the test passing now and expected output in my Rule test. I just need to spend a little time fine-tuning my test case so it only violates the intended rule and then will commit and push again, and clean up my PR description and make sure I follow all the steps for full review. Thank you again for all your help!

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,111 @@
...
+/// Creates an "Inline Lint Directive" diagnostic.
+fn inline_lint_directive(
+    span: Span,
+) -> Diagnostic {
+    Diagnostic::warning(
+        "lint directive must be on its own line")
+        .with_rule(ID)
+        .with_label("malformed lint directive", span,
+        )
+        .with_fix("place lint directive on new line")
```
a-frantz, line 30: re: #193 

I think it's best to leave this without a `with_fix()`. You were probably following the existing examples, which I want to clean up when I get a chance.

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,111 @@
...
+            if is_inline_comment(comment) {
+                state.add(inline_lint_directive(
+                    comment.span(),
+                ));
+            }
+
+            let mut directive = lint_directive.trim().split(" ").next().expect("directive");
+            if !directive.ends_with(":") {
+                state.add(no_colon_detected(
+                    Span::new(3 + directive.chars().count(), 1),
```
a-frantz, line 100: Here's your problem. With `Span::new(3 + directive.chars().count(), 1)` you are creating a new span pointing at the 3rd character _in the entire document_. I imagine you probably wanted to start the Span 3 characters past the start of the comment? That would be something like:
```suggestion
                    Span::new(comment.span().start() + 3 + directive.chars().count(), 1),
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,111 @@
...
+                ));
+            } else { directive = directive.strip_suffix(":").expect("stripped directive");}
+
+            if !ACCEPTED_LINT_DIRECTIVES.contains(&directive) {
+                state.add(invalid_lint_directive(
+                    Span::new(3, directive.chars().count()),
```
a-frantz, line 106: same deal here

Gauntlet.toml
a-frantz, line 1: just a thought sparked by this PR, but not to be addressed now.

@peterhuene there has to be a better way to deal with commit hash updates. I just scanned by eye, and there's no diagnostic changes in this PR (as we expect), but that's an annoying and error prone check to do.

Maybe in the future we add a new subcommand to gauntlet for updating the commit hashes, and then PRs like this one shouldn't run that. So `--refresh` would just update diagnostics based on the hashes present in the TOML.

The issue there is that we need to update the hashes at some interval. Maybe a github action that runs on a schedule? Once a week check each repo for new commits?

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,119 @@
...
+/// The identifier for the Malformed Lint Directive rule.
+const ID: &str = "MalformedLintDirective";
+/// The accepted lint directives.
+const ACCEPTED_LINT_DIRECTIVES: [&str; 1] = ["except"];
+
+/// Creates an "Inline Lint Directive" diagnostic.
+fn inline_lint_directive(span: Span) -> Diagnostic {
+    Diagnostic::warning("lint directive must be on its own line")
+        .with_rule(ID)
+        .with_label("malformed lint directive", span)
```
a-frantz, line 26:
```suggestion
        .with_label("inlined lint directive", span)
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,119 @@
...
+            accepted_directives
+        ))
+}
+
+/// Creates a "Missing Whitespace" diagnostic.
+fn missing_whitespace(span: Span) -> Diagnostic {
+    Diagnostic::warning("missing whitespace before lint directive")
+        .with_rule(ID)
+        .with_label("missing whitespace before lint directive", span)
+}
```
a-frantz, line 46: Same here with the repeated language. We should try to switch it up.
```suggestion
fn missing_whitespace(span: Span) -> Diagnostic {
    Diagnostic::warning("expected exactly one space before lint directive")
        .with_rule(ID)
        .with_label("missing whitespace", span)
}
```

a-frantz: Which actually, this makes me realize there's an unhandled edge case. I think that the comment `#@      except: Foo` would get passed your checks without logging any diagnostics, but it also wouldn't be recognized by the directive logic. That's a potentially silent and confusing error.

Can you update your check to see if the present whitespace is _exactly_ one space?

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,119 @@
...
+
+/// Creates a "No Colon Detected" diagnostic.
+fn no_colon_detected(span: Span) -> Diagnostic {
+    Diagnostic::warning("no colon detected following lint directive")
+        .with_rule(ID)
+        .with_label("no colon detected following lint directive", span)
```
a-frantz, line 52:
```suggestion
    Diagnostic::warning("expected colon following a lint directive")
        .with_rule(ID)
        .with_label("no colon detected", span)
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,119 @@
...
+#[derive(Default, Debug, Clone, Copy)]
+pub struct MalformedLintDirectiveRule;
+
+impl Rule for MalformedLintDirectiveRule {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Flags malformed lint directives"
```
a-frantz, line 65:
```suggestion
        "Flags malformed lint directives."
```
(Every other description ends in a period (or they all should))

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    fn explanation(&self) -> &'static str {
+        "Comments which begin with `#@` must only contain valid lint directives. Lint directives \
+         must be on their own line, only preceded by whitespace."
```
a-frantz, line 70: This text (that I wrote and gave you 😅 ) could use a little more detail. I'll get back to you with a rewrite.

wdl-lint/src/util.rs
```diff
@@ -1,5 +1,23 @@
 //! A module for utility functions for the lint rules.
 
+use wdl_ast::AstToken;
+use wdl_ast::Comment;
+use wdl_ast::SyntaxKind;
+
+/// Detect is a comment is in-line or not by looking for `\n` in the prior
```
a-frantz, line 7:
```suggestion
/// Detect if a comment is in-line or not by looking for `\n` in the prior
```

wdl-lint/tests/lints/preamble-ws/source.errors
a-frantz, line 1: Nice work, you have found a bug! The inline comment detection isn't working here. I'm going to log an issue for this one.

a-frantz (changes requested):
@michaelgattas this looks marvelous! Thank you for working on this. Lodged some nitpicky issues for you to address. Mostly stuff related to keeping this code consistent with the rest of the code-base.

We can definitely get this included for the next release! Very exciting 🎉

wdl-lint/src/rules/malformed_lint_directive.rs
michaelgattas: I added an excessive whitespace diagnostic to account for this case now! It's present in the latest commit

wdl-lint/src/rules/malformed_lint_directive.rs
michaelgattas: @a-frantz pending revision/approval, I think this is the last step that remains in this PR!

a-frantz:
```
        "Comments which begin with `#@` must only contain valid lint directives. Lint directives \
         must be on their own line, only preceded by whitespace. Lint directives should follow the \
         pattern `#@ <directive>: <value>` _exactly_. Currently the only accepted lint directive \
         is `except`. For example, `#@ except: MalformedLintDirective`."
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,137 @@
...
+fn inline_lint_directive(span: Span) -> Diagnostic {
+    Diagnostic::warning("lint directive must be on its own line")
+        .with_rule(ID)
+        .with_label("inlined lint directive", span)
+}
+
+/// Creates an "Invalid Lint Directive" diagnostic.
+fn invalid_lint_directive(span: Span) -> Diagnostic {
+    let accepted_directives = ACCEPTED_LINT_DIRECTIVES.join(", ");
+    Diagnostic::warning("lint directive not recognized")
```
peterhuene, line 39: Can we include the lint directive name in the diagnostic's message?

```suggestion
    Diagnostic::warning(format!("lint directive `{name}` is not recognized"))
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,137 @@
...
+            "use any of the recognized lint directives: [{:#?}]",
+            accepted_directives
+        ))
+}
+
+/// Creates a "Missing Whitespace" diagnostic.
+fn missing_whitespace(span: Span) -> Diagnostic {
+    Diagnostic::warning("expected exactly one space before lint directive")
+        .with_rule(ID)
+        .with_label("missing whitespace", span)
```
peterhuene, line 52:
```suggestion
        .with_label("expected a single space before this", span)
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+/// Creates a "No Colon Detected" diagnostic.
+fn no_colon_detected(span: Span) -> Diagnostic {
+    Diagnostic::warning("expected colon following a lint directive")
```
peterhuene, line 57:
```suggestion
    Diagnostic::warning("expected a colon to follow a lint directive")
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,137 @@
...
+        .with_rule(ID)
+        .with_label("no colon detected", span)
```
peterhuene, line 59:
```suggestion
        .with_label("expected a colon here", span)
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,137 @@
...
+        .with_rule(ID)
+        .with_label("inlined lint directive", span)
+}
+
+/// Creates an "Invalid Lint Directive" diagnostic.
+fn invalid_lint_directive(span: Span) -> Diagnostic {
+    let accepted_directives = ACCEPTED_LINT_DIRECTIVES.join(", ");
+    Diagnostic::warning("lint directive not recognized")
+        .with_rule(ID)
+        .with_label("lint directive not recognized", span)
```
peterhuene, line 41: This label feels redundant; I think we should use `with_highlight` instead of `with_label` for this diagnostic.

```suggestion
        .with_highlight(span)
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,137 @@
...
+    Diagnostic::warning("expected exactly one space before lint directive")
+        .with_rule(ID)
+        .with_label("excessive whitespace", span)
+}
+
+/// Creates an "Inline Lint Directive" diagnostic.
+fn inline_lint_directive(span: Span) -> Diagnostic {
+    Diagnostic::warning("lint directive must be on its own line")
+        .with_rule(ID)
+        .with_label("inlined lint directive", span)
```
peterhuene, line 33: This label feels redundant; I suggest using `with_highlight` and use a fix too:

```suggestion
        .with_highlight(span)
        .with_fix("move the lint directive to its own line")
```

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,137 @@
...
+/// The identifier for the Malformed Lint Directive rule.
+const ID: &str = "MalformedLintDirective";
+/// The accepted lint directives.
+const ACCEPTED_LINT_DIRECTIVES: [&str; 1] = ["except"];
+
+/// Creates an "Excessive Whitespace" diagnostic.
+fn excessive_whitespace(span: Span) -> Diagnostic {
+    Diagnostic::warning("expected exactly one space before lint directive")
+        .with_rule(ID)
+        .with_label("excessive whitespace", span)
```
peterhuene, line 26:
```suggestion
        .with_label("replace this whitespace with a single space", span)
```

wdl-lint/tests/lints/preamble-ws/source.errors
```diff
@@ -4,6 +4,12 @@ note[PreambleFormatting]: unnecessary whitespace in document preamble
 1 │     #@ except: DescriptionMissing
   │ ^^^^
 
+warning[MalformedLintDirective]: lint directive must be on its own line
```
peterhuene, line 7: Should we correct this in the `preamble-ws` test's source so that we don't need to update the test's diagnostics baseline?

peterhuene (commented):
Looks great!

Most of my feedback is centered around the new diagnostics (wording and labels).

We definitely need to handle `#@` as a comment too.

Gauntlet.toml
peterhuene: Yeah it is unfortunate churn. We could also perhaps change gauntlet so that it doesn't write a new `Gauntlet.toml` if there's no changes to the actual diagnostics (i.e. we don't bother bumping a repo's commit SHA until there's been a meaningful change to the repo's diagnostic set).

wdl-lint/src/rules/malformed_lint_directive.rs
```diff
@@ -0,0 +1,137 @@
...
+            if lint_directive.starts_with("  ") {
+                let leading_whitespace_len =
+                    lint_directive.len() - lint_directive.trim_start().len();
+                state.add(excessive_whitespace(Span::new(
+                    base_offset + 2,
+                    leading_whitespace_len,
+                )));
+            }
+
+            let mut directive = lint_directive.trim().split(" ").next().expect("directive");
```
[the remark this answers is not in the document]
michaelgattas: Excellent, I've added that new diagnostic earlier in the rule logic (before whitespace checking), but am curious: should I just `unwrap()` the value here instead? is there a better way to unwrap the Option returned by the iterator that avoids panicking? or is the thinking that with the check earlier in the function, this unwrap will never panic?

wdl-lint/tests/lints/preamble-ws/source.errors
a-frantz: That space is important to be checked for the `preamble-ws` rule, so we can't get rid of it. We could instead `except` `MalformedLintDirective` so this doesn't show up, but IMO we should leave it as an example of the bug. It should be a pretty trivial fix, and I can hopefully get to it today before we release, so don't think it matters much practically.

wdl-lint/src/rules/malformed_lint_directive.rs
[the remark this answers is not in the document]
a-frantz: So there's a bunch of ways to handle the `None` variant of `Option` without using `unwrap()/expect()`.
Some examples:
```
match <>.next() {
    Some(val) => <use the returned val>,
    None => <handle the None case>,
}
```
or
```
if let Some(val) = <>.next() {
    <use val>
}
<we are ignoring the None case and doing nothing with it>
```
or
```
let val = <>.next().unwrap_or(<default val>);
```
And there's more too!

------------
@peterhuene I just checked out the `split()` method, and if I'm reading it correctly I think Michael's `expect()` here is actually fine. `split()` will always return at least one element, so calling `next()` on it should always yield `Some()` (although the returned str might be empty, which I think might cause a panic a few lines later).

I think it should get a rewrite though just because that's not very clear. @michaelgattas can you take another stab at writing this without using `expect()` or `unwrap()`?

Gauntlet.toml
a-frantz: That could work, and is probably simpler than what I proposed. But I think the issue would still remain that when changes _do_ happen, there will be a lot of noise mixed in with the stuff we care about.

I envision there being two types of changes to `Gauntlet.toml` and they won't overlap in the same PR:
1. diagnostic changes (without hash changes)
2. hash changes (which may or may not include diagnostic changes)

If that sounds good to you, I'm going to write it up in an issue.

a-frantz (approved):

wdl-lint/src/rules/malformed_lint_directive.rs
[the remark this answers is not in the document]
peterhuene: @a-frantz you're right, for some reason I thought `split` on an empty string would return an empty iterator. It won't panic here then.

PR #206, opened 2024-10-01
fix: lint tweaks

a-frantz:

Lots of relatively minor tweaks, mostly to how whitespace is reported in lints.

      changes (when appropriate).

### DIFF
--- a/wdl-lint/src/rules/blank_lines_between_elements.rs
+++ b/wdl-lint/src/rules/blank_lines_between_elements.rs
@@ -98,7 +98,7 @@ impl Rule for BlankLinesBetweenElementsRule {
          `requirements`, or `hints` section. For workflows, the `workflow body` includes any \
          private declarations, call statements, conditional statements, and scatter statements. A \
          `task body` is any and all private declarations. Within a workflow or task body, \
-         individual elements may optionally be separated by a blank line. "
+         individual elements may optionally be separated by a blank line."
     }
 
     fn tags(&self) -> TagSet {
@@ -141,7 +141,6 @@ impl Visitor for BlankLinesBetweenElementsRule {
     }
 
     // Import spacing is handled by the ImportWhitespace rule
-    // fn import_statement
 
     fn task_definition(
         &mut self,
@@ -156,6 +155,7 @@ impl Visitor for BlankLinesBetweenElementsRule {
         let first = is_first_element(task.syntax());
         let actual_start = skip_preceding_comments(task.syntax());
         check_prior_spacing(&actual_start, state, true, first, &self.exceptable_nodes());
+        check_last_token(task.syntax(), state, &self.exceptable_nodes());
     }
 
     fn workflow_definition(
@@ -171,6 +171,7 @@ impl Visitor for BlankLinesBetweenElementsRule {
         let first = is_first_element(workflow.syntax());
         let actual_start = skip_preceding_comments(workflow.syntax());
         check_prior_spacing(&actual_start, state, true, first, &self.exceptable_nodes());
+        check_last_token(workflow.syntax(), state, &self.exceptable_nodes());
     }
 
     fn metadata_section(
@@ -190,6 +191,7 @@ impl Visitor for BlankLinesBetweenElementsRule {
         let actual_start = skip_preceding_comments(section.syntax());
         check_prior_spacing(&actual_start, state, true, first, &self.exceptable_nodes());
         flag_all_blank_lines_within(section.syntax(), state, &self.exceptable_nodes());
+        check_last_token(section.syntax(), state, &self.exceptable_nodes());
     }
 
     fn parameter_metadata_section(
@@ -560,6 +562,29 @@ fn check_prior_spacing(
                 }
             }
         }
+    }
+}
+
+/// Check that the node's last token does not have a blank before it.
+fn check_last_token(
+    syntax: &SyntaxNode,
+    state: &mut Diagnostics,
+    exceptable_nodes: &Option<&'static [SyntaxKind]>,
+) {
+    if syntax
+        .last_token()
+        .expect("Node should have last token")
+        .prev_sibling_or_token()
+        .is_some_and(|p| {
+            p.kind() == SyntaxKind::Whitespace
+                && p.to_string().chars().filter(|c| *c == '\n').count() > 1
+        })
+    {
+        state.exceptable_add(
+            excess_blank_line(syntax.text_range().to_span()),
+            SyntaxElement::from(syntax.clone()),
+            exceptable_nodes,
+        );
     }
 }
 
--- a/wdl-lint/src/rules/blank_lines_between_elements.rs
+++ b/wdl-lint/src/rules/blank_lines_between_elements.rs
@@ -211,6 +211,7 @@ impl Visitor for BlankLinesBetweenElementsRule {
         let actual_start = skip_preceding_comments(section.syntax());
         check_prior_spacing(&actual_start, state, true, first, &self.exceptable_nodes());
         flag_all_blank_lines_within(section.syntax(), state, &self.exceptable_nodes());
+        check_last_token(section.syntax(), state, &self.exceptable_nodes());
     }
 
     fn input_section(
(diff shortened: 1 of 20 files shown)


a-frantz:
closes #198 
closes #203

wdl-lint/src/rules/blank_lines_between_elements.rs
```diff
@@ -557,6 +577,30 @@ fn check_prior_spacing(
...
+    state: &mut Diagnostics,
+    exceptable_nodes: &Option<&'static [SyntaxKind]>,
+) {
+    let prev = syntax
+        .last_token()
+        .expect("Node should have last token")
+        .prev_sibling_or_token();
+    if let Some(prev) = prev {
+        if prev.kind() == SyntaxKind::Whitespace {
+            let count = prev.to_string().chars().filter(|c| *c == '\n').count();
```
peterhuene, line 592: With `prev` now a `SyntaxToken` with the above change, we can use `text` here:

```suggestion
            let count = prev.text().chars().filter(|c| *c == '\n').count();
```

That saves us a string allocation; we should probably edit other places in this file (and others in `wdl-lint` that we could use `into_token` with `text` to save on a string allocation.

wdl-lint/src/rules/blank_lines_between_elements.rs
```diff
@@ -557,6 +577,30 @@ fn check_prior_spacing(
...
+/// Check that the node's last token does not have a blank before it.
+fn check_last_token(
+    syntax: &SyntaxNode,
+    state: &mut Diagnostics,
+    exceptable_nodes: &Option<&'static [SyntaxKind]>,
+) {
+    let prev = syntax
+        .last_token()
+        .expect("Node should have last token")
+        .prev_sibling_or_token();
```
peterhuene, line 589: Since we know we're looking for a whitespace token, we can cast the element into a token like so:

```suggestion
        .prev_sibling_or_token()
        .and_then(SyntaxElement::into_token);
```

This will enable us to call `text()` to get the token's text without having to allocate a string below.

wdl-lint/src/rules/preamble_formatting.rs
```diff
@@ -208,12 +202,9 @@ impl Visitor for PreambleFormattingRule {
...
                 // The next token must be part of the version statement
-                // and since we've already established there's a prior comment,
-                // this whitespace must be _exactly_ two newlines.
-                if s != "\r\n\r\n" && s != "\n\n" {
-                    state.add(expected_blank_line_before_version(whitespace.span()));
-                }
-                return;
+                // and we should have already returned during the above check
+                // for the version statement
+                unreachable!("next token should be a comment");
```
peterhuene, line 207: Perhaps just an `assert!` instead of the conditional?

```rust
assert!(next_token.kind() == SyntaxKind::Comment, "next token should be a comment");
```

peterhuene (commented):
These changes look great. Just some minor feedback.

peterhuene (approved):
Looks great!

wdl-lint/src/rules/version_formatting.rs
```diff
@@ -40,21 +41,21 @@ fn expected_blank_line_after_version(span: Span) -> Diagnostic {
...
-    Diagnostic::error("unexpected whitespace before the version statement")
+    Diagnostic::note("unexpected whitespace before the version statement")
         .with_rule(ID)
         .with_highlight(span)
 }
 
 /// Creates a diagnostic for a comment inside the version statement.
 fn comment_inside_version(span: Span) -> Diagnostic {
-    Diagnostic::error("unexpected comment inside the version statement")
+    Diagnostic::note("unexpected comment inside the version statement")
```
adthrasher, line 51: I think this can stay an error. I think that's a reasonable position to take

wdl-lint/tests/lints/missing-comment-space/source.errors
```diff
@@ -1,11 +1,3 @@
-note[CommentWhitespace]: comment delimiter should be followed by a single space
-  ┌─ tests/lints/missing-comment-space/source.wdl:1:1
-  │
-1 │ ##This preamble comment is missing a space.
```
adthrasher, line 4: Is this an expected removal?

wdl-lint/tests/lints/preamble-comment-after-version/source.errors
```diff
@@ -4,14 +4,6 @@ note[PreambleFormatting]: unnecessary whitespace in document preamble
 2 │        ## Bad leading whitespace!
   │ ^^^^^^^
 
-note[CommentWhitespace]: comment has too much indentation
-  ┌─ tests/lints/preamble-comment-after-version/source.wdl:2:8
-  │
-2 │        ## Bad leading whitespace!
```
adthrasher, line 10: Is this correct now?

wdl-lint/tests/lints/one-line-after-version/source.errors
```diff
@@ -4,11 +4,3 @@ error: there must be at least one task, workflow, or struct definition in the fi
 6 │ 
   │  
 
-note[VersionFormatting]: expected exactly one blank line after the version statement
-  ┌─ tests/lints/one-line-after-version/source.wdl:5:12
-  │  
-5 │   version 1.1
```
adthrasher, line 10: Is this right?

wdl-lint/src/rules/preamble_formatting.rs
```diff
@@ -261,7 +250,29 @@ impl Visitor for PreambleFormattingRule {
...
+                    // There's a special case where the blank line has extra whitespace
+                    // but that doesn't appear in the printed diagnostic.
+                    let mut diagnostic = expected_blank_line_before_preamble_comment(span);
+
+                    if s.chars().filter(|&c| c == '\n').count() == 2 {
+                        for (line, start, end) in lines_with_offset(s) {
+                            if !line.is_empty() {
+                                let end_offset = if s.ends_with('\n') {
+                                    1
+                                } else if s.ends_with("\r\n") {
```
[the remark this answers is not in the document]
a-frantz: I believe that as part of the `format` PR @claymcleod is introducing some config for this that could hopefully be reused in `wdl-lint`. So yes, we do want to handle this at a higher level, but that will require some thought as to architecture.

wdl-lint/src/rules/version_formatting.rs
a-frantz: I don't think that jives with the definitions we now have [here](https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md#whats-the-difference-between-error-warning-and-note)

The comment doesn't interrupt parsing/validation/meaning/etc, so `error` would be wrong. I suppose `warning` could fit as it's potentially confusing? But IMO it really just boils down to style (i.e. a `note`), although a particularly egregious example of bad style.

wdl-lint/tests/lints/missing-comment-space/source.errors
a-frantz: Yes, click expand down and you'll see:
```
note[PreambleFormatting]: preamble comments must start with `##` and have at least one space between the `##` and the comment text
  ┌─ tests/lints/missing-comment-space/source.wdl:1:1
  │
1 │ ##This preamble comment is missing a space.
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

wdl-lint/tests/lints/one-line-after-version/source.errors
a-frantz: Maybe? I changed the code so that `VersionFormatting` won't emit anything if it's the last thing (aside from whitespace) in the document. That case is a clear `error` (which is already being handled), and the lints emitted at the same time seem inconsequential to me 🤷

wdl-lint/tests/lints/preamble-comment-after-version/source.errors
a-frantz: Yes. Expand 'up' to see:
```
note[PreambleFormatting]: unnecessary whitespace in document preamble
  ┌─ tests/lints/preamble-comment-after-version/source.wdl:2:1
  │
2 │        ## Bad leading whitespace!
  │ ^^^^^^^
```

PR #209, opened 2024-10-03
fix: improve analysis type checks.

### LINKED ISSUE #208: Audit new gauntlet errors
Now that #207 has merged and gauntlet does a full analysis of the repositories, we should audit the new errors to find false positives or places where we could improve the experience.

## 1

The below type calculation could be improved so that we infer the return type of `select_first` to be `Array[String]` even if `Array[String]` failed the type constraint on (i.e. it is _not_ an optional type).

Currently we're returning `Union` from the type calculation.

```
[[diagnostics]]
document = "broadinstitute/palantir-workflows:/Utilities/WDLs/MatchFingerprints.wdl"
message = "MatchFingerprints.wdl:67:33: error: type mismatch: argument to function `select_first` expects type `Array[X]` where `X`: any optional type, but found type `Array[Array[String]]`"
permalink = "https://github.com/broadinstitute/palantir-workflows/blob/56575fff0bfb76cf873f94f944d673f87734e237/Utilities/WDLs/MatchFingerprints.wdl/#L67"
```

## 2

As a result of the previous diagnostic, we're failing to consider _all_ of the overloads to `length` when called with a `Union` type.

In the below diagnostic, it should include the `Array[X]` and `Map[K, V]` overloads in the diagnostic.

Note that if the above returns `Array[String]` instead of `Union`, it would also eliminate this diagnostic since we would be calling `length` with an `Array[String]` argument type.

```
[[diagnostics]]
document = "broadinstitute/palantir-workflows:/Utilities/WDLs/MatchFingerprints.wdl"
message = "MatchFingerprints.wdl:67:13: error: ambiguous call to function `length` with conflicting signatures `(String) -> Int` and `(Object) -> Int`"
permalink = "https://github.com/broadinstitute/palantir-workflows/blob/56575fff0bfb76cf873f94f944d673f87734e237/Utilities/WDLs/MatchFingerprints.wdl/#L67"
```

## 3

There are _a lot_ of diagnostics relating to primitive type coercions to `String` that aren't [defined in the 1.0 spec](https://github.com/openwdl/wdl/blob/main/versions/1.0/SPEC.md#type-coercion) such as `Int` -> `String`.

This appears to be a widespread issue in these 1.0 documents; if we can find supporting language in the spec to allow such coercions, we should fix the type calculations.

## 4

The following diagnostic should _not_ be generated:

```
[[diagnostics]]
document = "chanzuckerberg/czid-workflows:/workflows/consensus-genome/run.wdl"
message = "run.wdl:247:61: error: type mismatch: expected type `String`, but found type `None`"
permalink = "https://github.com/chanzuckerberg/czid-workflows/blob/a04293a5271176885ce7f876b6353b20da3f7b98/workflows/consensus-genome/run.wdl/#L247"
```

We should calculate the type of the expression:

```
if output_bed then "~{primer_bed}" else None
```

The same goes for any other expression where a common type needs to be inferred, e.g. `["foo", None]` should be `Array[String?]`.

## 5

The following diagnostic is because we're inferring a type of `Array[Union]` for `X` when passed `Array[Union]` (`Union` coerces to `Array[X]` where `X` is any type).

This is also a problem for `Pair` and `Map` compound types.

```
[[diagnostics]]
document = "broadinstitute/palantir-workflows:/FunctionalEquivalence/FunctionalEquivalence.wdl"
message = "FunctionalEquivalence.wdl:236:45: error: type mismatch: argument to function `flatten` expects type `Array[Array[X]]`, but found type `Array[Union]`"
permalink = "https://github.com/broadinstitute/palantir-workflows/blob/56575fff0bfb76cf873f94f944d673f87734e237/FunctionalEquivalence/FunctionalEquivalence.wdl/#L236
```

peterhuene:
* Realize return values of generic functions without constraints if we've failed to bind a function to the arguments; this will help determine the return types of functions like `select_first` and `select_any` when there are argument type mismatches.

* Allow `Union` to bind to generic `Pair`, `Map`, and `Array` as `Pair[Union, Union]`, `Map[Union, Union]`, and `Array[Union]`, respectively.

* Fix common type calculations for `if` expressions, `Array` literals, and `Map` literals.

Fixes #208.

      changes (when appropriate).

### DIFF
--- a/Gauntlet.toml
+++ b/Gauntlet.toml
@@ -24,7 +24,7 @@ commit_hash = "56575fff0bfb76cf873f94f944d673f87734e237"
 
 [repositories."broadinstitute/warp"]
 identifier = "broadinstitute/warp"
-commit_hash = "30e4994f2585ff5ea429e6435acb2312810ef426"
+commit_hash = "78f8c5c3160b06a81648c68f02307e235acdb3a1"
 
 [repositories."chanzuckerberg/czid-workflows"]
 identifier = "chanzuckerberg/czid-workflows"
@@ -118,17 +118,17 @@ permalink = "https://github.com/PacificBiosciences/HiFi-human-WGS-WDL/blob/1cf2b
 
 [[diagnostics]]
 document = "PacificBiosciences/HiFi-human-WGS-WDL:/workflows/wdl-common/wdl/workflows/backend_configuration/backend_configuration.wdl"
-message = "backend_configuration.wdl:25:13: error: type mismatch: expected every map value to be type `Int`, but found type `String`"
+message = "backend_configuration.wdl:25:13: error: type mismatch: a type common to both type `Int` and type `String` does not exist"
 permalink = "https://github.com/PacificBiosciences/HiFi-human-WGS-WDL/blob/1cf2b2e80024290d0ec1ea93b6a279ea2de519b0/workflows/wdl-common/wdl/workflows/backend_configuration/backend_configuration.wdl/#L25"
 
 [[diagnostics]]
 document = "PacificBiosciences/HiFi-human-WGS-WDL:/workflows/wdl-common/wdl/workflows/backend_configuration/backend_configuration.wdl"
-message = "backend_configuration.wdl:26:17: error: type mismatch: expected every map value to be type `Int`, but found type `String`"
+message = "backend_configuration.wdl:26:17: error: type mismatch: a type common to both type `Int` and type `String` does not exist"
 permalink = "https://github.com/PacificBiosciences/HiFi-human-WGS-WDL/blob/1cf2b2e80024290d0ec1ea93b6a279ea2de519b0/workflows/wdl-common/wdl/workflows/backend_configuration/backend_configuration.wdl/#L26"
 
 [[diagnostics]]
 document = "PacificBiosciences/HiFi-human-WGS-WDL:/workflows/wdl-common/wdl/workflows/backend_configuration/backend_configuration.wdl"
-message = "backend_configuration.wdl:27:26: error: type mismatch: expected every map value to be type `Int`, but found type `String`"
+message = "backend_configuration.wdl:27:26: error: type mismatch: a type common to both type `Int` and type `String` does not exist"
 permalink = "https://github.com/PacificBiosciences/HiFi-human-WGS-WDL/blob/1cf2b2e80024290d0ec1ea93b6a279ea2de519b0/workflows/wdl-common/wdl/workflows/backend_configuration/backend_configuration.wdl/#L27"
 
 [[diagnostics]]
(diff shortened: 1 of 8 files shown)


Gauntlet.toml
```diff
@@ -1033,14 +1033,24 @@ permalink = "https://github.com/broadinstitute/palantir-workflows/blob/56575fff0
...
 
+[[diagnostics]]
+document = "broadinstitute/palantir-workflows:/BenchmarkSVs/BenchmarkSVs.wdl"
+message = "BenchmarkSVs.wdl:244:48: error: type mismatch: argument to function `select_all` expects type `Array[X]` where `X`: any optional type, but found type `Array[File]`"
+permalink = "https://github.com/broadinstitute/palantir-workflows/blob/56575fff0bfb76cf873f94f944d673f87734e237/BenchmarkSVs/BenchmarkSVs.wdl/#L244"
+
+[[diagnostics]]
+document = "broadinstitute/palantir-workflows:/BenchmarkSVs/BenchmarkSVs.wdl"
+message = "BenchmarkSVs.wdl:244:75: error: type mismatch: argument to function `select_all` expects type `Array[X]` where `X`: any optional type, but found type `Array[File]`"
+permalink = "https://github.com/broadinstitute/palantir-workflows/blob/56575fff0bfb76cf873f94f944d673f87734e237/BenchmarkSVs/BenchmarkSVs.wdl/#L244"
```
peterhuene, line 1052: These new diagnostics were being hidden by a previous return of `Union`.

Gauntlet.toml
```diff
@@ -2436,235 +2436,240 @@ document = "broadinstitute/warp:/verification/test-wdls/TestWholeGenomeGermlineS
 message = "TestWholeGenomeGermlineSingleSample.wdl:54:45: error: type mismatch: expected type `DragmapReference`, but found type `DragmapReference?`"
 permalink = "https://github.com/broadinstitute/warp/blob/30e4994f2585ff5ea429e6435acb2312810ef426/verification/test-wdls/TestWholeGenomeGermlineSingleSample.wdl/#L54"
 
+[[diagnostics]]
+document = "chanzuckerberg/czid-workflows:/workflows/amr/run.wdl"
+message = "run.wdl:146:24: error: type mismatch: expected type `Array[File]+`, but found type `Array[File]`"
+permalink = "https://github.com/chanzuckerberg/czid-workflows/blob/a04293a5271176885ce7f876b6353b20da3f7b98/workflows/amr/run.wdl/#L146"
```
peterhuene, line 2442: Likewise, this error was hidden by a previous return of `Union`, but now `select_all` is returning `Array[File]` despite it being invalid.

wdl-analysis/src/types/v1.rs
```diff
@@ -1751,4 +1734,33 @@ where
...
+    /// Calculates a common type between two types.
+    ///
+    /// Returns `None` if the types have no common type.
+    fn common_type(&self, first: Type, second: Type) -> Option<Type> {
+        // Check for the first type being coercible to the second type
+        if first.is_coercible_to(self.types, &second) {
+            return Some(second);
+        }
+
+        // Check for the second type being coercible to the second type
```
a-frantz, line 1747:
```suggestion
        // Check for the second type being coercible to the first type
```

peterhuene: Good eye!

a-frantz (approved):

PR #211, opened 2024-10-07
feat: unused import/input/decl/call analysis warnings.

### LINKED ISSUE #200: Implement "unused" analysis warnings
With static analysis mostly completed with #199, it's time to add additional warnings to analysis:

* `unused import`: there must be at least one reference to the namespace or a struct defined in the imported document.
* `unused declaration`: there must be an incoming dependency edge in the evaluation graph for the declaration node.

`miniwdl` has this warning, which I'm not entirely sure we need to implement or not:

* `unused call output`: ensure every output of a call is used or propagated as a workflow output (if the workflow has outputs); similar to "unused declarations", this would be a check to ensure there's an incoming dependency edge in the evaluation graph for the call node.

peterhuene:
This PR implements the first warnings for analysis itself:

* Unused import
* Unused task/workflow input
* Unused task/workflow private declaration
* Unused call in a workflow

The warnings can be excepted the same way a lint directive can: either by having the warning's rule identifier in a preamble `#@ except` comment (i.e. a global except) or by having such a comment immediately precede the import/input/decl/call.

Additionally, the warnings may be treated as errors with a `--deny <id>` option to the `wdl check` command or excepted with `--except <id>`.

Closes #200.

      changes (when appropriate).

### DIFF
--- a/Arena.toml
+++ b/Arena.toml
@@ -1227,11 +1227,6 @@ permalink = "https://github.com/getwilds/ww-vc-trio/blob/c2c13e85efda8ac7aca7f87
 
 [[diagnostics]]
 document = "getwilds/ww-vc-trio:/ww-vc-trio.wdl"
-message = "ww-vc-trio.wdl:362:10: warning[UnusedInput]: unused input `dbSNP_vcf`"
-permalink = "https://github.com/getwilds/ww-vc-trio/blob/c2c13e85efda8ac7aca7f8765fbaa7c2cacb08b2/ww-vc-trio.wdl/#L362"
-
-[[diagnostics]]
-document = "getwilds/ww-vc-trio:/ww-vc-trio.wdl"
 message = "ww-vc-trio.wdl:36:17: warning[SnakeCase]: input name `known_indels_sites_VCFs` is not snake_case"
 permalink = "https://github.com/getwilds/ww-vc-trio/blob/c2c13e85efda8ac7aca7f8765fbaa7c2cacb08b2/ww-vc-trio.wdl/#L36"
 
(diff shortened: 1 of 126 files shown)


peterhuene:
FYI: if we think the `Gauntlet.toml` churn isn't needed, we can simply `except_all` of the analysis diagnostic warnings for gauntlet run; most of the changed lines are coming from updating `Gauntlet.toml`.

wdl-lint/src/lib.rs
```diff
@@ -42,6 +42,14 @@ pub use tags::*;
 pub use visitor::*;
 pub use wdl_ast as ast;
 
+/// The reserved rule names that are used by analysis.
+const RESERVED_RULE_NAMES: &[&str] = &[
```
a-frantz, line 46: This seems backwards to me. `wdl-lint` comes before `wdl-analysis` in the dependency tree, so `wdl-lint` should have "first dibs" on rule names, and `wdl-analysis` has to ensure there's no overlap.

a-frantz (commented):
Should the four new rules be implemented using the `Rule` trait? That's maybe a significant rework and I don't want to hold up this merging, but it seems a better long-term solution.

As currently implemented, these rule IDs won't show up in `sprocket explain` and the new `Config` is going to be messy to maintain if we add a lot of new rules this way.

wdl-lint/src/lib.rs
peterhuene: `wdl-analysis` doesn't depend on `wdl-lint` (I'm not sure why there's an optional dependency in `wdl-analysis`'s `Cargo.toml`, but it never gets activated; it should be removed), so they're actually independent.

`wdl-analysis` doesn't know anything about what validation rules are running; it's just given a validator to use.

I feel like the rules should be reserved inside of `wdl-lint` as analysis is a required step whereas linting is always viewed as "additional".

I can think about this more, though.

Arena.toml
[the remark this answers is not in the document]
peterhuene: We've added about 12 new warnings to `Arena.toml` with these changes; the majority of lines come from bumping the commit SHA in permalinks.

We can certainly disable the analysis warning diagnostics in `Arena.toml`, although we'd still be duplicating analysis errors in both.

wdl-lint/src/lib.rs
a-frantz: Ah I misunderstood. That actually makes more sense now that you point it out, they are quite independent.

I'll think about this some more as well. I'm not sure what the right solution is. I think this is probably ok for at least the short-term, but we may be able to come up with something better for long term.

Arena.toml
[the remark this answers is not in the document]
a-frantz: Would it be difficult to filter out the analysis errors from Arena?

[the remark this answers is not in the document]
peterhuene: It'd be trivial, use a `except_all` diagnostics configuration when arena mode is activated; happy to do that.

peterhuene:
> Should the four new rules be implemented using the Rule trait? That's maybe a significant rework and I don't want to hold up this merging, but it seems a better long-term solution.

Right now there's no dependency on `wdl-lint` from `wdl-analysis` (only that optional dependency that shouldn't be in `Cargo.toml`) and `Rule` operates solely on an AST, whereas these particular warnings look at some (currently internal) data structures built-up during analysis to operate.

`Rule` is very much dependent on `Visitor` and these diagnostics are not.

> As currently implemented, these rule IDs won't show up in sprocket explain and the new Config is going to be messy to maintain if we add a lot of new rules this way.

Agreed, that is definitely the downside of these four diagnostics not being a part of `Rule` and one that I think is worthy of additional work here.

I'm also not convinced `Config` is the way to go; I don't anticipate there will be much more than these four warnings, but if that proves not to be the case, it isn't the most maintainable design.

I _do_ like the idea of being able to specifically deny one of these warnings, which isn't possible with `Rule` or `sprocket` today.

Let me think about the above a little bit more and see how I can improve this as part of this PR.

a-frantz:
Is there a future where `wdl-grammar` or `wdl-ast` has named rules? We could "preemptively" move the `Rule` trait to either one of those. That would also involve removing the `Visitor` dependencies from `Rule`, which I think is fine. `wdl-lint` could have a `RuleExt` with the visitor implementations.

I think the reason we would name a rule is so that we can have a web page dedicated to explaining it. That could fit in `wdl-grammar` or `wdl-ast` IMO, even if those diagnostics should never be silenced.

peterhuene:
Spitballing:

* Have a trait `Rule` in `wdl-ast` that has everything from the existing `Rule`, minus the `Visitor` supertrait and `tags` method; that should be enough to describe rules for `sprocket` (for lint rule tags, we can probably do some downcasting trickery to print them in `sprocket explain`).
* Implement a `Rule` trait in `wdl-analysis` that has `wdl_ast::Rule` as a supertrait; I'm not yet sure what the trait would implement as the analysis warnings are implemented in very specific places.
* Implement a `Rule` trait in `wdl-lint` that has both `wdl_ast::Rule` and `wdl_ast::Visitor` as supertraits; it also has a `tags` method on it.

a-frantz:
> Spitballing:
> 
> * Have a trait `Rule` in `wdl-ast` that has everything from the existing `Rule`, minus the `Visitor` supertrait and `tags` method; that should be enough to describe rules for `sprocket` (for lint rule tags, we can probably do some downcasting trickery to print them in `sprocket explain`).
> * Implement a `Rule` trait in `wdl-analysis` that has `wdl_ast::Rule` as a supertrait; I'm not yet sure what the trait would implement as the analysis warnings are implemented in very specific places.
> * Implement a `Rule` trait in `wdl-lint` that has both `wdl_ast::Rule` and `wdl_ast::Visitor` as supertraits; it also has a `tags` method on it.
> * Implement a `rules` function in `wdl-analysis` that's similar to the one in `wdl-lint`; this would be used primarily by `sprocket explain`.

Sounds pretty good to me. I'd be ok with saving that for a follow-up PR if we want to prioritize getting this merged.

wdl-ast/src/lib.rs
```diff
@@ -96,6 +97,62 @@ impl<T: AstNode> AstNodeExt for T {
     }
 }
 
+/// An extension trait for syntax nodes.
+pub trait SyntaxNodeExt {
```
a-frantz, line 101: I think this is how I initially wrote it and you just followed my lead, but in hindsight I think this might be better suited as a `SyntaxElementExt`?

How things are set up in `wdl-format` is that comments are "owned" by other tokens (never nodes). So it seems strange to treat "except comments" here as being owned solely by nodes and not tokens. It might be more consistent to implement this extension trait for `NodeOrToken`.

But that doesn't need to happen in this PR. Just something to think about and maybe change in the future.

a-frantz (approved):

PR #219, opened 2024-10-16
fix: inline-comment detection logic

a-frantz:

fixes stjude-rust-labs/sprocket#297 

      changes (when appropriate).

### DIFF
--- a/wdl-lint/src/util.rs
+++ b/wdl-lint/src/util.rs
@@ -8,14 +8,18 @@ use wdl_ast::SyntaxKind;
 /// whitespace.
 pub fn is_inline_comment(token: &Comment) -> bool {
     if let Some(prior) = token.syntax().prev_sibling_or_token() {
-        let not_whitespace = prior.kind() != SyntaxKind::Whitespace;
+        let whitespace = prior.kind() == SyntaxKind::Whitespace;
+        if !whitespace {
+            return true;
+        }
+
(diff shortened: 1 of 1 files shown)


wdl-lint/src/util.rs
```diff
@@ -8,12 +8,14 @@ use wdl_ast::SyntaxKind;
...
-        return prior.kind() != SyntaxKind::Whitespace
-            || !prior
-                .as_token()
-                .expect("should be a token")
-                .text()
-                .contains('\n');
+        let not_whitespace = prior.kind() != SyntaxKind::Whitespace;
+        let contains_newline = prior
+            .as_token()
+            .expect("should be a token")
```
peterhuene, line 14: Previously this `expect` was being short-circuited when the syntax kind was not whitespace; I fear we may now be panicking if the previous sibling is a node.

a-frantz: Ah you're right. I wonder why that didn't cause any test/gauntlet issues?

peterhuene (approved):

wdl-lint/src/util.rs
peterhuene: We probably don't have any tests that place a comment directly after a non-whitespace token (probably should add one). I doubt we'd encounter it in the wild either (plus we only lint a select few repos).

PR #224, opened 2024-10-16
fix: don't panic on certain optional types

a-frantz:


peterhuene:
With a test for optional `Object?`, I think this looks good.

a-frantz:
I edited the CHANGELOGs and Cargo.tomls in this PR, which is technically against protocol, but we'd have to make another commit or PR just for those changes otherwise. IMO knocking them out together in a targeted PR makes more sense.

peterhuene (approved):

PR #225, opened 2024-10-17
feat: add a `wdl-engine` crate.

peterhuene:
This PR introduces a `wdl-engine` crate that will eventually be responsible for implementing a WDL execution engine.

Currently the crate only defines two important types: `Engine` and `Value`.

`Engine` will be used to evaluate WDL source files, but is currently limited to instantiating values.

`Value` represents a runtime WDL value; currently it implements the coercion rules and a `display` method to format the value.

The next step is for `Engine` to analyze a WDL document and proceed to evaluation if there are no diagnostics.

Once that is implemented, support for the WDL JSON format will be added as a way to pass input values into the evaluation.

      changes (when appropriate).

### DIFF
--- a/wdl-engine/src/value.rs
+++ b/wdl-engine/src/value.rs
@@ -171,9 +171,9 @@ impl Value {
         }
     }
 
-    /// Gets the value as a `File`.
+    /// Gets the value as a `Directory`.
     ///
-    /// Returns `None` if the value is not a `File`.
+    /// Returns `None` if the value is not a `Directory`.
     pub fn as_directory(self, engine: &Engine) -> Option<&str> {
         match self {
             Self::Directory(_) => self.as_str(engine),
--- a/wdl-engine/src/value.rs
+++ b/wdl-engine/src/value.rs
@@ -1,0 +1,1184 @@
+//! Implementation of the WDL runtime and values.
+
+use std::fmt;
+use std::sync::Arc;
+
+use id_arena::Id;
+use indexmap::IndexMap;
+use ordered_float::OrderedFloat;
+use string_interner::symbol::SymbolU32;
+use wdl_analysis::types::Coercible as _;
+use wdl_analysis::types::CompoundTypeDef;
+use wdl_analysis::types::CompoundTypeDefId;
+use wdl_analysis::types::Optional;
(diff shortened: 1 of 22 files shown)


wdl-engine/src/value.rs
```diff
@@ -0,0 +1,1184 @@
...
+    fn boolean_coercion() {
+        let mut engine = Engine::default();
+        let types = Types::default();
+
+        // Boolean -> Boolean
+        assert_eq!(
+            Value::from(false).coerce(&mut engine, &types, PrimitiveTypeKind::Boolean.into()),
+            Some(Value::from(false))
+        );
+        // Boolean -> String (invalid)
```
a-frantz, line 666: But this is allowed in placeholders. How is that handled?

wdl-engine/src/value.rs
```diff
@@ -0,0 +1,1184 @@
...
+        );
+        // String -> Boolean (invalid)
+        assert_eq!(
+            value.coerce(&mut engine, &types, PrimitiveTypeKind::Boolean.into()),
+            None
+        );
+    }
+
+    #[test]
+    fn string_display() {
```
a-frantz, line 795: Is there a way to display as a single quoted string? Don't exactly think there should be. I'd be fine only displaying as double-quoted.

wdl-engine/src/value.rs
peterhuene: Placeholder evaluation is separate from this and will be implemented as part of the expression evaluator; basically it'll evaluate the placeholder expression and, if it's a primitive type, specially coerce the type.

The `coerce` method here only captures the normal implicit type coercions.

a-frantz: Makes sense. Carry on then!

wdl-engine/src/value.rs
peterhuene: There isn't; displaying a value is mostly for debugging/development purposes as generally users will see the values coming as the outputs of a workflow or task and then probably write them out as JSON rather than display them.

a-frantz: That makes sense!

a-frantz (approved):

a-frantz:
o err... The CI just broke. Pending that fix, 👍

PR #230, opened 2024-10-21
ci: makes relevant CI tests multi-platform

claymcleod:
We should make sure `wdl` is always tested to work on all major platforms. To that end, I've updated the CI here to ensure that it always is. Some steps don't need to be run multi-platform, others clearly do, and then there are a few that are in-between (e.g., linting).

In both the "definitely needs to be run on all platforms" and the "in-between" ones, I ended up just enabling multi-platform tests to be safe. For example, for linting, some lints may only show up when certain platform-based configuration options are available. This is only going to become more relevant as we add more platform-specific things (like the encoding of newlines for UNIX/Windows).

      changes (when appropriate).

### DIFF
--- a/gauntlet/src/repository.rs
+++ b/gauntlet/src/repository.rs
@@ -145,6 +145,8 @@ impl Repository {
             .join(self.identifier.organization())
             .join(self.identifier.name());
 
+        std::fs::create_dir_all(&repo_root).expect("failed to create repository root directory");
+
         let git_repo = match git2::Repository::open(&repo_root) {
             Ok(repo) => {
                 info!("opening existing repository: {:?}", repo_root);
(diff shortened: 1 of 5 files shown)


peterhuene:
I ended up implementing some of this in #231 as I was trying to run tests on Windows during investigating the LSP issue.

I rolled the "test examples" into the "test" job, but haven't updated gauntlet/arena to run on non-linux platforms.

claymcleod:
@peterhuene are you good with me enabling this on gauntlet and arena? Or do you think we should just ignore that? My gut feeling is that it should be tested.

gauntlet/src/repository.rs
```diff
@@ -145,6 +145,8 @@ impl Repository {
             .join(self.identifier.organization())
             .join(self.identifier.name());
 
+        std::fs::create_dir_all(&repo_root).expect("failesd to create repository root directory");
```
peterhuene, line 148:
```suggestion
        std::fs::create_dir_all(&repo_root).expect("failed to create repository root directory");
```

peterhuene:
@claymcleod totally fine with it; certainly can't hurt.

PR #236, opened 2024-10-23
docs: Update RULES.md

a-frantz:

The `RULES.md` file got out of date. Reiterating something we all know, this isn't a sustainable documentation solution. Not sure the exact priority for updating `sprocket.bio` with pages for each rule, but that is the end-game here.

      changes (when appropriate).

### DIFF
--- a/wdl-lint/RULES.md
+++ b/wdl-lint/RULES.md
@@ -25,24 +25,26 @@ be out of sync with released packages.
 | `ImportSort`                     | Clarity, Style                | Ensures that imports are sorted lexicographically.                                                |
 | `ImportWhitespace`               | Clarity, Style, Spacing       | Ensures that there is no extraneous whitespace between or within imports.                         |
 | `InconsistentNewlines`           | Clarity, Style                | Ensures that newlines are used consistently within the file.                                      |
-| `InputNotSorted`                 | Style                         | Ensures that input declarations are sorted                                                        |
-| `InputSpacing`                   | Style, Clarity, Spacing       | Ensures proper spacing for call inputs                                                            |
+| `InputSorting`                   | Style                         | Ensures that input declarations are sorted                                                        |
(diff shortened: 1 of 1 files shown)


adthrasher (commented):
At this point, can we consider removing this file? Or at least making it auto-generated? I'd much rather see a CI step (or something) that loads the `wdl-lint` package, gets the rules, and writes them to a file if we want to have a textual version.

a-frantz:
> At this point, can we consider removing this file? Or at least making it auto-generated? I'd much rather see a CI step (or something) that loads the `wdl-lint` package, gets the rules, and writes them to a file if we want to have a textual version.

That seems doable. I don't have an exact vision for what it looks like though. Spitballing: It could be written into the `ci` crate as a command. Something like `cargo run --bin ci -- lint-docs`. It could recreate the `RULES.md` file pretty easily. IDK how that then gets integrated into `docs.rs` or otherwise displayed though.

peterhuene:
I agree that having this automated would be ideal and putting it in the `ci` crate makes sense to me.

I think we can keep the `RULES.md` file committed in git so that we generate docs like we do now, but have a CI job that also passes a `--check` flag to the `ci` command that will fail if the generated `RULES.md` doesn't match what's on disk.

a-frantz:
There should be a `BLESS` mechanism that will create the RULES.md file from scratch. That way we don't have to bother with writing it by hand or pretty-formatting the file.

a-frantz:
I'm leaning towards having it in `wdl-lint` and operating similar to our other tests, as opposed to the `ci` crate.

a-frantz:
Do we want to merge this PR or not? It probably doesn't make a difference, as I would hope this automation could be completed in time for the next release (whenever that is).

peterhuene (approved):
Might as well merge it and ensure that whatever automation we create ends up basically not changing what we think is correct.

PR #240, opened 2024-10-29
fix: formatting of command sections

a-frantz:

Two big improvements to formatting of `command-text`.
1. Multi-line placeholders now have an appropriate amount of "temporary indentation" added so they are aligned with the indentation of the open delim.
2. Leading whitespace of command sections is stripped and then normalized to the appropriate level of indentation.

      changes (when appropriate).


wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -3336,15 +3328,14 @@ task call_peak {
...
         fi
-
         python3 $(which encode_task_post_call_peak_chip.py) \
             $(ls *Peak.gz) \
             ~{"--ta " + tas[0]} \
             ~{"--regex-bfilt-peak-chr-name '" + regex_bfilt_peak_chr_name + "'"} \
             ~{"--chrsz " + chrsz} \
             ~{"--fraglen " + fraglen} \
             ~{"--peak-type " + peak_type} \
-            ~{"--blacklist " + blacklist}
```
a-frantz, line 3347: I don't know why this trailing space was there, but it's gone now 🤷

wdl-format/tests/format/clays_complex_script/source.formatted.wdl
```diff
@@ -70,7 +70,8 @@ task a_task {
     # the TextMate language that it's a custom struct.
     }
 
-    command <<< >>>
+    command <<<
```
claymcleod, line 73: Is this desired? I think, for a totally blank command, I'd prefer `command <<<>>>`. But that is a special case, and I'm also not set on that if you all have made a decision otherwise (please move forward in that case).

a-frantz: @adthrasher has made similar comments about empty sections/tasks/workflows which all currently behave the same (a newline is always inserted after the open delim, even when the body is empty).

IMO these are corner cases not worth the added complexity to address. Granted, that added complexity isn't very great, a few lines in maybe half a dozen places. But that's still extra complexity for an edge case that I just don't think matters.

If folks really want to see empty bodies special cased, I'll add it. But my thinking is we leave as is.

wdl-format/tests/format/funky_command_leading_whitespace/source.formatted.wdl
```diff
@@ -0,0 +1,28 @@
...
+task long_leading_whitespace {
+    command <<<
+        echo "hello"
+    >>>
+}
+
+task mixed_leading_whitespace {
+    command <<<
+        echo "hello"
+              echo "world"
```
adthrasher, line 26: This makes sense based on what we've discussed previously, but I don't particularly like the result.

wdl-format/tests/format/clays_complex_script/source.formatted.wdl
adthrasher: I think that's fine, @a-frantz. The additional complication would be, here, this command block isn't truly empty, it contains a single space. Should that be deleted and treated as empty? That would differ from our approach elsewhere.

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
adthrasher, line 1: I wish there was a way in the GitHub diff to set the non-formatted as the original, so you could see the true changes that format made, rather than just the changes from the last commit.

adthrasher (approved):

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
a-frantz: That would be nice. Probably the best you can do is to locally checkout the branch and run `diff source.*wdl`, which wouldn't be as nice

wdl-format/tests/format/funky_command_leading_whitespace/source.formatted.wdl
a-frantz: How would you ideally format this? I think the only way to improve it is to actually parse the Bash to determine that extra whitespace is meaningless. And I've gone on record as saying parsing Bash is a barrel of worms we shouldn't open.

PR #241, opened 2024-10-30
feat: implement JSON inputs files in `wdl-engine`.

peterhuene:
This PR implements support for JSON input files in `wdl-engine`.

The file format conforms to the WDL specification for JSON input files.

Value coercion now returns errors for so users can better diagnose coercion
problems.

Also adds stub methods to `Engine` for evaluating workflows and tasks; the
implementation is forthcoming.

      changes (when appropriate).


wdl-engine/tests/inputs/invalid-runtime-type/error.txt
```diff
@@ -0,0 +1,5 @@
+failed to parse input file `tests/inputs/invalid-runtime-type/inputs.json`
+
+Caused by:
+    0: invalid input key `test.runtime.returnCodes`
+    1: expected type `Int`, type `String`, or type `Array[Int]` for runtime key `returnCodes`, but found type `None`
```
claymcleod, line 5: Do we want this to say `null` since that's what the value is in the JSON?

claymcleod (approved):
Looks good to me!

wdl-engine/src/inputs.rs
```diff
@@ -0,0 +1,756 @@
...
+
+/// Represents inputs to a task.
+#[derive(Default, Debug, Clone)]
+pub struct TaskInputs {
+    /// The task input values.
+    inputs: HashMap<String, Value>,
+    /// The overridden requirements section values.
+    requirements: HashMap<String, Value>,
+    /// The overridden hints section values.
+    hints: HashMap<String, Value>,
```
a-frantz, line 36: I didn't realize you could override hints. What's the use case for that? Spitballing: I guess it makes sense if an implementation leaves _out_ a `hints` section and you want to "fill it in" at runtime. But overriding specified values? That seems wrong to me, but maybe I'm just not thinking of the right use case.

wdl-engine/src/inputs.rs
```diff
@@ -0,0 +1,756 @@
...
+        }
+    }
+}
+
+/// Represents inputs to a WDL workflow or task.
+#[derive(Debug, Clone)]
+pub enum Inputs {
+    /// The inputs are to a task.
+    Task(TaskInputs),
+    /// The inputs are
```
a-frantz, line 524:
```suggestion
    /// The inputs are to a workflow.
```

wdl-analysis/src/types/v1.rs
```diff
@@ -97,6 +98,195 @@ pub fn task_member_type(name: &str) -> Option<Type> {
...
+/// if the given name is not a requirement.
+pub fn task_requirement_types(
+    version: SupportedVersion,
+    name: &str,
+) -> Option<(&'static [Type], &'static str)> {
+    /// The types for the `container` requirement.
+    static CONTAINER_TYPES: LazyLock<(Box<[Type]>, &'static str)> = LazyLock::new(|| {
+        (
+            Box::new([PrimitiveTypeKind::String.into(), STDLIB.array_string]),
+            "type `String` or type `Array[String]`",
```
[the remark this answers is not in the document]
peterhuene: That's easy enough to do, I'll add it.

wdl-engine/src/inputs.rs
peterhuene: Yeah, both requirements and hints [can be overridden](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#specifying--overriding-requirements-and-hints) and an engine _must_ support doing so.

Probably useful for when you might have a bespoke container image you want to run someone else's task in or the environment you'll be running a task on would benefit from higher minimum/maximum cpu/memory/storage.

wdl-engine/tests/inputs/invalid-hints-type/source.wdl
```diff
@@ -0,0 +1,8 @@
+## This is a test of an unknown task hint.
+## As an unknown hint are not an error, `error.txt` is expected to be empty.
```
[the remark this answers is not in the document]
peterhuene: Indeed, bad comment. I'll remove.

wdl-engine/tests/inputs/invalid-runtime-type/error.txt
peterhuene: I think it's best to consistently use WDL types in the error messages, as the expected types are always expressed as WDL types.

The [list of supported primitive types](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#primitive-types-1) is pretty small and I don't think folks will get tripped up between WDL's `None` and JSON's `null`, to be honest.

wdl-engine/src/inputs.rs
a-frantz: We've written `workflows` in a way that a user should never have to override a runtime/requirements value (or at least it should be rare), but that use case makes perfect sense to me. A task is failing due to too low a mem allocation? Override `runtime.memory`.

But `hints`??? Those don't seem like something that should be "messed with" by the user. But I haven't actually written any hints sections for production, so maybe I'm just thinking about it wrong.

@adthrasher just curious if you have thoughts on this. Seems an odd case to me.

a-frantz (approved):
LGTM 👍

PR #242, opened 2024-10-30
feat(format): respect max line width

a-frantz:

Took me a while to crack this problem, but with some help from @adthrasher I think we've finally got the feature working!

Still to do in a follow up PR: we want certain line breaks to be "matched", like if we break on one half of a parenthesis pair we should break on the other half as well. Until that gets implemented, some of the line breaks will look a little awkward.

      changes (when appropriate).


wdl-format/src/config/indent.rs
```diff
@@ -1,22 +1,50 @@
...
+/// The maximum number of spaces to represent one indentation level.
+pub const MAX_SPACE_INDENT: usize = 16;
 
 /// An indentation level.
 #[derive(Clone, Copy, Debug)]
 pub enum Indent {
     /// Tabs.
-    Tabs(NonZeroUsize),
-
+    Tabs,
```
adthrasher, line 15: So if a user indents with `tab`, they can only use one tab. I'm fine with that, but just flagging that to make sure we're OK with that choice.

wdl-format/src/config/max_line_length.rs
```diff
@@ -0,0 +1,52 @@
...
+        }
+    }
+}
+
+/// The default maximum line length.
+pub const DEFAULT_MAX_LINE_LENGTH: usize = 90;
+/// The minimum maximum line length.
+pub const MIN_MAX_LINE_LENGTH: usize = 60;
+/// The maximum maximum line length.
+pub const MAX_MAX_LINE_LENGTH: usize = 240;
```
adthrasher, line 25: Have you tested any non-default line lengths?

wdl-format/src/token.rs
```diff
@@ -52,6 +52,31 @@ impl<T: Token> TokenStream<T> {
...
+
+    /// Clears the stream.
+    pub fn clear(&mut self) {
+        self.0.clear();
+    }
+
+    /// Drains the stream.
+    pub fn drain(&mut self) -> std::vec::Drain<'_, T> {
+        self.0.drain(..)
+    }
```
adthrasher, line 74: Am I correct that this is `clear`, but it returns the items from the stream?

wdl-format/src/token.rs
```diff
@@ -64,27 +89,32 @@ impl<T: Token> IntoIterator for TokenStream<T> {
...
     /// A comment.
     Comment(Comment),
 }
 
-/// Whether optional blank lines are allowed in the current context.
+/// The policy for line spacing.
+///
+/// Blank lines before comments and between comments are always permitted.
 #[derive(Eq, PartialEq, Default, Debug, Clone, Copy)]
 pub enum LineSpacingPolicy {
```
adthrasher, line 113: Maybe I just haven't hit where this is used, but the comments suggest this applies only to comments. Should the naming reflect that?

wdl-format/src/token/post.rs
```diff
@@ -29,8 +34,14 @@ pub enum PostToken {
     /// One indentation.
     Indent,
 
+    /// A temporary indent.
+    ///
+    /// This is added after a [`PostToken::Indent`] during the formatting of
+    /// command sections.
```
adthrasher, line 40: It's not clear to me from these comments why this is needed. Maybe there is additional documentation below where the command block is formatted, but it may be useful to expand here.

adthrasher (commented):
Partial review before the interview starts

wdl-format/src/token.rs
a-frantz: Exactly. I was using this in a WIP implementation, but don't use it now. We can remove if we want.

wdl-format/src/config/builder.rs
```diff
@@ -30,37 +39,62 @@ pub type Result<T> = std::result::Result<T, Error>;
...
     indent: Option<Indent>,
+    /// The maximum line length.
+    max_line_length: Option<MaxLineLength>,
 }
 
 impl Builder {
     /// Creates a new builder with default values.
-    pub fn new() -> Self {
-        Default::default()
+    pub fn new(indent: Option<Indent>, max_line_length: Option<MaxLineLength>) -> Result<Self> {
```
peterhuene, line 48: I'm not use to seeing parameters to a builder constructor.

Perhaps we could leave this as `new()` and perform the `unwrap_or_default` logic in `try_build`?

It also seems like we have good defaults for both `ident` and `max_line_length`, so I don't see how `try_build` would need to be fallible (so simply change to `build`?).

wdl-format/src/token.rs
a-frantz: Yes it should probably have a name change.

This has no effect on `PreToken::BlankLine`s which are always rendered (up to a maximum of one in a row; `[PreToken::BlankLine, PreToken::BlankLine]` would get rendered as a single blank in the `PostToken` stream). `LineSpacingPolicy` is only checked when rendering `Trivia::BlankLine`s.

```suggestion
/// The policy for [`Trivia::BlankLine`] line spacing.
///
/// Blank lines before comments and between comments are always permitted.
#[derive(Eq, PartialEq, Default, Debug, Clone, Copy)]
pub enum TriviaBlankLineSpacingPolicy {
```
maybe?

wdl-format/src/config/max_line_length.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// The maximum line length.
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct MaxLineLength(Option<usize>);
+
+impl MaxLineLength {
+    /// Creates a new `MaxLineLength` with the provided value.
+    pub fn with_value(value: usize) -> Result<Self, Error> {
```
peterhuene, line 33: I think this could probably just be `new` as there is no other constructor.

wdl-format/src/config/max_line_length.rs
```diff
@@ -0,0 +1,52 @@
...
+        let val = match value {
+            0 => Self(None),
+            MIN_MAX_LINE_LENGTH..=MAX_MAX_LINE_LENGTH => Self(Some(value)),
+            _ => return Err(Error::Invalid("max_line_length")),
+        };
+        Ok(val)
+    }
+
+    /// Gets the maximum line length.
```
peterhuene, line 42: Could we include what a `None` value signifies in this comment?

wdl-format/src/lib.rs
```diff
@@ -209,15 +212,13 @@ impl Formatter {
 
     /// Gets the [`PostToken`] stream.
     ///
-    /// # Notes
-    ///
-    /// * This shouldn't be exposed publicly.
+    /// This shouldn't be exposed publicly.
```
peterhuene, line 215: I feel like this existing comment is redundant as the method isn't `pub`?

wdl-format/src/config/builder.rs
```diff
@@ -30,37 +39,62 @@ pub type Result<T> = std::result::Result<T, Error>;
...
     ///
-    /// # Notes
-    ///
     /// This silently overwrites any previously provided value for the
     /// indentation level.
-    pub fn indent(mut self, indent: Indent) -> Self {
+    pub fn indent(mut self, indent: Indent) -> Result<Self> {
+        if let Indent::Spaces(n) = indent {
+            if n > MAX_SPACE_INDENT {
+                return Err(Error::Invalid("indent"));
```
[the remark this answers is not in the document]
claymcleod: I'd also want a more descriptive error message so the user knows what to fix.

wdl-format/src/config/builder.rs
```diff
@@ -30,37 +39,62 @@ pub type Result<T> = std::result::Result<T, Error>;
...
+        }
         self.indent = Some(indent);
+        Ok(self)
+    }
+
+    /// Sets the maximum line length.
+    ///
+    /// This silently overwrites any previously provided value for the maximum
+    /// line length.
+    pub fn max_line_length(mut self, max_line_length: MaxLineLength) -> Self {
```
claymcleod, line 74: This should also check the max line length after you move the checks to the `new()` method.

wdl-format/src/token/post.rs
```diff
@@ -258,21 +375,131 @@ impl Postprocessor {
...
+        }
+
+        // At least one line in the post_buffer is too long.
+        // We iterate through the in_stream to find potential line breaks,
+        // and then we iterate through the in_stream again to actually insert
+        // them in the proper places.
+
+        let max_length = config.max_line_length().unwrap();
+
+        let mut potential_line_breaks: Vec<usize> = Vec::new();
```
peterhuene, line 411: Could this just start off as a `HashSet` rather than `Vec` that gets remade into a `HashSet`?

peterhuene (commented):
Looks really good and I like the formatting changes, especially for `if` expressions.

Just a few comments.

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -1511,38 +1380,26 @@ workflow chip {
...
         #     paired_end will override paired_ends[i]
-        Boolean paired_end_ = (
-            if !defined(paired_end) && i < length(paired_ends)
-            then paired_ends[i]
-            else select_first([
-                paired_end,
-            ])
-        )
+        Boolean paired_end_ = if !defined(paired_end) && i < length(paired_ends) then paired_ends[
+            i] else select_first([
```
claymcleod, line 1384: I'm on the fence whether I like this idea that `(` and `[` can split things down to the next line. For instance, here, I think it looks quite strange. I'm thinking that, for these lines that are being broken based on long line length, perhaps that should not be permitted.

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -1627,7 +1472,8 @@ workflow chip {
             }
         }
 
-        Boolean has_input_of_count_signal_track = has_output_of_bam2ta || defined(bam2ta.ta)
+        Boolean has_input_of_count_signal_track = has_output_of_bam2ta || defined(bam2ta.ta
+            )
```
claymcleod, line 1476: Similarly here, this is quite strange from a formatting perspective.

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -1783,50 +1616,35 @@ workflow chip {
...
-
-        Boolean has_input_of_align_ctl = i < length(ctl_fastqs_R1) && length(ctl_fastqs_R1[i]) > 0
+        Boolean ctl_paired_end_ = if !defined(ctl_paired_end) && i < length(
+            ctl_paired_ends) then ctl_paired_ends[i] else select_first([
+            ctl_paired_end,
+            paired_end,
+        ])
+
+        Boolean has_input_of_align_ctl = i < length(ctl_fastqs_R1) && length(ctl_fastqs_R1[
+            i]) > 0
```
claymcleod, line 1633: And here.

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -1841,13 +1659,10 @@ workflow chip {
...
-        File? ctl_bam_ = (
-            if has_output_of_align_ctl
-            then ctl_bams[i]
-            else align_ctl.bam
-        )
+        File? ctl_bam_ = if has_output_of_align_ctl then ctl_bams[i] else align_ctl.bam
 
-        Boolean has_input_of_filter_ctl = has_output_of_align_ctl || defined(align_ctl.bam)
+        Boolean has_input_of_filter_ctl = has_output_of_align_ctl || defined(align_ctl.bam
+            )
```
claymcleod, line 1665: And here. Sorry for so many comments, just want to call out places that require further examination.

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
a-frantz: I'm planning for a follow up PR that will line break "matching" tokens. So this example might become:
```
        Boolean paired_end_ = if !defined(paired_end) && i < length(paired_ends) then paired_ends[
            i
        ] else select_first([
```
or something similar. I haven't spent time thinking about the specifics yet, but I agree that the current implementation leaves something to be desired. I can push up a change the removes the open/close tokens from `can_be_line_broken` so we can see how it looks?

wdl-format/src/config/max_line_length.rs
a-frantz: Check out these 3 commits on `workflows` - https://github.com/stjudecloud/workflows/compare/main...tmp/format-examples

wdl-format/src/config/builder.rs
```diff
@@ -27,40 +35,40 @@ impl std::error::Error for Error {}
...
-    }
-
     /// Sets the indentation level.
     ///
-    /// # Notes
-    ///
     /// This silently overwrites any previously provided value for the
     /// indentation level.
-    pub fn indent(mut self, indent: Indent) -> Self {
+    pub fn indent(mut self, indent: Indent) -> Result<Self> {
```
peterhuene, line 51: We can make this infallible now.

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -3344,7 +2954,7 @@ task call_peak {
             ~{"--chrsz " + chrsz} \
             ~{"--fraglen " + fraglen} \
             ~{"--peak-type " + peak_type} \
-            ~{"--blacklist " + blacklist}
+            ~{"--blacklist " + blacklist}
```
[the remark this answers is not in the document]
a-frantz: It's present in the unformatted file. I'm not sure why the prior implementation was stripping it 🤔 but I think it's supposed to be there.

wdl-format/tests/format/ENCODE-DCC_chip-seq-pipeline/source.formatted.wdl
```diff
@@ -2235,25 +1990,19 @@ workflow chip {
...
-        ]
-    )
+    Array[File?] chosen_ctl_ta_pooled = if !has_all_input_of_choose_ctl || align_only_
+        then [] else if chosen_ctl_ta_pooled_subsample > 0 then [
+        subsample_ctl_pooled.ta_subsampled,
+    ] else if num_ctl < 2 then [
+        ctl_ta_[0],
+    ] else [
+        pool_ta_ctl.ta_pooled,
+    ]
```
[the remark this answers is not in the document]
a-frantz: Here's the unformatted source:
```wdl
    Array[File?] chosen_ctl_ta_pooled = if !has_all_input_of_choose_ctl || align_only_ then []
        else if chosen_ctl_ta_pooled_subsample > 0 then [ subsample_ctl_pooled.ta_subsampled ]
        else if num_ctl < 2 then [ ctl_ta_[0] ]
        else [ pool_ta_ctl.ta_pooled ]
```

peterhuene (approved):
Approved, but we should probably create a tracking issue for any follow-up tweaking needed to the line breaking output.

PR #244, opened 2024-10-31
Feat: redundant assignment rule

michaelgattas:

- **Rule Name**: `RedundantInputAssignment`
This rule checks if call statement's input assignment's name and expression can be reduced, like: 
```
call foo { input: {
        String x = x
    }
}
```
where `String x = x` can simply be written `x`. This rule will only be applied for versions 1.1 and greater.

Rule specific checks:

      is implemented.
~~- [ ] If you have implemented a new `Visitor` callback, you have also
      visitor callback will execute.~~
      (`Arena.toml`).


wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,90 @@
...
+    fn version_statement(&mut self, state: &mut Self::State, reason: VisitReason, stmt: &VersionStatement) {
+        // this current takes into account for any possible patch versions, which might be unnecessary
+        let binding = stmt.version();
+        let split_versions = binding.as_str().split(".").collect::<Vec<_>>();
+        if split_versions[0] == "1" && split_versions[1] == "0" {
+            self.included_version = false;
+        } else {
+            self.included_version = true;
+        }
+    }
```
michaelgattas, line 73: @a-frantz I wrote this function to account for patch versions, but if it isn't necessary, I can simply do a check for == "1.0"? I'm not the most familiar with current WDL SemVer

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    fn input_section(&mut self, state: &mut Self::State, reason: VisitReason, section: &InputSection) {
+        /// if the version statement is earlier than 1.1 return
+        if !self.included_version {
+            return
+        }
```
michaelgattas, line 79: @a-frantz is this the recommended way to access the state of the Visitor with regards to its version statement? Is there a guarantee that `version_statement` function is called before `input_section`?

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// for each declaration in the input section,
+        section.declarations().for_each(|dcl| {
+            if let Some(expr) = dcl.expr() {
+                // if dcl.name().as_str() == expr {
+                //     state.add(redundant_input_assignment(Span::new(0, 12)));
+                // }
```
michaelgattas, line 86: @a-frantz this is where I'm hitting my confusion. I'm not sure the best way to turn the `expr` into a string for direct comparison, and have been hitting some difficulties understanding how to access the span for each line of `input_section`

wdl-lint/src/rules/redundant_input_assignment.rs
a-frantz: WDL version statements do not include a patch number (but there are patch releases, which can be confusing. The idea is that patch releases are mostly fixing typos and making clarifications in the text, which don't impact parsing or semantics, so the patch can be ignored in this context). A WDL version statement will always look like `version 1.2` or `version 1.0`, where the version is just `<major>.<minor>`.

michaelgattas: Are there any versions to consider prior to "1.0"? If there aren't zero versions, I think this function can be less verbose

wdl-lint/src/rules/redundant_input_assignment.rs
a-frantz: There's a better way to do the version check. You're going about it rather manual, but we've written an API layer for doing it that will be more robust and readable. Will link the docs in a moment.

> Is there a guarantee that version_statement function is called before input_section?

Yes there is! The version statement _must_ be the first non-trivia item in a document, so you can rely on the guarantee that the visitor will see the version statement before the visitor gets to an input section.

wdl-lint/src/rules/redundant_input_assignment.rs
a-frantz: There are "draft" releases before 1.0, but we as a team have explicitly decided we will not support them. So you can assume any version will be `>=v1.0`. If someone tries to run our tool on a draft version, they will be hit with an error long before reaching your part of the code.

wdl-lint/src/rules/redundant_input_assignment.rs
a-frantz: You can use this API to do facilitate the version check for you: https://docs.rs/wdl/latest/wdl/ast/version/enum.V1.html

i.e. do not try to parse the string of the version, we've already written that logic with all the proper checks and what not.

wdl-lint/src/rules/redundant_input_assignment.rs
a-frantz: I think you can get this done without ever calling `as_str()`. A few scattered thoughts here:
- `as_str()` is actually quite expensive to call _for nodes_ (and `Expr` is a node)
    - you should try to only call `as_str()` once you've parsed down to a token, which is cheap
    - It's expensive for nodes because it involves traversing the CST to visit every child node and calling `as_str()` on them
- `if let Some(expr) = dcl.expr()` this is a good start. In the case that `.expr()` returns none, that means the declaration is already using the abbreviated syntax you're checking for. So this part looks good 👍 
- Next you want to check what kind of expression is present. You only care about when it is a `NameRef`
    - https://docs.rs/wdl/latest/wdl/ast/v1/enum.Expr.html#variant.Name
    - So check if it can be ceorced to a `NameRef`, and if so check if the name is equivalent to the `dcl.name()`
    - I could be wrong, but I think that check can be done be comparing the tokens to each other (without converting to str)

a-frantz: O and for the span: I think you want to highlight the entire declaration, so you can do something like `dcl.to_span()` instead of trying to construct a new span yourself.

wdl-lint/tests/lints/redundant_input_assignment/source.wdl
```diff
@@ -0,0 +1,28 @@
...
+        a: ""
+        b: 2
+        c: ""
+    }
+
+    input {
+        int b
+        String a = a
+        String c = c
+    }
```
a-frantz, line 20: You are testing the wrong `input `section here. Understandable confusion if you aren't familiar with WDL.
```suggestion
    input {
        int b
        String a
        String c
    }
```
What we're looking at here is the input section to a task (could also be in a workflow). What you're trying to test is the input block of _a workflow call_.
So add something like
```
workflow test {
    input {
        Int b
        String a
        String c
    }
    call bar { input:
         a,  # should not flag
         b = b + 3,  # should not flag
         c = c,  # should flag
   }
}

File: wdl-lint/src/rules/redundant_input_assignment.rs (hunk 2 above)
Comment (a-frantz) [reply wdl-lint/src/rules/redundant_input_assignment.rs:79]: Check out how this rule uses the Version API:
https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/rules/missing_requirements.rs#L75

File: wdl-lint/src/rules/redundant_input_assignment.rs (hunk 3 above)
Comment (a-frantz) [reply wdl-lint/src/rules/redundant_input_assignment.rs:86]: Er. Actually. In light of #243 I would just highlight the `Expr`/`NameRef`

File: wdl-lint/src/rules/redundant_input_assignment.rs [hunk 5]
```diff
@@ -0,0 +1,90 @@
...
+        let binding = stmt.version();
+        let split_versions = binding.as_str().split(".").collect::<Vec<_>>();
+        if split_versions[0] == "1" && split_versions[1] == "0" {
+            self.included_version = false;
+        } else {
+            self.included_version = true;
+        }
+    }
+
+    fn input_section(&mut self, state: &mut Self::State, reason: VisitReason, section: &InputSection) {
```
a-frantz, line 75: Oops. I should have caught this earlier. This is the wrong visitor. You want to be visiting each `call_statement`, not `input_section`.

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,102 @@
...
+        reason: VisitReason,
+        _: &Document,
+        version: SupportedVersion,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        // Reset the visitor upon document entry
+        *self = Self(Some(version));
```
michaelgattas, line 74: @a-frantz I tried mirroring the logic used in `missing_requirements.rs` but am getting an error from the call to `Self`, and I'm not sure why. The error the compiler gives me is `the `Self` constructor can only be used with tuple or unit structs`

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,102 @@
...
+        &mut self,
+        state: &mut Self::State,
+        reason: VisitReason,
+        section: &InputSection,
+    ) {
+        /// if the version statement is earlier than 1.1 return
+        if let SupportedVersion::V1(minor_version) = self.0.expect("version should exist here") {
+            if minor_version < V1::One {
+                return;
+            }
```
michaelgattas, line 87: @a-frantz for some reason, this mirrored logic from `missing_requirements.rs` is giving me the compiler error: `failed to resolve: `V1` is a variant, not a module`, and I'm not sure why. Do you know how I can achieve this version comparison?

wdl-lint/src/rules/redundant_input_assignment.rs
michaelgattas: I resolved this! I forgot to incorporate the Option within the struct declaration

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,103 @@
...
+        "Input assignments that are redundant can be shortened. For example, `{ input: a = a }` \
+         can be shortened to `{ input: a }`."
+    }
+
+    fn tags(&self) -> TagSet {
+        TagSet::new(&[Tag::Clarity, Tag::Correctness])
+    }
+
+    fn exceptable_nodes(&self) -> Option<&'static [wdl_ast::SyntaxKind]> {
+        None
```
a-frantz, line 53:
```suggestion
        Some(&[
            wdl_ast::SyntaxKind::VersionStatementNode,
            wdl_ast::SyntaxKind::WorkflowDefinitionNode,
            wdl_ast::SyntaxKind::CallStatementNode,
        ])
```
Your last rule was a special case and you didn't need to deal with exceptions, but this one needs to be "exceptable".

I would take a quick glance over my PR that introduced the current exception infrastructure if you are unclear how it works (#135).

Make sure you're test files check that `#@ except: RedundantInputAssignment` comments work as expected.

a-frantz (commented):
Looks like you're headed in the right direction 🚀

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    fn call_statement(
+        &mut self,
+        state: &mut Self::State,
+        reason: VisitReason,
+        stmt: &CallStatement,
+    ) {
+        if let SupportedVersion::V1(minor_version) = self.0.expect("version should exist here") {
+            if minor_version < V1::One {
```
michaelgattas, line 86: @a-frantz  This line is giving me a problem, where `V1::One` is erroring, returning: `failed to resolve: `V1` is a variant, not a module`. I tried to mirror the `missing_requirements.rs` implementation so am confused (what's new!)

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,107 @@
...
+            stmt.sections()
+                .filter_map(|section| section.as_input())
+                .for_each(|section| {
+                    section
+                        .declarations()
+                        .for_each(|dcl| {
+                            if let Some(expr) = dcl.expr() {
+                                if let expr_name = NameRef(expr.clone()) {
+                                    if expr_name == dcl.name() {
+                                        state.push(redundant_input_assignment(expr_name.to_span()));
```
michaelgattas, line 99: @a-frantz  Any tips for how to get the `Expr` or `NameRef(Expr`) span for the sake of highlighting in the diagnostic?

wdl-lint/src/rules/redundant_input_assignment.rs
a-frantz:
```suggestion
            if minor_version < wdl_ast::version::V1::One {
```
And then remove from the top of the file `use wdl_ast::Ast::V1;`. You imported the wrong thing (although they are named the same, so that's confusing 😬 )

wdl-lint/RULES.md
```diff
@@ -41,6 +41,7 @@ be out of sync with released packages.
 | `PreambleCommentAfterVersion`    | Clarity                       | Ensures that documents have correct comments in the preamble.                                     |
 | `PreambleFormatting`             | Spacing, Style, Clarity       | Ensures that documents have correct whitespace in the preamble.                                   |
 | `RuntimeSectionKeys`             | Completeness, Deprecated      | Ensures that runtime sections have the appropriate keys.                                          |
+| `RedundantInputAssignment`       | Clarity, Correctness          | Ensures that redundant input assignments are shortened                                            |
```
michaelgattas, line 44: @a-frantz just made my best guess, but hoping these tags are correct

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    fn document(
+        &mut self,
+        _: &mut Self::State,
+        reason: VisitReason,
+        _: &Document,
+        version: SupportedVersion,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
```
michaelgattas, line 77: @a-frantz I haven't implemented code that considers the two-visit model yet, so wasn't sure to opt for Exit or Enter. Hopefully made the right call

wdl-lint/RULES.md
a-frantz: I don't think `Correctness` applies. I do think `Style` should be added. I'm a little on the fence about `Clarity`. Leaning towards removing that as well, and leaving it `Style` by itself. If anyone else reviews this, I'll let them weigh in.

wdl-lint/src/rules/redundant_input_assignment.rs
a-frantz: This is correct.

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,113 @@
...
+        *self = Self(Some(version));
+    }
+
+    fn call_statement(
+        &mut self,
+        state: &mut Self::State,
+        reason: VisitReason,
+        stmt: &CallStatement,
+    ) {
+        if reason == VisitReason::Exit {
```
a-frantz, line 90: This one could have been either `Exit` or `Enter` and I don't think it would have made a meaningful difference. In the future, general guideline would be to prefer doing business logic on `Enter` rather than `Exit`. So this is good here, just pointing out for the future 👍

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,113 @@
...
+
+/// Create a "Redundant Input Assignment" diagnostic.
+fn redundant_input_assignment(span: Span, name: &str) -> Diagnostic {
+    Diagnostic::note("redundant input assignment can be shortened")
+        .with_rule(ID)
+        .with_label(
+            format!("redundant input assignment can be shortened to `{}`", name),
+            span,
+        )
+}
```
a-frantz, line 33:
```suggestion
fn redundant_input_assignment(span: Span, name: &str) -> Diagnostic {
    Diagnostic::note("redundant input assignment")
        .with_rule(ID)
        .with_highlight(
            span,
        ).with_fix(
            format!("can be shortened to `{}`", name),
        )
}
```
I think this is a place where it makes sense to use a `fix`, despite that open issue saying we should use those more sparingly (#193)

a-frantz: I just typed this in GH it might not compile lol. I think what I was going for is clear though

Arena.toml
a-frantz, line 1: Please go through each of these additions looking for any false positives. Lmk when you've done that and they all look good to you.

a-frantz (commented):
Looking pretty great. Good job so far! Close to the finish line with this one.

wdl-lint/src/rules/redundant_input_assignment.rs
michaelgattas: Missed the exact language at first, but just re-committed with this exact verbiage

Arena.toml
a-frantz: Unfortunately I do think that's currently the most efficient way to check changes to gauntlet/arena.

I have plans for changes to gauntlet (#226) but they do not include improving this part of the experience. However if you have thoughts for improvements, feel free to pitch them in an issue!

michaelgattas: No worries at all, just wanted to make sure I didn't miss anything! In that case, I think I've incorporated all your feedback thus far. Let me know if there's anything outstanding

Gauntlet.toml
a-frantz, line 1: Going to ask you to ignore the stated instructions which you followed lol. Can you please revert all changes to this file? Gauntlet should still pass after you do that (just don't run it with `--refresh`).

Reasoning is explained by #226.

a-frantz (commented):
Going to take another close look in the morning, but aside from the gauntlet changes didn't see anything glaring. Might have some small nits, but I expect us to merge this tomorrow 🎉

a-frantz:
Ah one more small thing. Please update the PR description, the current explanation of the changes is a touch off. Current length/level of detail is good, just a little inaccurate. You can just edit the comment in place.

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,111 @@
...
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Flags redundant input assignments."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "Input assignments that are redundant can be shortened. For example, `{ input: a = a }` \
+         can be shortened to `{ input: a }`."
```
a-frantz, line 48:
```suggestion
        "Redundant input assignments can be shortened in WDL versions >=v1.1. For example, `{ input: a = a }` can be shortened to `{ input: a }`."
```
This will need to be run through `cargo fmt` again after committing the wording change 👍

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    fn tags(&self) -> TagSet {
+        TagSet::new(&[Tag::Style])
+    }
+
+    fn exceptable_nodes(&self) -> Option<&'static [wdl_ast::SyntaxKind]> {
+        Some(&[
+            wdl_ast::SyntaxKind::VersionStatementNode,
+            wdl_ast::SyntaxKind::WorkflowDefinitionNode,
+            wdl_ast::SyntaxKind::CallStatementNode,
```
a-frantz, line 59:
```suggestion
            wdl_ast::SyntaxKind::CallStatementNode,
            wdl_ast::SyntaxKind::CallInputItemNode,
```
A user should be able to except a single input in addition to every input for a call. Please update one of the tests to have something like:
```
#@ except: RedundantInputAssignment
call foo { input:
    baz = baz  # should not flag
}
call foo as something_else { input:
     #@ except: RedundantInputAssignment
    baz = baz,  # should not flag
   kazam = kazam,  # should flag
}
```

wdl-lint/tests/lints/blank-lines-between-elements/source.errors
a-frantz, line 1: Please go into `wdl-lint/tests/lints/blank-lines-between-elements/source.wdl` and change all these `s = s` statements into just `s`. We don't want these diagnostics "muddying" the error messages being checked in this test.

So after you make that change, this `source.errors` file should _not_ have any changes to it in your PR.

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.0/source.wdl
```diff
@@ -0,0 +1,19 @@
...
+        String a
+        String c
+        Int b
+    }
+
+    # This should not flag any notes, since version is 1.0
+    call bar { input:
+         a,  # should not flag
+         b = b + 3,  # should not flag
+         c = c,  # should flag
```
a-frantz, line 17:
```suggestion
         a.
         b = b + 3,
         c = c,
```
We shouldn't have comments that "disagree" with each other, so best to just remove the ones not adding much.

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.0/source.wdl
```diff
@@ -0,0 +1,19 @@
+#@ except: DescriptionMissing, DisallowedInputName, MissingRequirements
```
a-frantz, line 1:
```suggestion
#@ except: DescriptionMissing, MissingRequirements
```
Can you change the variable names to something that doesn't trigger this lint?

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.1/source.wdl
```diff
@@ -0,0 +1,33 @@
+#@ except: DescriptionMissing, DisallowedInputName, MissingRequirements
```
a-frantz, line 1:
```suggestion
#@ except: DescriptionMissing, MissingRequirements
```
Same here

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.2/source.wdl
```diff
@@ -0,0 +1,37 @@
+#@ except: DescriptionMissing, DisallowedInputName, MissingRequirements
```
a-frantz, line 1:
```suggestion
#@ except: DescriptionMissing, MissingRequirements
```
and here

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.2/source.wdl
```diff
@@ -0,0 +1,37 @@
...
+        Int b
+    }
+
+    # Should flag since version >= 1.1 and there are redundant input assignments
+    # This test was created to ensure the rule works without explicit "input" keyword
+    #@ except: RedundantInputAssignment
+    call bar {
+         a,  # should not flag
+         b = b + 3,  # should not flag
+         c = c,  # should flag
```
a-frantz, line 35:
```suggestion
         c = c,  # This should not flag a note due to the except statement
```

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.2/source.wdl
```diff
@@ -0,0 +1,37 @@
...
+
+workflow test2 {
+    input {
+        String a
+        String c
+        Int b
+    }
+
+    # Should flag since version >= 1.1 and there are redundant input assignments
+    # This test was created to ensure the rule works without explicit "input" keyword
```
a-frantz, line 30:
```suggestion
```

a-frantz (commented):
Looking really good here. Just some minor nit picks, and then we should be good to go 👍

Gauntlet.toml
michaelgattas: Is this for both `cargo run --release --bin gauntlet -- --refresh` and `cargo run --release --bin gauntlet -- --refresh --arena`? Or just the former?

a-frantz: Sorry for not being clear. Revert the changes to `Gauntlet.toml`, but leave `Arena.toml` as it is currently. Then run `cargo run --release --bin gauntlet` (without `--refresh`) to make sure it still passes (it should if you did the first step correctly). You can run `cargo run --release --bin gauntlet -- --refresh --arena` and commit any changes that makes.

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Flags redundant input assignments."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "Redundant input assignments can be shortened in WDL versions >=v1.1. For example, `{ input: a = a }` can be shortened to `{ input: a }`."
```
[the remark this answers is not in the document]
a-frantz: Good change. Let's commit it 👍

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.0/source.wdl
```diff
@@ -0,0 +1,19 @@
...
+workflow test {
+    input {
+        String arm
+        String cam
+        Int bam
+    }
+
+    # This should not flag any notes, since version is 1.0
+    call bar { input:
+        arm,
```
a-frantz, line 15: @adthrasher do you know why this isn't causing an error? It shouldn't be allowed in 1.0 doc, or am I misunderstanding something in the spec?

wdl-lint/tests/lints/blank-lines-between-elements/source.errors
a-frantz: > So after you make that change, this source.errors file should not have any changes to it in your PR.

The change present in `source.errors` makes sense. Ignore my statement there shouldn't be any changes in it. This looks good 😎

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.0/source.wdl
adthrasher: Shouldn't there be an error that `bar` is not defined?

adthrasher: But yes, I would think that `arm` should trigger a parse error since it isn't allowed in 1.0.

a-frantz: > Shouldn't there be an error that `bar` is not defined?

I think this is ok, but I need to code dive to verify. I believe that the undefined call check doesn't happen prior to our lint checks (only as part of `analysis` which these tests don't run).

> But yes, I would think that `arm` should trigger a parse error since it isn't allowed in 1.0.

This however is concerning to me. Going to do some investigating.

@michaelgattas you may have discovered a bug here. Not caused by your changes, but we should hold off merging till we know what's going on here.

wdl-lint/src/rules/redundant_input_assignment.rs
```diff
@@ -0,0 +1,112 @@
...
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Flags redundant input assignments."
+    }
+
+    fn explanation(&self) -> &'static str {
+        "Redundant input assignments can be shortened in WDL versions >=v1.1. For example, `{ \
+         input: a = a }` can be shortened to `{ input: a }`."
```
adthrasher, line 48:
```suggestion
        "Redundant input assignments can be shortened in WDL versions >=v1.1 with an implicit \
         binding. For example, `{ input: a = a }` can be shortened to `{ input: a }`."
```

adthrasher: Not sure why the other shows as outdated, but copied the change here.

adthrasher (approved):

wdl-lint/tests/lints/redundant-input-assignment-wdl-1.0/source.wdl
michaelgattas: If this should have thrown an error, then when everything works as expected, this test file should be modified so that it doesn't indicate an error, right?

a-frantz: We'll cross that bridge when we get to it. I actually want to leave the error in because I want to see it trip when that bug gets addressed.

PR #246, opened 2024-11-01
feat: support analysis of URLs

a-frantz:
Adds remote file support to `wdl-analysis` and the `wdl` binary.

With this change, we have the option to stop cloning repos in Gauntlet.

      changes (when appropriate).


wdl-analysis/tests/analysis.rs
```diff
@@ -222,7 +224,7 @@ async fn main() {
         let document = test.join("source.wdl");
         let uri = path_to_uri(&document).expect("should be valid URI");
         analyzer
-            .add_documents(vec![document])
+            .add_document(path_to_uri(&document).expect("should be valid URI"))
             .await
             .expect("should add documents");
```
adthrasher, line 229:
```suggestion
            .expect("should add document");
```

wdl-analysis/tests/analysis.rs
```diff
@@ -169,10 +169,12 @@ async fn main() {
...
     let analyzer = Analyzer::new(rules(), |_, _, _, _| async {});
-    analyzer
-        .add_documents(tests.clone())
-        .await
-        .expect("should add documents");
+    for test in &tests {
+        analyzer
+            .add_directory(test.clone())
+            .await
+            .expect("should add documents");
```
adthrasher, line 176:
```suggestion
            .expect("should add directory");
```

wdl-engine/tests/inputs.rs
```diff
@@ -194,10 +194,12 @@ async fn main() {
...
     let analyzer = Analyzer::new(rules(), |_, _, _, _| async {});
-    analyzer
-        .add_documents(tests.clone())
-        .await
-        .expect("should add documents");
+    for test in &tests {
+        analyzer
+            .add_directory(test.clone())
+            .await
+            .expect("should add documents");
```
adthrasher, line 201:
```suggestion
            .expect("should add directory");
```

wdl-lsp/src/server.rs
```diff
@@ -591,19 +592,14 @@ impl LanguageServer for Server {
...
-                .await
-            {
-                error!("failed to add documents to analyzer: {e}");
+            for folder in &params.event.added {
+                if let Err(e) = self
+                    .analyzer
+                    .add_directory(folder.uri.to_file_path().expect("should be a file path"))
+                    .await
+                {
+                    error!("failed to add documents to analyzer: {e}");
```
adthrasher, line 601:
```suggestion
                    error!("failed to add documents from directory to analyzer: {e}");
```

wdl/src/bin/wdl.rs
```diff
@@ -103,7 +105,15 @@ async fn analyze<T: AsRef<dyn Rule>>(
         },
     );
 
-    analyzer.add_documents(vec![path]).await?;
+    // Convert the file to a either a local URL or a remote URL
+    let file = if file.starts_with("http://") || file.starts_with("https://") {
```
adthrasher, line 109: Is this sufficient? What about other protocols such as FTP? Or URLs that omit the leading protocol?

a-frantz: Would it be better to just always call `Url::parse()`, and if that fails attempt `PathBuf::from()` (and if that fails, `bail!()`)?

a-frantz: Just typed up the following and it seems to work:
```rust
    let file = if let Ok(url) = Url::parse(&file) {
        url
    } else if let Some(url) = path_to_uri(&PathBuf::from(file.clone())) {
        url
    } else {
        bail!("failed to convert `{file}` to a URI", file = file)
    };
```

adthrasher: That's what I was wondering. I'm curious to hear other thoughts on this, but it makes sense to me.

peterhuene: I think it makes sense to attempt to parse it as a URI and fallback to treating it as a path (also, `path_to_uri` should probably be changed to take `impl AsRef<Path>` instead of `&Path`).

Regarding the supported schemes for the URIs, I think only `http`, `https`, and `file` should be supported as it's the only schemes required by the standard for import statements (with `file` currently deprecated and `http` should definitely be deprecated, but isn't currently).

wdl/src/bin/wdl.rs
```diff
@@ -103,7 +105,15 @@ async fn analyze<T: AsRef<dyn Rule>>(
         },
     );
 
-    analyzer.add_documents(vec![path]).await?;
+    let file = if let Ok(url) = Url::parse(&file) {
+        url
+    } else if let Some(url) = path_to_uri(&PathBuf::from(file.clone())) {
```
peterhuene:
```suggestion
    } else if let Some(url) = path_to_uri(&Path::new(&file)) {
```

Or simply `path_to_uri(&file)` if we change `path_to_uri` to take `impl AsRef<Path>`.

peterhuene (commented):
Looks good, just the nit comment.

peterhuene (approved):

PR #247, opened 2024-11-03
feat: adds groundwork for formatting in the Sprocket VSCode extension

claymcleod:
This PR is the first of several needed to get formatting working in VSCode.

      changes (when appropriate).

### DIFF
--- a/wdl-analysis/src/queue.rs
+++ b/wdl-analysis/src/queue.rs
@@ -24,6 +24,10 @@ use tracing::info;
 use url::Url;
 use wdl_ast::Ast;
 use wdl_ast::AstToken;
+use wdl_ast::Node;
+use wdl_ast::Severity;
+use wdl_format::Formatter;
+use wdl_format::element::node::AstNodeFormatExt as _;
 
 use crate::AnalysisResult;
 use crate::DiagnosticsConfig;
@@ -51,6 +55,8 @@ pub enum Request<Context> {
     NotifyIncrementalChange(NotifyIncrementalChangeRequest),
     /// A request to process a document's change.
     NotifyChange(NotifyChangeRequest),
+    /// A request to format a document.
+    Format(FormatRequest),
 }
 
 /// Represents a request to add documents to the graph.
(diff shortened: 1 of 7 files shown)


wdl-analysis/src/queue.rs
```diff
@@ -239,6 +259,57 @@ where
...
+                                    // NOTE: if we haven't parsed the document yet, then
+                                    // we don't have the line lengths of the document,
+                                    // so we can't proceed with formatting and we should
+                                    // just silently return.
+                                    ParseState::NotParsed | ParseState::Error(_) => None,
+                                    ParseState::Parsed {
+                                        version: _,
+                                        root: _,
+                                        lines,
+                                        diagnostics,
```
a-frantz, line 282: This is just "parse" diagnostics, correct? The output of running `wdl_ast::Document::parse()`? Just checking my understanding that this does not include analysis diagnostics.

claymcleod: Well, this does include _validation_ diagnostics. If you go look at the definition of the struct, it's a `wdl_analysis::graph::Diagnostics`.

That being said, we want to catch errors whether they happen during parsing or validation and decline to format in those cases. For example, if you include two input blocks, that fails _validation_, but not parsing. In that case, if you don't catch the validation errors, it will just delete one of the input blocks (because the assumption that the WDL has been validated is inherent in the formatting code).

There might even be a case to be made that we don't depend on the caller to ensure that the WDL is validated when calling format—instead, we require programmatically that the WDL parses and validates.

a-frantz: Correct me if wrong, but as I understand there are 3 classes of errors we have to deal with:
- **parse** errors (like a missing close delimiter)
- **validation** errors (like two input blocks for the same task)
- **analysis** errors (mismatched type or _any class of error in an import_)

Formatting a document with either parse or validation errors will lead to odd behavior (such as things getting deleted). But I think we should tolerate analysis errors, as none of the format code makes assumptions about analysis validity.

A bad import shouldn't prohibit formatting IMO.

claymcleod: Yes—what you describe is my understanding of how the current code will work.

peterhuene: That's correct, these are either the parse diagnostics or the validation diagnostics (if there were no parse diagnostics), so this is the correct collection to check to see if we should format the document.

a-frantz: Sounds good. Gets my 👍

peterhuene (approved):
Looks great, excited for folks to use `wdl-format` this way!

a-frantz (approved):

PR #249, opened 2024-11-10
feat: implement expression evaluation in `wdl-engine`.

peterhuene:
This PR implements expression evaluation in `wdl-engine` with the exception of function calls, which are dependent upon an upcoming implementation of the WDL standard library.

Additionally, certain WDL 1.2 features (such as task variables) and call output access expressions haven't been fully implemented yet; those require changes to the representation of WDL values, which will be completed alongside task and workflow evaluation.

These changes also include some refactorings:

* String interning was removed from `Engine` as it complicated the API; we can reintroduce interning of strings if it proves necessary. This change removed the requirement that values needed to be constructed via the engine.
* Likewise, values are no longer stored in an arena to simplify the API.
* An `EvaluationContext` trait was introduced in `wdl-analysis` so that context could be provided for expression type evaluation that wasn't dependent on `wdl-analysis`'s representation of scopes; expression evaluation in `wdl-engine` needs to statically determine types in the ternary operator, but use its own scope representation.
* `Document` in `wdl-analysis` now keeps a reference to the root CST node.
* A maximum number of parameters for functions in the standard library was added; this allows call expression evaluation to stack alloc space for the values and types of the arguments; the number will increase when more functions are defined in future specifications.
* `Type` and `Value` no longer implement `PartialEq`/`Eq`, as the former has a separate concept of "type equality" that resolves through the types arena and the latter has to handle certain type coercions for value equality.
* The underlying representation of `Value` has been split into `PrimitiveValue` and `CompoundValue`. This helped to simplify the coercion logic and make it so that only `PrimitiveValue` can implement `PartialEq`, `Eq`, and `Hash` for use in maps.
* Constructing a CST from parser events was made public in `wdl-grammar`; this will allow for `wdl-engine` unit tests to parse expressions directly. In the future, it'll be used for partial reparses in `wdl-analysis`.

      changes (when appropriate).

### DIFF
--- a/wdl-engine/src/eval/v1.rs
+++ b/wdl-engine/src/eval/v1.rs
@@ -1,0 +1,2984 @@
+//! Implementation of an expression evaluator for 1.x WDL documents.
+
+use std::cmp::Ordering;
+use std::fmt::Write;
+use std::iter::once;
+use std::sync::Arc;
+
+use indexmap::IndexMap;
+use ordered_float::Pow;
+use wdl_analysis::diagnostics::ambiguous_argument;
+use wdl_analysis::diagnostics::argument_type_mismatch;
+use wdl_analysis::diagnostics::cannot_access;
+use wdl_analysis::diagnostics::cannot_coerce_to_string;
+use wdl_analysis::diagnostics::cannot_index;
+use wdl_analysis::diagnostics::comparison_mismatch;
+use wdl_analysis::diagnostics::if_conditional_mismatch;
+use wdl_analysis::diagnostics::index_type_mismatch;
+use wdl_analysis::diagnostics::logical_and_mismatch;
+use wdl_analysis::diagnostics::logical_not_mismatch;
+use wdl_analysis::diagnostics::logical_or_mismatch;
+use wdl_analysis::diagnostics::map_key_not_primitive;
+use wdl_analysis::diagnostics::missing_struct_members;
+use wdl_analysis::diagnostics::no_common_type;
+use wdl_analysis::diagnostics::not_a_pair_accessor;
+use wdl_analysis::diagnostics::not_a_struct_member;
+use wdl_analysis::diagnostics::numeric_mismatch;
+use wdl_analysis::diagnostics::too_few_arguments;
+use wdl_analysis::diagnostics::too_many_arguments;
+use wdl_analysis::diagnostics::type_mismatch_custom;
+use wdl_analysis::diagnostics::unknown_function;
+use wdl_analysis::diagnostics::unsupported_function;
+use wdl_analysis::stdlib::FunctionBindError;
+use wdl_analysis::stdlib::MAX_PARAMETERS;
+use wdl_analysis::stdlib::STDLIB;
+use wdl_analysis::types::ArrayType;
+use wdl_analysis::types::Coercible as _;
+use wdl_analysis::types::CompoundTypeDef;
+use wdl_analysis::types::MapType;
+use wdl_analysis::types::Optional;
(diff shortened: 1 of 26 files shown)


peterhuene:
A considerable number of changes lines are coming from a fresh to `Arena.toml` and `Gauntlet.toml`. I review the actual diagnostic changes in `Gauntlet.toml` and they look sound, resulting from upstream repo WDL changes.

The next largest chunk of lines comes from all the unit tests in the engine's expression evaluator; it's like a 2:1 ratio from implementation to tests.

wdl-engine/src/eval/v1.rs
```diff
@@ -0,0 +1,2984 @@
...
+                target.span(),
+            )),
+        }
+    }
+
+    /// Evaluates the type of an access expression.
+    fn evaluate_access_expr(&mut self, expr: &AccessExpr) -> Result<Value, Diagnostic> {
+        let (target, name) = expr.operands();
+
+        // TODO: implement support for task values (required for WDL 1.2 support)
```
claymcleod, line 1190: Just a random question: do you keep track of these separately? Or are you planning to finish the current prototype and then you'll search the codebase for `TODO`?

wdl-engine/src/eval/v1.rs
```diff
@@ -0,0 +1,2984 @@
...
+                self.context.types(),
+                value.ty(),
+                target.span(),
+            )),
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
claymcleod, line 1222: The comprehensiveness of the tests here is epic.

wdl-engine/src/eval/v1.rs
```diff
@@ -0,0 +1,2984 @@
...
+        let value = eval_v1_expr(V1::Zero, r#"true || false"#, &mut types, scope).unwrap();
+        assert!(value.unwrap_boolean());
+
+        let value = eval_v1_expr(V1::Zero, r#"true || true"#, &mut types, scope).unwrap();
+        assert!(value.unwrap_boolean());
+
+        let value = eval_v1_expr(V1::Zero, r#"true || nope"#, &mut types, scope).unwrap();
+        assert!(value.unwrap_boolean());
+
+        let diagnostic = eval_v1_expr(V1::Zero, r#"false || nope"#, &mut types, scope).unwrap_err();
```
claymcleod, line 1936: Though it's cool that something like this is left to runtime, I do wonder if it needs to be statically checked to avoid errors (possibly a new lint rule?).

claymcleod (approved):
LGTM. Granted, there are so many changes here that I found it difficult to go through with a fine-toothed comb. That being said, everything generally seemed reasonable, and I'm of the mind to ship it and then react if we come across any issues.

wdl-engine/src/eval/v1.rs
peterhuene: I'm keeping track of these separately, but also when the task evaluation is implemented, a bunch of these will be as well.

wdl-engine/src/eval/v1.rs
peterhuene: `true || nope` (where `nope` is an unknown name, like in this case) or where `nope` might be a name that resolves to a non-boolean type are caught at static analysis time.

The expression evaluator, though, is implemented in such a way that it doesn't care what the RHS is.

When we hook this all up to `sprocket`, we'll be reporting on any error-level static analysis diagnostics and exiting non-zero before we even get to evaluation.

claymcleod: Okay, great! Then that is taken care of.

PR #251, opened 2024-11-12
feat: implement the standard library numeric functions.

peterhuene:
This PR implements call expressions for the following standard library
functions:

* `floor`
* `ceil`
* `round`
* `min`
* `max`

Additionally, this PR refactors binding functions in `wdl-analysis` to provide
more information to allow for dispatching the call to the implementation in
`wdl-engine`.

One of the bigger changes is that the minimum WDL version specification
for functions has moved from the function itself to the signatures; this
means that future versions of WDL can add a new overload to an existing
function and the new signature will not be included in the overload set
when evaluating for a previous version of WDL.

      changes (when appropriate).

### DIFF
--- a/wdl-analysis/src/diagnostics.rs
+++ b/wdl-analysis/src/diagnostics.rs
@@ -657,7 +657,7 @@ pub fn unknown_function(name: &str, span: Span) -> Diagnostic {
 /// Creates an "unsupported function" diagnostic.
 pub fn unsupported_function(minimum: SupportedVersion, name: &str, span: Span) -> Diagnostic {
     Diagnostic::error(format!(
-        "function `{name}` requires a minimum WDL version of {minimum}"
+        "this use of function `{name}` requires a minimum WDL version of {minimum}"
     ))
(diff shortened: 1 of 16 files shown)


wdl-analysis/src/diagnostics.rs
```diff
@@ -657,7 +657,7 @@ pub fn unknown_function(name: &str, span: Span) -> Diagnostic {
 /// Creates an "unsupported function" diagnostic.
 pub fn unsupported_function(minimum: SupportedVersion, name: &str, span: Span) -> Diagnostic {
     Diagnostic::error(format!(
-        "function `{name}` requires a minimum WDL version of {minimum}"
+        "this use of function `{name}` requires a minimum WDL version of {minimum}"
```
peterhuene, line 660: FYI, the language here was changed as it's possible that the function itself has a supported minimum version, but the way the function was called required a newer version of WDL to be used (i.e. you're calling an overload that exists in a later WDL version).

a-frantz (approved):

PR #254, opened 2024-11-18
feat: implement the file function from the WDL standard library.

peterhuene:
This PR implements the following functions from the WDL standard library:

- `basename`
- `join_paths`
- `glob`
- `size`
- `stdout`
- `stderr`
- `read_string`
- `read_int`
- `read_float`
- `read_boolean`
- `read_lines`
- `write_lines`
- `read_tsv`
- `write_tsv`
- `read_map`
- `write_map`
- `read_json`
- `write_json`
- `read_object`
- `read_objects`
- `write_object`
- `write_objects`

It also corrects some issues in the definitions of the functions in `wdl-analysis` and constraining certain generic function parameter types to structs with _only_ primitive members.

The function call implementation in `wdl-engine` has been refactored so that the current working directory and temp directory and available to the call context; consequently, the function tests in `wdl-engine` have been refactored to make it easier to create files in a temporary directory.

The function implementations themselves have been refactored into their own modules so as to make the implementations/tests isolated from one another.

      changes (when appropriate).


peterhuene:
The tests have some Windows-path-specific logic to them that I haven't tested on Windows yet, so the Windows CI might fail; I'll investigate if it does.

peterhuene:
So the gauntlet error is missing a diagnostic that should have been previously emitted.

The problematic line is:

```wdl
disk_size          = size(flatten(hifi_reads), "GB") + 10
```

Where `disk_size` is expected to be an `Int`, but `size` returns `Float` and thus the right hand side expression is `Float` (which doesn't coerce to `Int`).

This was actually fixed in the repository by wrapping the `size` call with a ` ceil` call.

Looking into which specific change in this PR caused gauntlet to fail (in a good way).

peterhuene:
The windows failures are due to path separator issues; I'll get that cleaned up.

peterhuene:
After determining that none of these changes should have affected the gauntlet diagnostic, I can confirm that I'm seeing the missing diagnostic in `main` as well. I'm not sure why the diagnostic is missing; I'll push up a refresh of gauntlet.

wdl-engine/src/stdlib/read_object.rs
```diff
@@ -0,0 +1,229 @@
...
+        let diagnostic =
+            eval_v1_expr(&mut env, V1::Two, "read_object('duplicate.tsv')").unwrap_err();
+        assert!(diagnostic.message().contains(
+            "call to function `read_object` failed: duplicate column name `foo` in line 1 of file"
+        ));
+
+        let diagnostic =
+            eval_v1_expr(&mut env, V1::Two, "read_object('invalid-name.tsv')").unwrap_err();
+        assert!(diagnostic.message().contains(
+            "call to function `read_object` failed: column name `bar-wrong` in line 1 of file"
```
claymcleod, line 215: Can this error message be better? It appears to be an incorrect column name.

claymcleod (approved):

wdl-engine/src/stdlib/read_object.rs
peterhuene: it's split this way because it contains the path to the file it failed to read, which in this case are always in the temp directory:

```
call to function `read_object` failed: column name `bar-wrong` in line 1 of file `/var/folders/9t/yj3jt2pn5zx5mzpy6hqc8f540000gp/T/.tmpsDwkHf/invalid-name.tsv` is not a valid WDL object field name
```

Happy to reword it as it is definitely clunky.

wdl-engine/src/stdlib/glob.rs
```diff
@@ -0,0 +1,162 @@
...
+use super::Function;
+use super::Signature;
+use crate::Array;
+use crate::PrimitiveValue;
+use crate::Value;
+use crate::diagnostics::function_call_failed;
+use crate::diagnostics::invalid_glob_pattern;
+
+/// Returns the Bash expansion of the glob string relative to the task's
+/// execution directory, and in the same order.
```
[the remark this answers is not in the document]
peterhuene: I think it means to say "and in the same order that a bash glob would", i.e. lexographically (which is what Rust's `glob` implementation does as well).

These comments were just taken from the 1.2 standard mostly verbatim. I can amend this one, though.

wdl-engine/src/stdlib/read_tsv.rs
```diff
@@ -0,0 +1,480 @@
...
+        }
+    }
+}
+
+/// Reads a tab-separated value (TSV) file as an Array[Array[String]]
+/// representing a table of values.
+///
+/// Trailing end-of-line characters (\r and \n) are removed from each line.
+///
+/// `Array[Array[String]] read_tsv(File, [false])``: Returns each row of the
```
a-frantz, line 58: I don't understand this signature. What's `[false]` mean? And looks like an extra backtick.

wdl-engine/src/stdlib/read_lines.rs
```diff
@@ -0,0 +1,117 @@
...
+    let elements = BufReader::new(file)
+        .lines()
+        .map(|line| {
+            let mut line = line
+                .with_context(|| format!("failed to read file `{path}`", path = path.display()))
+                .map_err(|e| {
+                    function_call_failed("read_lines", format!("{e:?}"), context.call_site)
+                })?;
+
+            let trimmed = line.trim_end_matches(['\r', '\n']);
```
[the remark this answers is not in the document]
peterhuene: Yep, you're right; looks like it pops off a `\n` and then a `\r` if there is one regardless of platform, so I'll remove this.

wdl-engine/src/stdlib/read_tsv.rs
```diff
@@ -0,0 +1,480 @@
...
+/// table as an Array[String]. There is no requirement that the rows of the
+/// table are all the same length.
```
a-frantz, line 60: Understand you're just implementing the spec, but I do not like supporting TSVs with variable length rows. TSVs can be such a headache. I am very particular in how I like my TSVs 😅

wdl-engine/src/stdlib/read_tsv.rs
```diff
@@ -0,0 +1,480 @@
...
+            .collect::<Vec<Value>>();
+        rows.push(
+            Array::new_unchecked(ANALYSIS_STDLIB.array_string_type(), Arc::new(values)).into(),
+        );
+    }
+
+    Ok(Array::new_unchecked(ANALYSIS_STDLIB.array_array_string_type(), Arc::new(rows)).into())
+}
+
+/// Reads a tab-separated value (TSV) file as an Array[Array[String]]
```
a-frantz, line 95: Shouldn't this be:
```suggestion
/// Reads a tab-separated value (TSV) file as an Array[Object]
```
?

wdl-engine/src/stdlib/read_lines.rs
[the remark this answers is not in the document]
a-frantz: Watch out that you use this pattern in a few places. I might not have commented on all of them, I'm not sure.

wdl-engine/src/stdlib/read_tsv.rs
peterhuene: There's a couple of extra backticks in comments that VS Code auto inserted for me that I'll fix.

Regarding the signature, that's a great question. It takes a `Boolean` here, but it's _required_ to be `false` for this overload, which means this function is actually overloaded on its _return type_ only, which is incorrect.

I filed https://github.com/openwdl/wdl/issues/690 earlier to ask them to clean this up.

wdl-engine/src/stdlib/read_tsv.rs
peterhuene: Indeed 😢

a-frantz: Not sure how much experience with TSVs you have, but they are rampant in bioinformatics. And the conventions are... erratic to put it one way. Which is how I've come to form _very_ strong opinions on what a "good" TSV looks like. Unfortunately the world is not ready for treatise on TSV formatting, and we have to support such nonsense as variable length rows 😨

wdl-engine/src/stdlib/size.rs
```diff
@@ -0,0 +1,300 @@
...
+    Function::new(
+        const {
+            &[
+                Signature::new("(None, <String>) -> Float", size),
+                Signature::new("(File?, <String>) -> Float", size),
+                Signature::new("(String?, <String>) -> Float", size),
+                Signature::new("(Directory?, <String>) -> Float", size),
+                Signature::new(
+                    "(X, <String>) -> Float where `X`: any compound type that recursively \
+                     contains a `File` or `Directory`",
```
a-frantz, line 190: So what if not all the members are `File` or `Directory`? Are other primitive types just ignored?

peterhuene: That's correct, it only looks at `File` and `Directory` values; all other values are treated as size 0.

peterhuene: If a type is passed which contains no `File` or `Directory`, then it'll be an analysis error as the type will fail the constraint on the overload.

wdl-engine/src/stdlib/size.rs
```diff
@@ -0,0 +1,300 @@
...
+                    "failed to read metadata for file `{path}`",
+                    path = path.display()
+                )
+            })?;
+
+            if !metadata.is_file() {
+                bail!("path `{path}` is not a file", path = path.display());
+            }
+
+            Ok(unit.convert(metadata.len()))
```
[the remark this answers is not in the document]
peterhuene: Which part, specifically? This is just converting the file size in bytes to the requested optional unit argument.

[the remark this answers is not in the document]
a-frantz: Ah I didn't realize `metadata.len()` was the size of the file. NVM. I get it now 👍

a-frantz (approved):
I think you've addressed everything I pointed out, so LGTM! Making good progress here 🚂

PR #257, opened 2024-11-21
feat: implement the generic map functions from the WDL standardy library.

peterhuene:
This PR implements the following WDL standard library functions:

* `as_pairs`
* `as_map`
* `keys`
* `contains_key`
* `values`
* `collect_by_key`

It also corrects a few issues discovered during testing:
    
* Map values in `wdl-engine` should accept `None` for a key.
* Common type calculation now supports discovering common types between the
   compound types containing `Union` and `None` as inner types, e.g.
   `Array[String] | Array[None] -> Array[String?]`
* The "required primitive type" constraint has been removed as every place the
   constraint was used should allow for optional primitive types as well;
   consequently, the `AnyPrimitiveTypeConstraint` was renamed to simply
   `PrimitiveTypeConstraint`.
* The common type calculation now favors the "left-hand side" of the
   calculation rather than the right, making it more intuitive to use. For
   example, a calculation of `File | String` is now `File` rather than
   `String`.
    
Added test cases to existing functions that accept maps with optional primitive keys.

      changes (when appropriate).

### DIFF
--- a/wdl-engine/src/stdlib/contains_key.rs
+++ b/wdl-engine/src/stdlib/contains_key.rs
@@ -1,0 +1,354 @@
+//! Implements the `contains_key` function from the WDL standard library.
+
+use std::sync::Arc;
+
+use wdl_analysis::stdlib::STDLIB as ANALYSIS_STDLIB;
+use wdl_analysis::types::PrimitiveTypeKind;
+use wdl_analysis::types::Type;
+use wdl_ast::Diagnostic;
+
+use super::CallContext;
+use super::Function;
+use super::Signature;
+use crate::CompoundValue;
+use crate::Object;
+use crate::PrimitiveValue;
+use crate::Struct;
+use crate::Value;
+
+/// Given a Map and a key, tests whether the collection contains an entry with
+/// the given key.
+///
+/// `Boolean contains_key(Map[P, Y], P)`: Tests whether the Map has an entry
+/// with the given key. If P is an optional type (e.g., String?), then the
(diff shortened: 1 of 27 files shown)


wdl-engine/src/stdlib/contains_key.rs
```diff
@@ -0,0 +1,354 @@
...
+
+        let value = eval_v1_expr(
+            &mut env,
+            V1::Two,
+            "contains_key(object { foo: 1, bar: 2, baz: 3 }, ['baz'])",
+        )
+        .unwrap();
+        assert!(value.unwrap_boolean());
+
+        let value = eval_v1_expr(
```
claymcleod, line 239: Should we also test `baz` after this?

a-frantz (approved):

wdl-engine/src/stdlib/contains_key.rs
peterhuene: So that I better understand, what would be the expression to test?

peterhuene: We do test `['bar', 'baz']` further below (should be true), if that's what you mean.

Otherwise, a test for `['baz']` would be the same as a test for `['qux']` here (neither are keys of the `Foo` struct); I can change the test to use `baz` instead, though.

peterhuene: I'm going to merge this as-is and put up a follow-up PR if warranted after discussion.

PR #261, opened 2024-11-27
feat: break `refresh` opt into `bless` and `update`

### LINKED ISSUE #226: Gauntlet update to reduce hash changes
The current issue: Everytime a gauntlet or arena repo pushes a new commit, we can get hundreds of lines show up in the git diff but none of them we care about. This leads to a situation where diagnostic changes we care about may get lost in the noise. It also means PRs end up with a bunch of git changes that are unrelated to the PR.

The fix: (originally a comment left on #194 that I have now expanded on)
I envision there being two types of changes to `Gauntlet.toml`/`Arena.toml` and they won't overlap in the same PR:
1. diagnostic changes (without hash changes)
    - these changes happen in our "typical" PRs (for lack of a better word)
    - New lint diagnostics, changes in diagnostic language, etc.
    - These are the changes we really care about
2. hash changes (which may or may not include diagnostic changes)
    - these will happen in dedicated PRs that run maybe once a week?
        - Could be automated, could be manual
    - any diagnostic changes happening here, we will know is because of a change "upstream" and not because of a regression in our code

To facilitate this, `gauntlet` will need some changes. The "regular" and "arena" run modes will need to stop pulling in new changes. `--refresh` could be renamed `--bless`. Then we'd need a new flag (which is mutually exclusive with `--refresh/bless`) that goes and updates all the commit hashes (probably just call it `--update`?).

a-frantz:

closes #226 

      changes (when appropriate).

### DIFF
--- a/gauntlet/src/lib.rs
+++ b/gauntlet/src/lib.rs
@@ -143,7 +143,7 @@ pub async fn gauntlet(args: Args) -> Result<()> {
     let mut work_dir = WorkDir::default();
 
     if args.update {
-        info!("refreshing repository commit hashes.");
+        info!("updating repository commit hashes.");
         config.inner_mut().update_repositories(work_dir.root());
     }
 
--- a/gauntlet/src/lib.rs
+++ b/gauntlet/src/lib.rs
@@ -104,13 +104,13 @@ pub struct Args {
     /// Overwrites the configuration file with new expected diagnostics.
     ///
     /// Mutually exclusive with `--update`.
-    #[arg(long)]
+    #[arg(long, group = "action")]
     pub bless: bool,
 
     /// Updates the commit hashes for all repositories.
     ///
     /// Mutually exclusive with `--bless`.
-    #[arg(long)]
+    #[arg(long, group = "action")]
     pub update: bool,
 
     /// Displays warnings as part of the report output.
--- a/gauntlet/src/lib.rs
+++ b/gauntlet/src/lib.rs
@@ -101,12 +101,17 @@ pub struct Args {
     #[arg(short, long)]
     pub quiet: bool,
 
-    /// Overwrites the configuration file with new expected diagnostics and the
-    /// latest commit hashes. This will create temporary shallow clones of every
-    /// test repository. Normally, there is only one repository on disk at a
-    /// time. The difference in disk space usage should be negligible.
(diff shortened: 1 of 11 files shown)


gauntlet/src/lib.rs
```diff
@@ -380,7 +385,7 @@ pub async fn gauntlet(args: Args) -> Result<()> {
 
     println!("\nTotal analysis time: {total_time:?}");
 
-    if args.refresh {
+    if args.bless || args.update {
```
adthrasher, line 388: Should this only fire on `--bless`

a-frantz: No, we need to save the new diagnostics (and new hashes) back to the relevant config file when `--update` is passed. Otherwise `--update` would be a no-op that fails whenever something changed.

gauntlet/src/lib.rs
```diff
@@ -101,12 +101,17 @@ pub struct Args {
...
-    /// Overwrites the configuration file with new expected diagnostics and the
-    /// latest commit hashes. This will create temporary shallow clones of every
-    /// test repository. Normally, there is only one repository on disk at a
-    /// time. The difference in disk space usage should be negligible.
+    /// Overwrites the configuration file with new expected diagnostics.
+    ///
+    /// Mutually exclusive with `--update`.
     #[arg(long)]
-    pub refresh: bool,
+    pub bless: bool,
```
claymcleod, line 108: You need to put these into `clap` argument groups so that they will be shown as mutually exclusive and checked at argument parse time.

claymcleod: E.g., https://docs.rs/clap/latest/clap/builder/struct.ArgGroup.html. There is a way to do that with derive as well but you'll need to search for it.

gauntlet/src/lib.rs
```diff
@@ -137,7 +142,7 @@ pub async fn gauntlet(args: Args) -> Result<()> {
 
     let mut work_dir = WorkDir::default();
 
-    if args.refresh {
+    if args.update {
         info!("refreshing repository commit hashes.");
```
claymcleod, line 146:
```suggestion
        info!("updating repository commit hashes.");
```

claymcleod (approved):
After the above is addressed, looks good to me.

PR #262, opened 2024-12-05
feat: initial commit of `wdl-docs` style

claymcleod:
This PR adds a basic style scaffold for producing the CSS file used in `wdl-doc`. Styles can be added in the `src` folder and compiled into development (`pnpm run dev`, watches the directory and serves the `example.html` file live) or production (`pnpm run prod`, which needs a subsequent `pnpm run serve` to serve the `example.html` file).

You can update the `example.html` file with a fully-featured generated page that we can test with.

- Please use `pnpm`, as it's way better than `npm` in my experience.

### DIFF
--- a/wdl-doc/theme/.gitignore
+++ b/wdl-doc/theme/.gitignore
@@ -1,0 +1,2 @@
+node_modules/
+dist/
--- a/wdl-doc/theme/example.html
+++ b/wdl-doc/theme/example.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="dist/styles.css" />
+    <title>Document</title>
+  </head>
+  <body>
+    <h1>Foo</h1>
+  </body>
+</html>
--- a/wdl-doc/theme/package.json
+++ b/wdl-doc/theme/package.json
@@ -1,0 +1,21 @@
+{
+  "name": "wdl-docs-theme",
+  "author": "The Rust WDL project developers",
+  "private": true,
+  "scripts": {
+    "prod": "NODE_ENV=production pnpm postcss src/index.css -o dist/styles.css",
+    "dev": "concurrently --kill-others \"pnpm postcss src/index.css -o dist/styles.css -w\" \"pnpm http-server\" --names POSTCSS,SERVER",
+    "serve": "pnpm http-server"
+  },
+  "devDependencies": {
+    "autoprefixer": "^10.4.20",
+    "concurrently": "^9.1.0",
+    "cssnano": "^7.0.6",
+    "http-server": "^14.1.1",
+    "postcss": "^8.4.49",
+    "postcss-cli": "^11.0.0",
+    "postcss-import": "^16.1.0",
+    "postcss-url": "^10.1.3",
+    "tailwindcss": "^3.4.16"
+  }
+}
--- a/wdl-doc/theme/pnpm-lock.yaml
+++ b/wdl-doc/theme/pnpm-lock.yaml
@@ -1,0 +1,2146 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .:
+    devDependencies:
+      autoprefixer:
+        specifier: ^10.4.20
+        version: 10.4.20(postcss@8.4.49)
+      concurrently:
+        specifier: ^9.1.0
+        version: 9.1.0
+      cssnano:
+        specifier: ^7.0.6
+        version: 7.0.6(postcss@8.4.49)
+      http-server:
+        specifier: ^14.1.1
+        version: 14.1.1
+      postcss:
+        specifier: ^8.4.49
+        version: 8.4.49
+      postcss-cli:
+        specifier: ^11.0.0
+        version: 11.0.0(jiti@1.21.6)(postcss@8.4.49)
+      postcss-import:
+        specifier: ^16.1.0
+        version: 16.1.0(postcss@8.4.49)
+      postcss-url:
+        specifier: ^10.1.3
+        version: 10.1.3(postcss@8.4.49)
+      tailwindcss:
+        specifier: ^3.4.16
+        version: 3.4.16
+
+packages:
+
+  '@alloc/quick-lru@5.2.0':
+    resolution: {integrity: sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==}
+    engines: {node: '>=10'}
+
+  '@isaacs/cliui@8.0.2':
+    resolution: {integrity: sha512-O8jcjabXaleOG9DQ0+ARXWZBTfnP4WNAqzuiJK7ll44AmxGKv/J2M4TPjxjY3znBCfvBXFzucm1twdyFybFqEA==}
+    engines: {node: '>=12'}
+
+  '@jridgewell/gen-mapping@0.3.5':
+    resolution: {integrity: sha512-IzL8ZoEDIBRWEzlCcRhOaCupYyN5gdIK+Q6fbFdPDg6HqX6jpkItn7DFIpW9LQzXG6Df9sA7+OKnq0qlz/GaQg==}
+    engines: {node: '>=6.0.0'}
+
+  '@jridgewell/resolve-uri@3.1.2':
+    resolution: {integrity: sha512-bRISgCIjP20/tbWSPWMEi54QVPRZExkuD9lJL+UIxUKtwVJA8wW1Trb1jMs1RFXo1CBTNZ/5hpC9QvmKWdopKw==}
+    engines: {node: '>=6.0.0'}
+
+  '@jridgewell/set-array@1.2.1':
+    resolution: {integrity: sha512-R8gLRTZeyp03ymzP/6Lil/28tGeGEzhx1q2k703KGWRAI1VdvPIXdG70VJc2pAMw3NA6JKL5hhFu1sJX0Mnn/A==}
+    engines: {node: '>=6.0.0'}
+
+  '@jridgewell/sourcemap-codec@1.5.0':
+    resolution: {integrity: sha512-gv3ZRaISU3fjPAgNsriBRqGWQL6quFx04YMPW/zD8XMLsU32mhCCbfbO6KZFLjvYpCZ8zyDEgqsgf+PwPaM7GQ==}
+
+  '@jridgewell/trace-mapping@0.3.25':
+    resolution: {integrity: sha512-vNk6aEwybGtawWmy/PzwnGDOjCkLWSD2wqvjGGAgOAwCGWySYXfYoxt00IJkTF+8Lb57DwOb3Aa0o9CApepiYQ==}
+
+  '@nodelib/fs.scandir@2.1.5':
+    resolution: {integrity: sha512-vq24Bq3ym5HEQm2NKCr3yXDwjc7vTsEThRDnkp2DK9p1uqLR+DHurm/NOTo0KG7HYHU7eppKZj3MyqYuMBf62g==}
+    engines: {node: '>= 8'}
+
+  '@nodelib/fs.stat@2.0.5':
+    resolution: {integrity: sha512-RkhPPp2zrqDAQA/2jNhnztcPAlv64XdhIp7a7454A5ovI7Bukxgt7MX7udwAu3zg1DcpPU0rz3VV1SeaqvY4+A==}
+    engines: {node: '>= 8'}
+
+  '@nodelib/fs.walk@1.2.8':
+    resolution: {integrity: sha512-oGB+UxlgWcgQkgwo8GcEGwemoTFt3FIO9ababBmaGwXIoBKZ+GTy0pP185beGg7Llih/NSHSV2XAs1lnznocSg==}
+    engines: {node: '>= 8'}
+
+  '@pkgjs/parseargs@0.11.0':
+    resolution: {integrity: sha512-+1VkjdD0QBLPodGrJUeqarH8VAIvQODIbwh9XpP5Syisf7YoQgsJKPNFoqqLQlu+VQ/tVSshMR6loPMn8U+dPg==}
+    engines: {node: '>=14'}
+
+  '@sindresorhus/merge-streams@2.3.0':
+    resolution: {integrity: sha512-LtoMMhxAlorcGhmFYI+LhPgbPZCkgP6ra1YL604EeF6U98pLlQ3iWIGMdWSC+vWmPBWBNgmDBAhnAobLROJmwg==}
+    engines: {node: '>=18'}
+
+  '@trysound/sax@0.2.0':
+    resolution: {integrity: sha512-L7z9BgrNEcYyUYtF+HaEfiS5ebkh9jXqbszz7pC0hRBPaatV0XjSD3+eHrpqFemQfgwiFF0QPIarnIihIDn7OA==}
+    engines: {node: '>=10.13.0'}
+
+  ansi-regex@5.0.1:
+    resolution: {integrity: sha512-quJQXlTSUGL2LH9SUXo8VwsY4soanhgo6LNSm84E1LBcE8s3O0wpdiRzyR9z/ZZJMlMWv37qOOb9pdJlMUEKFQ==}
+    engines: {node: '>=8'}
+
+  ansi-regex@6.1.0:
+    resolution: {integrity: sha512-7HSX4QQb4CspciLpVFwyRe79O3xsIZDDLER21kERQ71oaPodF8jL725AgJMFAYbooIqolJoRLuM81SpeUkpkvA==}
+    engines: {node: '>=12'}
+
+  ansi-styles@4.3.0:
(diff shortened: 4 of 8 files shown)


wdl-doc/tests/output_docs/data_structures/read_group/read_group_to_string-task.html
```diff
@@ -1,6 +1,3 @@
-<h1>read_group_to_string</h1><ul><li>Meta:meta {
-        description: "Stringifies a ReadGroup struct"
-        outputs: {
-            stringified_read_group: "Input ReadGroup as a string"
-        }
-    }</li><li>Inputs:<ul><li><h2>read_group</h2><p>Type: ReadGroup</p><p>Expr: None</p></li></ul></li><li>Outputs:<ul><li><h2>stringified_read_group</h2><p>Type: String</p><p>Expr: read_string("out.txt")</p></li></ul></li></ul>
\ No newline at end of file
+<!DOCTYPE html><html lang="en"><head><meta charset="utf-8"><title>read_group_to_string</title><link rel="stylesheet" type="text/css" href="/Users/afrantz/wdl/wdl-doc/theme/dist/styles.css"></head></html><h1>read_group_to_string</h1><h2>Meta</h2><ul><li><b>description:</b> &quot;Stringifies a ReadGroup struct&quot;</li><li><b>outputs:</b> {
```
claymcleod, line 1: 1. We should remove the references to your home directory somehow.
2. It looks like the `<html>` tag closes before your elements are added in.

a-frantz: 1. Already got a WIP addressing that by using a relative path instead (though it complicates things a touch). I noticed that right after I made the WIP commit.
2. I'll look into this!

wdl-doc/src/docs_tree.rs
```diff
@@ -0,0 +1,336 @@
...
+        &self.path
+    }
+
+    /// Get the page associated with the node.
+    pub fn page(&self) -> Option<Rc<HTMLPage>> {
+        self.page.clone()
+    }
+
+    /// Iterate over the children of the node in a Depth First Traversal order.
+    pub fn depth_first_traversal(&self) -> Vec<&Node> {
```
peterhuene, line 98: I think this function could be implemented with a local function that takes the `Vec` by mutable reference so that it doesn't have to allocate and deallocate a `Vec` at every node.

For example:

```rust
pub fn depth_first_traversal(&self) -> Vec<&Node> {
    fn recurse_depth_first(node: &Node, nodes: &mut Vec<&Node>) {
        nodes.push(self);

        for child in self.children.values() {
            recurse_depth_first(child, nodes);
        }
    }

    let mut nodes = Vec::new();
    recurse_depth_first(self, &mut nodes);

    nodes
}
```

peterhuene: Alternatively, depending on how this method is used, we could not return a `Vec` and instead accept a `FnMut` callback that would "visit" each node in DFS order.

It certainly looks like the places we use this method could be translated to:

```rust
root.depth_first_traversal(|node| {
    // do something with node
});
```

wdl-doc/src/docs_tree.rs
```diff
@@ -0,0 +1,336 @@
...
+    ///
+    /// `root.path` is the path to the DOCS directory and is absolute.
+    root: Node,
+    /// The absolute path to the stylesheet, if it exists.
+    stylesheet: Option<PathBuf>,
+}
+
+impl DocsTree {
+    /// Create a new DOCS tree.
+    pub fn new<P: AsRef<Path>>(root: P, stylesheet_to_copy: Option<P>) -> Self {
```
peterhuene, line 123: When repeating a type parameter in a signature like this, it's probably best to use `impl AsRef<Path>` in the signatures so that they don't have to be the same type (i.e. one should be able to call this with `new("foo", Some(Path::new("foo")))`. It's functionally equivalent to having two type parameters `P: AsRef<Path>, Q: AsRef<Path>`.

However, the `Option<impl AsRef<Path>>` is annoying to deal with as you would have to qualify it in the `None` case.

When I see something like that, I tend to break the method into two: `new(root: impl AsRef<Path>)` and `new_with_stylesheet(root: impl AsRef<Path>, stylesheet: impl AsRef<Path>)` and let it internally deal with the `Option`.

wdl-doc/src/callable/task.rs
```diff
@@ -0,0 +1,148 @@
...
+        parameter_meta: Option<ParameterMetadataSection>,
+        input_section: Option<InputSection>,
+        output_section: Option<OutputSection>,
+        runtime_section: Option<RuntimeSection>,
+    ) -> Self {
+        let meta = if let Some(ref mds) = meta_section {
+            parse_meta(mds)
+        } else {
+            MetaMap::default()
+        };
```
peterhuene, line 44: Nit for consistency since the other `if let` expressions don't use `ref`:

```suggestion
        let meta = if let Some(mds) = meta_section {
            parse_meta(&mds)
        } else {
            MetaMap::default()
        };
```

wdl-doc/src/callable/workflow.rs
```diff
@@ -0,0 +1,136 @@
...
+        meta_section: Option<MetadataSection>,
+        parameter_meta: Option<ParameterMetadataSection>,
+        input_section: Option<InputSection>,
+        output_section: Option<OutputSection>,
+    ) -> Self {
+        let meta = if let Some(ref mds) = meta_section {
+            parse_meta(mds)
+        } else {
+            MetaMap::default()
+        };
```
peterhuene, line 41: Same comment as above re: `let` usage:

```suggestion
        let meta = if let Some(mds) = meta_section {
            parse_meta(&mds)
        } else {
            MetaMap::default()
        };
```

wdl-doc/src/docs_tree.rs
```diff
@@ -0,0 +1,336 @@
...
+
+        let html = full_page(
+            page.name(),
+            html! {
+                (sidebar)
+                (content)
+            },
+            stylesheet.as_deref(),
+        );
+        std::fs::write(path, html.into_string()).unwrap();
```
peterhuene, line 334: We should probably functions that interact with the file system fallible rather than panic on IO error.

wdl-doc/src/docs_tree.rs
peterhuene: Although, looking at what `maud` supports, we probably can't use this within a `html!` macro, so returning a `Vec` is fine.

peterhuene:
I was going to message this on Slack, but apparently Slack is having an outage:

> I put some minor comments on the PR, although I know it's not done yet; generally I think the approach looks really good!

wdl-doc/src/docs_tree.rs
a-frantz: Would ask over Slack, but it's still down! I've updated `DocsTree::new` to just take an `impl AsRef<Path>` and then wrote `new_with_stylesheet` like you usggested.

The pattern you pointed here is also in use for the `document_workspace()` entrypoint function, but I'm not sure the right way to refactor that. Should there be a pair of `document_worskspace()` and `document_workspace_with_stylesheet()` functions? Or is there another way to refactor it that you would recommend?

wdl-doc/src/docs_tree.rs
a-frantz: good point! I pretty heavily use `unwrap()` in this PR as a shortcut, but just went through and updated some functions to return `anyhow::Result` instead and also changed some unwraps to expects.

wdl-doc/src/docs_tree.rs
peterhuene: `document_workspace_with_stylesheet` definitely is a mouthful; if we want two different functions, I'm sure we'd all prefer something less verbose.

Still, perhaps having the `Option` parameter is fine and folks won't mind passing something like `None::<&str>`? I don't have any suggested alternatives, but I still think we should be using different type parameters at the very least.

wdl-doc/src/callable.rs
```diff
@@ -0,0 +1,292 @@
...
+                    tbody class="border" {
+                        @for param in iter {
+                            (param.render())
+                        }
+                    }
+                }
+            };
+        };
+        html! {}
+    }
```
adthrasher, line 134: Do we need to treat `Common` separately? Or can we apply a custom sort to `input_groups`?

wdl-doc/src/callable.rs
```diff
@@ -0,0 +1,292 @@
...
+                        (param.render())
+                    }
+                }
+            }
+        }
+    }
+}
+
+/// Parse a [`MetadataSection`] into a [`MetaMap`].
+fn parse_meta(meta: &MetadataSection) -> MetaMap {
```
adthrasher, line 227: Can we process `meta` and `parameter_meta` together? Is there a generic trait or type that both share to enable a single function?

a-frantz: I can play around with this locally and see if I can join these methods together. Good thought

wdl-doc/src/callable.rs
a-frantz: That should be possible. Another good thought.
I think if we're going down a custom sort order for `input_groupts`, `Common` should come first, and `Resources` should come last? Everything else would just be alphabetical between those two?

wdl-doc/src/callable.rs
a-frantz: Actually this isn't so straightforward. I played around with it for a bit locally, and ran into other issues. I think we have to do it like this 🤷

wdl-doc/src/callable.rs
```diff
@@ -0,0 +1,295 @@
...
+    fn input_groups(&self) -> BTreeSet<Group> {
+        self.inputs()
+            .iter()
+            .filter_map(|param| param.group())
+            .map(|arg0: Group| Group(arg0.0.clone()))
+            .collect()
+    }
+
+    /// Get the inputs of the callable that are part of `group`.
+    fn inputs_in_group<'a>(&'a self, group: &'a Group) -> impl Iterator<Item = &'a Parameter> {
```
claymcleod, line 98: For this and the next method, there's actually a cool method I wasn't aware of called [`partition()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition) that can partition the inputs into "grouped" or "non-grouped" in one call. I'd consider using that.

claymcleod: I'm unclear if this is easy with what's going on with required. So it's more just a thing to consider.

wdl-doc/src/callable.rs
```diff
@@ -0,0 +1,295 @@
...
+            let name = decl.name().as_str().to_owned();
+            let meta = parameter_meta.get(&name).or_else(|| output_meta.get(&name));
+            Parameter::new(
+                wdl_ast::v1::Decl::Bound(decl.clone()),
+                meta.cloned(),
+                InputOutput::Output,
+            )
+        })
+        .collect()
+}
```
claymcleod, line 295: Is there a way to add some simple tests for this?

wdl-doc/src/callable/task.rs
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    fn inputs(&self) -> &[Parameter] {
+        &self.inputs
+    }
+
+    fn outputs(&self) -> &[Parameter] {
+        &self.outputs
+    }
+}
```
claymcleod, line 148: Needs simple tests.

wdl-doc/src/callable/workflow.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    fn inputs(&self) -> &[Parameter] {
+        &self.inputs
+    }
+
+    fn outputs(&self) -> &[Parameter] {
+        &self.outputs
+    }
+}
```
claymcleod, line 136: Needs simple tests.

wdl-doc/src/docs_tree.rs
```diff
@@ -0,0 +1,367 @@
...
+        &self.name
+    }
+
+    /// Get the type of the page.
+    pub fn page_type(&self) -> &PageType {
+        &self.page_type
+    }
+}
+
+/// A node in the DOCS directory tree.
```
claymcleod, line 58:
```suggestion
/// A node in the docs directory tree.
```

wdl-doc/src/docs_tree.rs
```diff
@@ -0,0 +1,367 @@
...
+    /// The root of the tree.
+    ///
+    /// `root.path` is the path to the DOCS directory and is absolute.
+    root: Node,
+    /// The absolute path to the stylesheet, if it exists.
+    stylesheet: Option<PathBuf>,
+}
+
+impl DocsTree {
+    /// Create a new DOCS tree.
```
claymcleod, line 126:
```suggestion
    /// Create a new docs tree.
```

wdl-doc/src/docs_tree.rs
```diff
@@ -0,0 +1,367 @@
...
+        }
+
+        let mut nodes = Vec::new();
+        recurse_depth_first(self, &mut nodes);
+
+        nodes
+    }
+}
+
+/// A tree representing the DOCS directory.
```
claymcleod, line 114: The `DOCS` thing needs to be universally searched for and replaced—seems like an errant find and replace maybe.

wdl-doc/src/lib.rs
```diff
@@ -6,59 +6,162 @@
...
+        let mut options = Options::empty();
+        options.insert(Options::ENABLE_TABLES);
+        options.insert(Options::ENABLE_STRIKETHROUGH);
+        let parser = Parser::new_ext(self.0.as_ref(), options);
+        pulldown_cmark::html::push_html(&mut unsafe_html, parser);
+        // Sanitize it with ammonia
+        let safe_html = ammonia::clean(&unsafe_html);
+
+        // Remove the outer `<p>` tag that `pulldown_cmark` wraps single lines in
+        let safe_html = if safe_html.starts_with("<p>") && safe_html.ends_with("</p>\n") {
```
claymcleod, line 111: This needs a test, both the positive and negative case.

wdl-doc/src/lib.rs
```diff
@@ -67,282 +170,198 @@ impl Document {
...
+                        }}
+                        tbody class="border" {
+                            @for page in &self.local_pages {
+                                tr class="border" {
+                                    td class="border" {
+                                        a href=(page.0.to_str().unwrap()) { (page.1.name()) }
+                                    }
+                                    td class="border" {
+                                        @match page.1.page_type() {
+                                            PageType::Index(_) => { "TODO ERROR" }
```
claymcleod, line 207: Can this not be a literal `todo!()`?

wdl-doc/src/lib.rs
```diff
@@ -67,282 +170,198 @@ impl Document {
...
-            .file_name()
-            .expect("current directory should have a file name")
-            .to_string_lossy();
         let ast_doc = result.document().node();
         let version = ast_doc
             .version_statement()
-            .expect("document should have a version statement")
-            .version();
-        let preamble = fetch_preamble_comments(ast_doc.clone());
+            .expect("Document should have a version statement");
```
claymcleod, line 288:
```suggestion
            .expect("document should have a version statement");
```

These expect statements should be all lowercase to follow the existing convention (unless you're starting with a proper noun, like `WDL`).

claymcleod, line 288:
```suggestion
            .expect("document should have a version statement");
```

These expect statements should be all lowercase to follow the existing convention (unless you're starting with a proper noun, like `WDL`).

claymcleod: Please double check all the expect statements here.

wdl-doc/src/lib.rs
```diff
@@ -67,282 +170,198 @@ impl Document {
...
+        let mut local_pages = Vec::new();
 
         for item in ast.items() {
             match item {
                 DocumentItem::Struct(s) => {
-                    let struct_name = s.name().as_str().to_owned();
-                    let struct_file = cur_dir.join(format!("{}-struct.html", struct_name));
-                    let mut struct_file = tokio::fs::File::create(struct_file).await?;
+                    let name = s.name().as_str().to_owned();
+                    let path = cur_dir.join(format!("{}-struct.html", name));
```
claymcleod, line 297: It might be a minor thing, but I would advocate for `struct-{}.html`, since that more closely follows the hierarchy here.

claymcleod (commented):
I can't approve the PR, as I'm the one who made it, but I'm generally good with this provided these are addressed.

wdl-doc/src/lib.rs
a-frantz: Unfortunately, that doesn't work within the `maud` macro. Trying to write `unreachable!()` or other such panic gets the entire macro triggered as "unreachable code block" by clippy. I saw some folks online using those panics, but they were also silencing the clippy warning.

wdl-doc/src/lib.rs
claymcleod: Okay that's fine I suppose. I definitely think that classifying it into buckets of struct/workflow/task is a higher-order ranking of the objects. But it would probably make more sense to do `/struct/<name`, `/task/<name>`, and `/workflow/<name>` in that case.

adthrasher: Somewhat related, I can see some repositories wanting the sidebar to look like:

- structs
-- struct 1
-- struct 2
- tasks
-- task 1
-- task 2

Rather than being grouped by source file. Clay's naming convention would make that easy to implement.

adthrasher (approved):

peterhuene:
Ugh. I need to look into authenticating CI so we don't hit the rate limit.

Also, probably need to implement caching of the check for image presence in crankshaft so that we don't unnecessarily query for the image again for the same run after ensuring it is present.

Issue #191, opened 2024-09-26
feature request: incorporate shellcheck for linting command blocks


a-frantz:
This feature will be a complete rip-off from miniwdl. It's my favorite feature of `miniwdl check` and I want us to copy it.

For completeness: link to [`shellcheck` site](https://www.shellcheck.net/)

It's been a while since I looked into this, but when I last looked there wasn't a good Rust crate for doing Bash linting. `miniwdl` shells out to `shellcheck` and parses the resulting STDOUT for a nice user interface. We are probably going to be stuck doing the same, which isn't ideal bc it's not very robust. But linting command blocks is a great feature I really want us to have, and it's worth getting our hands a little dirty IMO.

thatRichman:
Very selfishly interested in this. I might be able to take a stab at it if this is still desired. Would I be correct in thinking this would be more appropriate for sprocket now?

a-frantz:
We would certainly welcome that contribution! I've wanted this feature for a long time, but it hasn't made it up high enough in my priority list.

I think this code should live in the `wdl-lint` crate, not `sprocket`.

claymcleod:
Yeah I'd say this should live in `wdl-lint` as well. It's probably a "not on by default" lint that you have to opt in to. This will be quite a tricky one to implement gracefully—feel free to drop into the Zulip channel pointed to in the `README.md` to discuss with us!

thatRichman:
Sure, would be happy to discuss! Looks like the linked Zulip channel is invite only, though.

claymcleod:
Oh, thanks for pointing that out! I've updated the link with the invitation code—could you try again?

PR #264, opened 2024-12-08
feat: implement shellcheck lints in command blocks (#191)

thatRichman:

- **Rule Name**: `ShellCheckRule`

 I have implemented a lint rule that runs shellcheck against command sections and reports the comments as diagnostics. This is to address #191 .

Rule specific checks:

      is implemented.
- [ ] You have run `wdl-gauntlet --refresh` to ensure that there are no 
- [ ] You have run `wdl-gauntlet --refresh --arena` to ensure that all of the 
      (`Arena.toml`).


thatRichman:
There's still a lot of testing and polishing to do, which will undoubtedly reveal some issues, but I wanted to open for comments before I went any further in case I'm way off from what's desired.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+            SyntaxKind::VersionStatementNode,
+            SyntaxKind::TaskDefinitionNode,
+            SyntaxKind::CommandSectionNode,
+        ])
+    }
+}
+
+/// Convert a WDL placeholder to a bash variable
+fn to_bash_var(placeholder: &Placeholder) -> String {
+    let placeholder_text = placeholder.syntax().text().to_string();
```
a-frantz, line 155: `text().to_string()` can be expensive to call for complex expressions (it's a recursive traversal of the CST if I'm not mistaken). We might want to rethink this part. Maybe generate a random sequence of ASCII?

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+use crate::Rule;
+use crate::Tag;
+use crate::TagSet;
+use crate::util::{count_leading_whitespace, lines_with_offset, program_exists};
+
+/// The shellcheck executable
+const SHELLCHECK_BIN: &str = "shellcheck";
+
+/// shellcheck lints that we want to suppress
+const SHELLCHECK_SUPPRESS: &[&str] = &[
```
thatRichman, line 42: This is a copy-paste of miniwdl's list. I think we can remove most of these since most were added to work around their use of dummy literals, which we're not (currently) doing.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+    "1072", // Unexpected..
+    "1083", // this {/} is literal
+    "2043", // this loop will only ever run once for constant value
+    "2050", // This expression is constant
+    "2157", // Argument to -n is always true due to literal strings
+    "2193", // The arguments to this comparison can never be equal
+];
+
+/// The identifier for the command section ShellCheck rule.
+const ID: &str = "CommandSectionShellCheck";
```
a-frantz, line 53: I think the rule ID should just be `ShellCheck`. The only context we'd ever call it is in the command section, so the prefix seems redundant to me.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    for decl in task.declarations() {
+        decls.insert(decl.name().as_str().to_owned());
+    }
+    decls
+}
+
+/// Creates a "ShellCheck lint" diagnostic from a ShellCheckComment
+fn shellcheck_lint(comment: &ShellCheckComment, span: Span) -> Diagnostic {
+    dbg!(
```
thatRichman, line 189: oops

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+        &mut self,
+        state: &mut Self::State,
+        reason: VisitReason,
+        section: &CommandSection,
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        if !program_exists(SHELLCHECK_BIN) {
```
a-frantz, line 228: Performance nit: this should be done once instead of at every command block.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                    // we need to save the var so we can suppress later
+                    decls.insert(bash_var.clone());
+                    let mut expansion = String::from("\"$");
+                    expansion.push_str(&bash_var);
+                    expansion.push('"');
+                    sanitized_command.push_str(&expansion);
+                }
+            });
+        } else {
+            // TODO add a diagnostic explaining why this is being skipped?
```
a-frantz, line 255: This case is that the command used mixed indentation. We may potentially still want to parse and lint it using SC? Or we can throw up our hands and say "you have a weird command block and we aren't SC linting it" 🤷

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+            return;
+        }
+
+        // get leading whitespace so we can add it to each span
+        // TODO there may be a more elegant way to do this
```
a-frantz, line 260: I don't understand why this needs to be done. We just stripped the whitespace in the prior block.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                    let inner_span = {
+                        let outer_span = rel_token.text_range().to_span();
+                        Span::new(
+                            outer_span.start() + leading_whitespace + start + comment.column - 1,
+                            comment.end_column - comment.column,
+                        )
+                    };
+                    state.exceptable_add(
+                        shellcheck_lint(&comment, inner_span),
+                        SyntaxElement::from(rel_token.clone()),
```
a-frantz, line 338: The `element` here can just be the `section`, as that's the lowest granularity we allow `except` comments at. As is, we're going to check a bunch of lower level ancestors that we know won't have a comment.

a-frantz (commented):
Looks like this is on the right track! Didn't go through it with a fine-tooth comb yet, but left some comments/Qs on things that jumped out at me.

wdl-lint/src/rules/command_shellcheck.rs
thatRichman: My code comment is poorly worded. This is just counting the leading whitespace of the first CommandPart so we can add the length to the span. SC is receiving the stripped version, so otherwise the span will be misplaced. It's not actually adding any whitespace back.

thatRichman: Duh, I see what you mean. We can get the span from the StrippedCommandPart instead and then it will align.

wdl-lint/Cargo.toml
```diff
@@ -16,6 +16,9 @@ wdl-ast = { path = "../wdl-ast", version = "0.9.0" }
 convert_case = { workspace = true }
 indexmap = { workspace = true }
 rowan = { workspace = true }
+serde_json = { workspace = true }
```
claymcleod, line 19: Nit: let's make these alphabetical (except for the `wdl-ast` package---we generally use the convention of putting our own, local packages first in the dependencies).

wdl-lint/src/lib.rs
```diff
@@ -84,6 +84,7 @@ pub trait Rule: Visitor<State = Diagnostics> {
 /// Gets the default rule set.
 pub fn rules() -> Vec<Box<dyn Rule>> {
     let rules: Vec<Box<dyn Rule>> = vec![
+        Box::<rules::CommandSectionShellCheck>::default(),
```
claymcleod, line 87: Nit: let's put this last so that it reflects the implementation order.

wdl-lint/src/rules/command_shellcheck.rs
claymcleod: Yeah I think this is good for now. It would be nice to make these configurable in the future. Perhaps you could add a `TODO` comment above it and/or an issue when this is merged?

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+use crate::TagSet;
+use crate::util::{count_leading_whitespace, lines_with_offset, program_exists};
+
+/// The shellcheck executable
+const SHELLCHECK_BIN: &str = "shellcheck";
+
+/// shellcheck lints that we want to suppress
+const SHELLCHECK_SUPPRESS: &[&str] = &[
+    "1009", // the mentioned parser error was in...
+    "1072", // Unexpected..
```
claymcleod, line 44: Should this be "Unexpected <something here>"? Or is the rule literally titled "Unexpected" (that's weird if so).

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+    "2043", // this loop will only ever run once for constant value
+    "2050", // This expression is constant
+    "2157", // Argument to -n is always true due to literal strings
+    "2193", // The arguments to this comparison can never be equal
+];
+
+/// The identifier for the command section ShellCheck rule.
+const ID: &str = "CommandSectionShellCheck";
+
+/// A ShellCheck comment
```
claymcleod, line 55:
```suggestion
/// A ShellCheck comment.
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+///
+/// The file and fix fields are ommitted as we have no use for them.
+#[derive(Clone, Debug, Deserialize)]
+struct ShellCheckComment {
```
claymcleod, line 59: If ShellCheck calls these "comments", I'd be okay with what you have. If not, I suggest we call it a "diagnostic" to keep it consistent with the other terminology we use here.

```suggestion
struct ShellCheckDiagnostic {
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                    }
+                    line_num += last_line;
+                }
+                CommandPart::Placeholder(_) => {
+                    on_same_line = true;
+                }
+            }
+        }
+
+        match run_shellcheck(&sanitized_command) {
```
claymcleod, line 310: What happens if the user enables the `ShellCheck` rule but doesn't have `shellcheck` installed? I would suggest adding a different check prior to this that outputs a diagnostic at the beginning of the document explaining that `shellcheck` must be installed (and how to do it).

EDIT: I now see what happens, and I've commented on this again in an above comment.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+            "-f",
+            "json",
+            "-e",
+            &SHELLCHECK_SUPPRESS.join(","),
+            "-",
+        ])
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .spawn()
+        .context("error calling shellcheck")?;
```
claymcleod, line 96:
```suggestion
        .context("spawning the `shellcheck` process")?;
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+    {
+        let mut proc_stdin = sc_proc
+            .stdin
+            .take()
+            .context("error obtaining stdin of shellcheck")?;
```
claymcleod, line 101:
```suggestion
            .context("obtaining the STDIN handle of the `shellcheck` process")?;
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+            "json",
+            "-e",
+            &SHELLCHECK_SUPPRESS.join(","),
+            "-",
+        ])
+        .stdin(Stdio::piped())
+        .stdout(Stdio::piped())
+        .spawn()
+        .context("error calling shellcheck")?;
+    {
```
claymcleod, line 97: Perhaps add a `DEBUG` log here telling the process id of the `shellcheck` process.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+        let mut proc_stdin = sc_proc
+            .stdin
+            .take()
+            .context("error obtaining stdin of shellcheck")?;
+        proc_stdin.write_all(command.as_bytes())?;
+    }
+
+    let output = sc_proc
+        .wait_with_output()
+        .context("error calling shellcheck")?;
```
claymcleod, line 107:
```suggestion
        .context("waiting for the `shellcheck` process to complete")?;
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+
+    // shellcheck returns exit code 1 if
+    // any checked files result in comments
+    // so cannot check with status.success()
+    match output.status.code() {
+        Some(0) | Some(1) => serde_json::from_slice::<Vec<ShellCheckComment>>(&output.stdout)
+            .context("failed to parse shellcheck output"),
```
claymcleod, line 114:
```suggestion
            .context("deserializing STDOUT from `shellcheck` process"),
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+        Some(code) => bail!("error calling shellcheck: exit code {}", code),
```
claymcleod, line 115:
```suggestion
        Some(code) => bail!("unexpected `shellcheck` exit code: {}", code),
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+        None => bail!("error calling check: process interrupted"),
```
claymcleod, line 116:
```suggestion
        None => bail!("the `shellcheck` process appears to have been interrupted"),
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+    }
+}
+
+/// Runs ShellCheck on a command section and reports violations
+#[derive(Default, Debug, Clone, Copy)]
+pub struct CommandSectionShellCheck;
```
claymcleod, line 122:
```suggestion
pub struct ShellCheckRule;
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+impl Rule for CommandSectionShellCheck {
+    fn id(&self) -> &'static str {
+        ID
+    }
+
+    fn description(&self) -> &'static str {
+        "Ensures that command blocks are free of ShellCheck violations."
+    }
+
+    fn explanation(&self) -> &'static str {
```
claymcleod, line 133: Perhaps provide a link to ShellCheck in parenthesis after the first introduction of the name.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+    fn exceptable_nodes(&self) -> Option<&'static [SyntaxKind]> {
+        Some(&[
+            SyntaxKind::VersionStatementNode,
+            SyntaxKind::TaskDefinitionNode,
+            SyntaxKind::CommandSectionNode,
+        ])
+    }
+}
+
+/// Convert a WDL placeholder to a bash variable
```
claymcleod, line 153:
```suggestion
/// Convert a WDL placeholder to a bash variable.
```

wdl-lint/src/rules/command_shellcheck.rs
claymcleod: Agree.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                } else {
+                    '_'
+                }
+            })
+            .collect::<String>(),
+    );
+    bash_var
+}
+
+/// retrieve all input and private declations for a task
```
claymcleod, line 172:
```suggestion
/// Retrieve all input and private declarations for a task.
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+fn task_declarations(task: &TaskDefinition) -> HashSet<String> {
```
claymcleod, line 173:
```suggestion
fn gather_task_declarations(task: &TaskDefinition) -> HashSet<String> {
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+    for decl in task.declarations() {
+        decls.insert(decl.name().as_str().to_owned());
+    }
+    decls
+}
+
+/// Creates a "ShellCheck lint" diagnostic from a ShellCheckComment
+fn shellcheck_lint(comment: &ShellCheckComment, span: Span) -> Diagnostic {
+    dbg!(
+        Diagnostic::note("shellcheck violations within a command")
```
claymcleod, line 190:
```suggestion
        Diagnostic::note("`shellcheck` reported the following diagnostic")
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+            .with_rule(ID)
+            .with_label(
+                format!("SC{}[{}]: {}", comment.code, comment.level, comment.message),
+                span,
+            )
+            .with_fix("address ShellCheck violations")
```
claymcleod, line 196:
```suggestion
            .with_fix("address the diagnostics as recommended in the message")
```

wdl-lint/src/rules/command_shellcheck.rs
claymcleod: I'm not sure this should be the behavior as well. I made a comment earlier (though it is further down in the code), but I think we should indicate to the user that they need to install `shellcheck` rather than silently return here.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+
+        if !program_exists(SHELLCHECK_BIN) {
+            return;
+        }
+
+        // collect declarations so we can ignore placeholder variables
+        let parent_task = section.parent().into_task().expect("parent is a task");
+        let mut decls = task_declarations(&parent_task);
+
+        // replace all placeholders in the command with
```
claymcleod, line 236:
```suggestion
        // Replace all placeholders in the command with
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+    ) {
+        if reason == VisitReason::Exit {
+            return;
+        }
+
+        if !program_exists(SHELLCHECK_BIN) {
+            return;
+        }
+
+        // collect declarations so we can ignore placeholder variables
```
claymcleod, line 232:
```suggestion
        // Collect declarations so we can ignore placeholder variables
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+        let parent_task = section.parent().into_task().expect("parent is a task");
+        let mut decls = task_declarations(&parent_task);
+
+        // replace all placeholders in the command with
+        // dummy bash variables
+        let mut sanitized_command = String::new();
```
claymcleod, line 238: Sanitizing the command feels like it should get its own function so as to not distract from the reading of this current one.

wdl-lint/src/rules/command_shellcheck.rs
claymcleod: If another diagnostic will catch it and spit out a response, I feel it's best for this rule to silently return. Once the user fixes _that_ diagnostic, then these will start showing up.

If we go this route, ot would also be good to document explicitly above the `if` statement that a `None` result there means there is mixed indentation.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                        )
+                    }
+                }
+            } else {
+                0
+            }
+        };
+
+        // Map each actual line of the command to its corresponding
+        // CommandPart and start position
```
claymcleod, line 284:
```suggestion
        // `CommandPart` and start position.
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                }
+                CommandPart::Placeholder(_) => {
+                    on_same_line = true;
+                }
+            }
+        }
+
+        match run_shellcheck(&sanitized_command) {
+            Ok(comments) => {
+                for comment in comments {
```
claymcleod, line 312: Diagnostics?

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                    // skip declarations that shellcheck is unaware of
+                    if comment.code == 2154
```
claymcleod, line 314: `2154` looks like somewhat of a magic number: it would be good to make it a constant with a descriptive name and, if that's not enough for a future reader to understand what's going on here, add more to the comment.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                CommandPart::Placeholder(_) => {
+                    on_same_line = true;
+                }
+            }
+        }
+
+        match run_shellcheck(&sanitized_command) {
+            Ok(comments) => {
+                for comment in comments {
+                    // skip declarations that shellcheck is unaware of
```
claymcleod, line 313:
```suggestion
                    // Skip declarations that shellcheck is unaware of.
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                    if comment.code == 2154
+                        && decls.contains(comment.message.split_whitespace().next().unwrap_or(""))
```
claymcleod, line 315: It might be good to assign what's in the parentheses to a variable first with a descriptive name. I'm not sure that this line is checking from first glance.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                        SyntaxElement::from(rel_token.clone()),
+                        &self.exceptable_nodes(),
+                    )
+                }
+            }
+            Err(e) => {
+                let command_keyword = support::token(section.syntax(), SyntaxKind::CommandKeyword)
+                    .expect("should have a command keyword token");
+                state.exceptable_add(
+                    Diagnostic::error("failed to run shellcheck on command block")
```
claymcleod, line 347:
```suggestion
                    Diagnostic::error("running `shellcheck` on command block")
```

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,357 @@
...
+                        .with_label(e.to_string(), command_keyword.text_range().to_span())
+                        .with_rule(ID)
+                        .with_fix("ensure shellcheck is installed or disable this lint"),
```
claymcleod, line 350: Does the check above for `command_exists()` not silently return if `shellcheck` isn't present? If so, I'm a bit confused on this one.

claymcleod: If we do keep this error message, `shellcheck` needs to be in backticks.

wdl-lint/src/util.rs
```diff
@@ -1,9 +1,20 @@
 //! A module for utility functions for the lint rules.
 
+use std::process::Command;
+use std::process::Stdio;
+
 use wdl_ast::AstToken;
 use wdl_ast::Comment;
 use wdl_ast::SyntaxKind;
 
+/// Count amount of leading whitespace in a str
```
claymcleod, line 10:
```suggestion
/// Counts the amount of leading whitespace in a string slice.
```

wdl-lint/src/util.rs
```diff
@@ -60,6 +71,17 @@ pub fn lines_with_offset(s: &str) -> impl Iterator<Item = (&str, usize, usize)>
     })
 }
 
+/// check if a program exists via `which`
```
claymcleod, line 74:
```suggestion
/// Check whether or not a program exists via `which`.
```

Note that you'll almost certainly have to do something different for Windows here.

wdl-lint/tests/lints/command-shellcheck-ok/source.errors
```diff
@@ -0,0 +1,49 @@
+note[SectionOrdering]: sections are not in order for task `test1`
+   ┌─ tests/lints/command-shellcheck-ok/source.wdl:7:6
```
claymcleod, line 2: You can except the rules that aren't relevant in these examples to make it just check the relevant ones.

claymcleod: My expectation from the name is that actually no diagnostics would be in this file.

claymcleod (changes requested):
Looking really good!

wdl-lint/src/rules/command_shellcheck.rs
thatRichman: yeah, copy-paste typo

wdl-lint/src/rules/command_shellcheck.rs
a-frantz: Yes, another diagnostic should fire to warn about the mixed indentation. I agree it's fine to silently return here (with a comment explaining this).

wdl-lint/src/rules/command_shellcheck.rs
thatRichman: It does silently return above if shellcheck isn't installed. The label for this diagnostic will be whatever anyhow error `run_shellcheck` returns. The `fix` message is just poor wording and should be updated.

wdl-lint/src/util.rs
thatRichman: Yeah, can provide an alternate implementation via cfg

wdl-lint/tests/lints/command-shellcheck-ok/source.errors
thatRichman: Agreed, just some messiness I haven't cleaned up yet.

wdl-lint/src/rules/command_shellcheck.rs
thatRichman: We can use a OnceLock here to check if it exists only once and also emit a diagnostic (once) that shellcheck needs to b e installed.

wdl-lint/src/rules/command_shellcheck.rs
thatRichman: Though, the rule is literally titled just "Unexpected" https://www.shellcheck.net/wiki/SC1072

wdl-lint/src/rules/command_shellcheck.rs
thatRichman: Not perfect, but users can add `#shellcheck disable=` directives right into the comment of a command section. Or use `SHELLCHECK_OPTS` to globally disable.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,361 @@
...
+            });
+        } else {
+            // This is the case where the command section contains
+            // mixed indentation. We silently return and allow
+            // the mixed indentation lint to report this.
+            return;
+        }
+
+        // get leading whitespace so we can add it to each span
+        let leading_whitespace = {
```
thatRichman, line 265: FYI this algorithm is broken in all but the simplest case. Placeholder placement (and spacing) relative to a shellcheck lint causes the diagnostic to be misplaced. Working on updating it.

wdl-lint/src/rules/command_shellcheck.rs
```diff
@@ -0,0 +1,361 @@
...
+                0
+            }
+        };
+
+        // Map each actual line of the command to its corresponding
+        // `CommandPart` and start position.
+        let mut line_map = HashMap::new();
+        let mut on_same_line = false;
+        let mut line_num = 0usize;
+        for part in section.parts() {
```
thatRichman, line 292: FYI this algorithm is broken in all but the simplest case. Placeholder placement (and spacing) relative to a shellcheck lint causes the diagnostic to be misplaced. Working on updating it.

wdl-lint/tests/lints/deprecated-placeholder-options-v1.1/source.errors
```diff
@@ -46,3 +46,11 @@ note[DeprecatedPlaceholderOption]: use of the deprecated `default` placeholder o
    │
    = fix: replace the `default` placeholder option with a call to the `select_first()` standard library function
 
+note[CommandSectionShellCheck]: `shellcheck` reported the following diagnostic
```
thatRichman, line 49: I think this must have been caused by the random variable starting with two digits. I'll update so that it always starts with a letter.

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -0,0 +1,387 @@
...
+                            text.span().start() + start + leading_whitespace - 1,
+                        );
+                        line_num += 1;
+                    }
+                }
+                CommandPart::Placeholder(_) => {
+                    on_same_line = true;
+                }
+            }
+        }
```
a-frantz, line 342: I don't think this algorithm will work if there's wonky leading whitespace?
e.g.
```
command <<<     # this is funky
         # more weirdness
echo ~{greeting}
     echo ~{name}
   >>>
```

Regardless, I think we need some more documentation for this. I would move this block to its own function, and write a doc comment explaining what it does in detail. Just skimming this, I'm not as clear on what's happening as I'd like.

wdl-lint/tests/lints/shellcheck-error/source.errors
```diff
@@ -0,0 +1,48 @@
+note[CommandSectionShellCheck]: `shellcheck` reported the following diagnostic
+   ┌─ tests/lints/shellcheck-error/source.wdl:18:10
+   │
+18 │       if [ -f "$broken"]
+   │          ^ SC1073[error]: Couldn't parse this test expression. Fix to allow more checks.
+   │
+   = fix: address the diagnostics as recommended in the message
```
a-frantz, line 7: When I use SC on the CL, it usually includes a URL to read more about the rule. The SC docs site is pretty great, and I think we want to include that link if possible (and eventually create _our own_ doc site that's similar).

a-frantz (commented):
This is looking great so far. Really excited to get this feature merged in 😎

wdl-lint/src/rules/shellcheck.rs
a-frantz: I think multi-line placeholders will also screw it up, if I'm reading correctly.
e.g.
```
command <<<
    python script.py ~{(
        if bool
        then complex
        else expression
    )} -more -flags -all-one-same-line
>>>
```

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -0,0 +1,387 @@
...
+            SyntaxKind::VersionStatementNode,
+            SyntaxKind::TaskDefinitionNode,
+            SyntaxKind::CommandSectionNode,
+        ])
+    }
+}
+
+/// Convert a WDL `Placeholder` to a bash variable declaration.
+///
+/// returns "WDL" + <placeholder length - 6> random alphnumeric characters.
```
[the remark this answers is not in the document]
thatRichman: It performs a saturating sub, so the at minimum it generates just "WDL" which is the same number of characters as '~{}'

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -0,0 +1,387 @@
...
+
+    fn explanation(&self) -> &'static str {
+        "ShellCheck (https://shellcheck.net) is a static analysis tool and linter for sh / bash. \
+        The lints provided by ShellCheck help prevent common errors and \
+        pitfalls in your scripts. Following its recommendations will increase \
+        the robustness of your command sections."
+    }
+
+    fn tags(&self) -> TagSet {
+        TagSet::new(&[Tag::Correctness, Tag::Style, Tag::Portability])
```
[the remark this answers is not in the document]
thatRichman: Shellcheck's `style` severity is disabled by default, but can be enabled with `-S style`. I chose to do this above which is why I added Tag::Style, but I can remove if desired.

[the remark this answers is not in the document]
a-frantz: Reviewing the "style" tests below, I think SC uses the term "style" differently than we do. In my mind, all the SC style lints are more accurately about clarity/succinctness. So IMO we remove the style tag here, but leave the SC style lints enabled.

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -0,0 +1,387 @@
...
+                    Diagnostic::error("running `shellcheck` on command section")
+                        .with_label(
+                            "could not find `shellcheck` executable.",
+                            command_keyword.text_range().to_span(),
+                        )
+                        .with_rule(ID)
+                        .with_fix("install shellcheck or disable this lint."),
+                    SyntaxElement::from(section.syntax().clone()),
+                    &self.exceptable_nodes(),
+                );
```
a-frantz, line 291: Making a comment to open the floor for discussion.

Is this how we want to handle the case that SC isn't installed?

If we want to create a diagnostic that `ShellCheck` couldn't be run, it should be a `note` I think. _Not_ an `error`. Errors should be reserved for broken WDL, and this isn't that. I could maybe see this at `warning` level? But I'd prefer `note` personally.

Putting myself in the shoes of a naive user, if I've never heard of ShellCheck I'm just disabling this lint going forward. After all, we are asking folks to install another dependency without offering them any guidance on how to do that. And it would potentially annoy me seeing this every time I forget to disable the shellcheck rule. I think we need some more here. A link to shellcheck.net? And a one sentence sales pitch on the benefit of the SC?

claymcleod: I thought we decided that ShellCheck was only going to be an opt-in rule? I think the behavior depends largely on that decision. I like making it opt-in because it delineates the behavior between silent failure because of the missing tool and just being turned off.

a-frantz: > I thought we decided that ShellCheck was only going to be an opt-in rule? I think the behavior depends largely on that decision. I like making it opt-in because it delineates the behavior between silent failure because of the missing tool and just being turned off.

Maybe I missed something then. But the complication with making it opt-in is that this would be the first rule like that, and would probably require some tinkering of the architecture. We don't have a config system yet, so how exactly do we "opt-in"? Is that OOS for this PR?

claymcleod: Well, I was thinking that this wouldn't be in the default set, and we would just add something in Sprocket that injects this rule in on a flag or similar. But, that being said, I'm fine with keeping it in the default set with no diagnostic on missing, so long as the TODO is marked in an issue (and, I would ask that, if we're close to having that working, we include it in the commit just commented out for now so it's easy to enable when we're ready).

thatRichman: I'll go ahead and leave it in the default set but remove the diagnostic message on missing. I'll just comment out the diagnostic as requested (will fix up the wording a bit based on other feedback).

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -237,23 +237,39 @@ fn sanitize_command(section: &CommandSection) -> Option<(String, HashSet<String>
...
+/// Maps each line as shellcheck sees it to its corresponding start position in the source.
+fn map_shellcheck_lines(section: &CommandSection) -> HashMap<usize, usize> {
+    let mut line_map = HashMap::new();
+    let mut line_num = 1;
+    let mut skip_next_line = false;
+    for part in section.parts() {
+        match part {
+            CommandPart::Text(ref text) => {
+                for (line, line_start, _) in lines_with_offset(text.as_str()) {
+                    // Add back leading whitespace that is stripped from the sanitized command.
```
thatRichman, line 249: this took an embarrassingly long time to figure out.

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -367,12 +351,18 @@ impl Visitor for ShellCheckRule {
                     let start = line_map
                         .get(&diagnostic.line)
                         .expect("shellcheck line corresponds to command line");
-                    let inner_span = {
-                        Span::new(
-                            start + diagnostic.column,
-                            diagnostic.end_column - diagnostic.column,
-                        )
+                    let len = if diagnostic.end_line > diagnostic.line {
```
thatRichman, line 354: need a test for multiline diagnostic. Have to figure out how to trigger one in shellcheck first.

wdl-lint/src/rules/shellcheck.rs
thatRichman: Concerns should be addressed with the corrected algorithm now committed. Tests now include a variety of weirdness and multiline placeholders.

thatRichman:
Rebased. Gauntlet in arena mode is giving me the following:

```
thread 'main' panicked at /home/spencer/projects/wdl/gauntlet/src/lib.rs:258:25:
duplicate diagnostic: `ww-fastq-to-cram.wdl:101:20: note[ShellCheck]: `shellcheck` reported the following diagnostic` at /ww-fastq-to-cram.wdl:101
```
 
 so I'll look into that tomorrow.

a-frantz:
@thatRichman do you think you can figure out how to install `shellcheck` on the Arena CI runner? I'd rather not merge this until we can see the lints that will show up "in the wild". I won't be surprised if we find another SC rule or two should be disabled as being unhelpful in the context of a WDL command block.

Feel free to ping me on Zulip if that poses any issues! I imagine the windows runner could pose a challenge 🤔

thatRichman:
> @thatRichman do you think you can figure out how to install `shellcheck` on the Arena CI runner? I'd rather not merge this until we can see the lints that will show up "in the wild". I won't be surprised if we find another SC rule or two should be disabled as being unhelpful in the context of a WDL command block.
> 
> Feel free to ping me on Zulip if that poses any issues! I imagine the windows runner could pose a challenge 🤔

@a-frantz  Sure, shouldn't be too bad. Shellcheck provides a pre-built windows binary that we can pull from their releases. Might just take a little fiddling.

thatRichman:
@a-frantz  Here's a [link](https://github.com/thatRichman/wdl/actions/runs/12309212103/job/34355868925) to an example of a new install-shellcheck action running on my fork.

wdl-lint/src/util.rs
```diff
@@ -24,6 +34,22 @@ pub fn is_inline_comment(token: &Comment) -> bool {
     false
 }
 
+/// Determines if a string containing embedded quotes is properly quoted.
```
thatRichman, line 37:
```suggestion
/// Determines whether or not a string containing embedded quotes is properly quoted.
```

claymcleod:
> @a-frantz Here's a [link](https://github.com/thatRichman/wdl/actions/runs/12309212103/job/34355868925) to an example of a new install-shellcheck action running on my fork. One thing to note is Ubuntu being Ubuntu only has 0.8.0 by default. macOS and Windows will always run latest. Maybe a good thing to be testing across versions. But if we want to forego the apt caching we could just pull from GH the same as Windows.
> 
> edit: it's also suspicious that Arena passed on Windows when the duplicate diagnostic bug was still present. Makes me think it probably isn't finding the executable.

My two cents: let's just always pull from latest on everything (including Mac). It makes it a bit more straightforward to maintain the Github action associated with that, and it also ensures that every platform always works with the bleeding edge Shellcheck.

thatRichman:
I'm starting to see why the miniwdl folks went with literals instead of variables / expansions when substituting placeholders.

The common false positives I'm seeing from Arena are:

[SC2076](https://www.shellcheck.net/wiki/SC2076)
```
❌ read_group.wdl:200:43: note[ShellCheck]: Remove quotes from right-hand side of =~ to match as a regex rather than literally.
```
caused by:
```
| [[ ~{read_group.ID} =~ ~{restrictive_pattern} ]]
```
because we're replacing both of these placeholders with double-quoted expansions.

and [SC2001](https://www.shellcheck.net/wiki/SC2001)
```
❌ util.wdl:132:9: note[ShellCheck]: See if you can use ${variable//search/replace} instead.
```
caused by:
```
echo ~{string} | sed 's/~{delimiter}/\n/g' > split_strings.txt
```
because with a bash variable you can use the suggested syntax but that obviously doesn't apply to WDL placeholders.

as well as [SC2066](https://www.shellcheck.net/wiki/SC2066):
```
❌ samtools.wdl:577:21: note[ShellCheck]: Since you double quoted this, it will not word split, and the loop will only run once.
```
caused by:
```
 for file in ~{sep(" ", bams)}
```
Though, this one is also basically identical to one that miniwdl has to suppress.

I've found (what I think) is an elegant solution to determine whether to use a literal or expansion, which reduces but does not eliminate false positives.

That being said, I think miniwdl's approach of guessing the placeholder's type and using an appropriate literal would be overly complicated here. So I think it comes down to the severity and frequency of the lints we end up needing to suppress. I still think (with the commit I'm about to push) that we're better off than always using literals. We could probably spend weeks working to get the false positives to an absolute minimum without suppressing anything, but that feels a bit bikesheddy to me. Of course I will defer.

a-frantz:
> I'm starting to see why the miniwdl folks went with literals instead of variables / expansions when substituting placeholders.

So am I. However with `wdl-lint` currently being based off of `wdl-ast`, we don't have a good way of knowing the right type of literal to inject. We've discussed potentially refactoring `wdl-lint` to be based off of `wdl-analysis` so that we'd have access to resolved type information. That refactor would certainly be OOS for this PR, but would potentially simplify this problem by a great amount.

I'm thinking we shouldn't expend a large amount of effort eliminating these false positives with the current AST-based implementation.

@thatRichman can you remove this rule from the default set of rules (as Clay originally suggested) so that this is opt-in only?

We should re-architect `wdl-lint` so that it can have the results of a full static-analysis and revisit this rule at that time. In the meantime, this rule can be in a bit of a beta stage. It's certainly still useful and worth merging into `main`, but the false-positive lints are too big an issue to include this in the default rule set IMO.

thatRichman:
Sounds good. I'll disable the shellcheck-install action (once I get it working on all OSes) so that it be re-enabled at a later date. 

I know the intent is to support non-default rules in sprocket eventually, but should I put any work into supporting non-default rules in the dev CLI in this PR? Seems OOS, so happy to throw in an issue if there isn't one already. 

Also, I think I will wait for the resolution of #268 before making any final changes to this PR, as it throws off diagnostic placements.

a-frantz:
> I know the intent is to support non-default rules in sprocket eventually, but should I put any work into supporting non-default rules in the dev CLI in this PR? Seems OOS, so happy to throw in an issue if there isn't one already.

I think it should be straightforward to handle this as a special case (by adding a `--shellcheck` flag to the `wdl lint` dev command), but I wouldn't call it necessary for merging. You could leave that change out and open an issue as well 🤷 your call.

thatRichman:
@a-frantz  This should be ready for final review. One question I have is whether or not to run gauntlet with shellcheck enabled (I added this option), since this rule is in 'beta' and thus gauntlet will report missing lints if not run with --shellcheck (if it has previously been blessed while using --shellcheck). 

Also, I fully intend to squash these 39 commits into 1. I can do that before or after final review, whatever you prefer.

a-frantz:
Sweetness! Will give this a review in the morning. I also added the rest of the team as reviewers in case I miss something.

> One question I have is whether or not to run gauntlet with shellcheck enabled (I added this option), since this rule is in 'beta' and thus gauntlet will report missing lints if not run with --shellcheck (if it has previously been blessed while using --shellcheck).

IMO we leave it out of gauntlet/arena for now. I imagine it is going to add a **boatload** of diagnostics, and we like to review each one. Considering we know upfront some portion will be false positives, I don't think it's worth the effort.

> Also, I fully intend to squash these 39 commits into 1. I can do that before or after final review, whatever you prefer.

We squash as part of the merge, so you shouldn't need to do anything to deal with it.

gauntlet/src/lib.rs
a-frantz, line 1: Can you add an entry in `gauntlet/CHANGELOG.md` for this?

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -0,0 +1,424 @@
...
+                //     Diagnostic::note("running `shellcheck` on command section")
+                //         .with_label(
+                //             "could not find `shellcheck` executable.",
+                //             command_keyword.text_range().to_span(),
+                //         )
+                //         .with_rule(ID)
+                //         .with_fix("install shellcheck (https://www.shellcheck.net) or disable this lint."),
+                //     SyntaxElement::from(section.syntax().clone()),
+                //     &self.exceptable_nodes(),
+                // );
```
a-frantz, line 367: This should be enabled now.

wdl/src/bin/wdl.rs
a-frantz, line 1: Please add a `wdl/CHANGELOG.md` entry for this

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -0,0 +1,424 @@
...
+
+/// Creates a "ShellCheck lint" diagnostic from a `ShellCheckDiagnostic`
+fn shellcheck_lint(diagnostic: &ShellCheckDiagnostic, span: Span) -> Diagnostic {
+    let label = format!(
+        "SC{}[{}]: {}",
+        diagnostic.code, diagnostic.level, diagnostic.message
+    );
+    Diagnostic::note(&diagnostic.message)
+        .with_rule(ID)
+        .with_label(label, span)
```
a-frantz, line 206: This repeats the `diagnostic.message` twice. Is that desirable? Should we omit it from the `label` as being redundant, or do we like it appearing twice?

a-frantz:
@thatRichman can you add a test that `#shellcheck disable=` directives work as expected?

wdl-lint/src/rules/shellcheck.rs
adthrasher: I'm conflicted. Pulling an example from below, I think the initial instance by the `note` is fine. The second provides the additional info link, which could be helpful. I do notice that the span isn't highlighted correctly. I also think the `fix` needs to be something else.
```
note[ShellCheck]: You need a space before the ].
   ┌─ tests/lints/shellcheck-error/source.wdl:18:25
   │
18 │       if [ -f "$broken"]
   │                         ^
   │                         
   │                         SC1020[error]: You need a space before the ].
   │                         more info: https://www.shellcheck.net/wiki/SC1020
   │
   = fix: address the diagnostic as recommended in the message
```

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -0,0 +1,424 @@
...
+        .with_label(
+            format!("more info: {}/SC{}", &SHELLCHECK_WIKI, diagnostic.code),
+            span,
+        )
+        .with_fix("address the diagnostic as recommended in the message")
```
adthrasher, line 211: @a-frantz - I know you just added/updated our `fix` messages. What do you think of this one? The generic message isn't helpful to users, but the alternative would presumably involve parsing the warnings from shellcheck and writing a fix for each type.

wdl-lint/src/rules/shellcheck.rs
thatRichman: Re: the message

@peterhuene  and I discussed this some in the Zulip. This initially came about because Gauntlet requires all diagnostic messages to be unique. The solution I came up with was to include the SC###[<severity>] in the message, but only the message in the label. But happy to rework. Agree the fix could be much more useful.

wdl-lint/src/rules/shellcheck.rs
a-frantz: To be perfectly honest, yall didn't really convince me that it's the right move for every diagnostic to include a `fix`. I was just outvoted and didn't feel strongly enough to put up a fight 😅 

So my frank opinion is that this `fix` should be dropped _along with some of our other "unhelpful" fix messages_. But we've set the precedent that every diagnostic should include some kind of `fix`, and I'm not sure how we can improve this one without some rather intense work 🤷 

@thatRichman I see [here](https://github.com/stjude-rust-labs/wdl/pull/264/files#diff-9cabcff67f58680d419fec1e45eb2ac94d72365b16534b9351744b65ffa370f1R69) that `shellcheck` provides it's own `fix` message. Maybe we should just copy that? I haven't seen what they look like, so maybe they aren't of good quality 🤔 
Can you share some examples of the shellcheck fix messages?

thatRichman: The fix messages describe the structural changes required to address the lint, and are not always available.
Example:
```
    "fix": {
      "replacements": [
        {
          "line": 4,
          "endLine": 4,
          "precedence": 7,
          "insertionPoint": "afterEnd",
          "column": 11,
          "replacement": "\"",
          "endColumn": 11
        },
        {
          "line": 4,
          "endLine": 4,
          "precedence": 7,
          "insertionPoint": "beforeStart",
          "column": 15,
          "replacement": "\"",
          "endColumn": 15
        }
      ]
```

We _could_ convert these into a human-readable message, but I can't say if we want to.

wdl-lint/src/rules/shellcheck.rs
adthrasher: So it's the underlying `shellcheck` that returns the column position after `]`. That's good to know. It still seems like the highlight should be one position earlier, so that it flags the `]`.

wdl-lint/tests/lints/shellcheck-error/source.errors
```diff
@@ -0,0 +1,66 @@
...
+note[ShellCheck]: You need a space before the ].
+   ┌─ tests/lints/shellcheck-error/source.wdl:18:25
+   │
+18 │       if [ -f "$broken"]
+   │                         ^
+   │                         
+   │                         SC1020[error]: You need a space before the ].
+   │                         more info: https://www.shellcheck.net/wiki/SC1020
+   │
+   = fix: address the diagnostic as recommended in the message
```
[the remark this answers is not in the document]
thatRichman: Indeed. We had discussed collapsing the lints into a single diagnostic with one label per lint (example in the Zulip) but it's significantly less legible especially when including the wiki links.

wdl-lint/src/rules/shellcheck.rs
adthrasher: @a-frantz - This could be used to get toward my goal for `fix`. I would want to use that `fix` to print the correct line. 
```
echo [ -z "$foo" ]
```

a-frantz: Oo. That's kind of messy... At the very least it wouldn't be trivial to convert that to something I would want us to print.

@adthrasher I agree that we should work towards having every `fix` message print the "corrected" code where possible. But that's not very easy to implement robustly. This one in particular looks tough to parse _when it's present_.

My two cents on this issue is we leave the `fix` message as Spencer has written it, but we can make a new issue to revisit this specific `fix` (and many of our other fixes for that matter).

adthrasher: @a-frantz - That particular example doesn't look too bad to convert to text. You've just got the line, column, replacement text, and how to place it. The bigger issue, to me, is that it isn't always present (though some errors probably aren't straightforward like this one). But I agree, we can leave it as-is and have a new issue to revisit.

wdl-lint/src/lib.rs
```diff
@@ -150,3 +152,32 @@ pub fn rules() -> Vec<Box<dyn Rule>> {
 
     rules
 }
+
+/// Gets the optional rule set.
+pub fn optional_rules() -> Vec<Box<dyn Rule>> {
```
thatRichman, line 157: this is probably overkill since this will (hopefully) be a temporary home for shellcheck.

gauntlet/src/lib.rs
```diff
@@ -185,6 +190,9 @@ pub async fn gauntlet(args: Args) -> Result<()> {
                 if args.arena {
                     validator.add_visitor(LintVisitor::default());
                 }
+                if args.shellcheck {
```
a-frantz, line 193: This should only be added in arena mode.

a-frantz:
We're pretty dang close here @thatRichman . Please get the CI all green, then I think this will have my approval
https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md#the-ci-has-turned-red-how-do-i-make-it-green-again

The general rule we follow is to require approval from two core team members for a merge.

thatRichman:
@a-frantz  Sounds good. One thing to note is that I did end up needing to re-enable shellcheck install in the CI. This is because if shellcheck isn't present, the first lint test that is run will end up with the 'shellcheck not found' warning, and the source.errors for the shellcheck-* lints will be empty. I thought about modifying the lints test script to selectively skip the shellcheck lints, but this would not address the 'missing shellcheck' lint. 

Here is an example of the modified CI [going green on my repo](https://github.com/thatRichman/wdl/actions/runs/12421364066/job/34680892250?pr=3) but I don't think there's any way to reflect that here unless we merge the actions changes to main in a separate PR first.

gauntlet/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* `--shellcheck` flag to run shellcheck lints in both standard and arena modes ([#264](https://github.com/stjude-rust-labs/wdl/pull/264))
```
a-frantz, line 12:
```suggestion
* `--shellcheck` flag to run shellcheck lints in arena mode ([#264](https://github.com/stjude-rust-labs/wdl/pull/264))
```

wdl-lint/src/rules/shellcheck.rs
a-frantz: This thread got a little sidetracked. Bringing up the original Q again, are we ok with the SC message being repeated? Maybe I'm lacking in imagination here, but I don't really see a way to get rid of the duplication that still looks ok.

I like it at the top of the diagnostic and that ensures uniqueness for gauntlet. Removing it from the label leaves that part lacking... I'm stumped.
I'm ok with it appearing twice, but that's mostly because I can't think of an alternative I like. Open to any ideas here.

wdl-lint/tests/lints/shellcheck-ok/source.wdl
```diff
@@ -0,0 +1,79 @@
...
+        echo "all is well"
+      fi
+    }
+
+    output {}
+
+    runtime {}
+}
+
+task test3 {
```
a-frantz, line 56: Can we add a `test4` that uses the shellcheck suppression comment _within_ the command block? e.g.
```
command <<<
    #shellcheck-disable=SCfoo
    if [[ $im_a_bad_example ]]; then
       echo bar
   fi
>>>
```

wdl-lint/tests/lints/shellcheck-ok/source.wdl
```diff
@@ -77,3 +77,28 @@ task test3 {
...
+    meta {}
+
+    parameter_meta {}
+
+    input {
+      Int placeholder
+    }
+
+    command {
+      #@ except: ShellCheck
```
a-frantz, line 91: Ah sorry, I was unclear. Pulled the syntax from memory instead of looking it up 😅 

I want to make sure that using _the shellcheck directives_ works, as explained on this page - https://github.com/koalaman/shellcheck/wiki/Ignore#ignoring-one-specific-instance-in-a-file

So please remove _our directive_ and instead use the SC specific syntax. A little confusing here lol.

thatRichman: Ahh that makes more sense. will fix!

wdl-lint/src/rules/shellcheck.rs
adthrasher: I'm fine with the duplication. We can always revisit it in the future.

a-frantz (approved):
LGTM!

adthrasher (approved):

PR #265, opened 2024-12-09
feat: implement WDL 1.2 task evaluation.

peterhuene:
This commit adds evaluation of WDL tasks with full 1.2 support.

It completes 1.2 expression evaluation except for accessing call values; those
will be implemented along with workflow evaluation.

It includes several big refactorings to `wdl-analysis` to facilitate in
evaluation, namely storing diagnostics, id, and uri in `Document` instead of
`AnalysisResult`.

Evaluation of string and command literals now properly unescape any escape
sequences.

Task evaluation tests were implemented and derived from examples taken directly
from the WDL 1.2 spec.

Storage of compound types has been changed such that empty compound types (e.g.
`[]`, `{}`, `object {}`) do not cause an allocation.

Whitespace stripping for multiline strings and commands has been refactored to
reduce the number of allocations made.

      changes (when appropriate).


peterhuene:
Note: I haven't tested this on Windows yet, so expect CI failures there.

I plan on pushing up more file tests before merging this.

wdl-engine/src/backend/local.rs
```diff
@@ -0,0 +1,285 @@
...
+            .get("cpu")
+            .map(|v| {
+                v.coerce(engine.types(), PrimitiveTypeKind::Float.into())
+                    .expect("type should coerce")
+                    .unwrap_float()
+            })
+            .unwrap_or(1.0);
+        if num_cpus < min_cpu {
+            bail!(
+                "task requires at least {min_cpu} CPU{s}, but only {num_cpus} CPU{s2} are \
```
claymcleod, line 126: This should be "is" if `s2` is `1.0`—not "are".

wdl-engine/src/backend/local.rs
```diff
@@ -0,0 +1,285 @@
...
+                }
+
+                unreachable!("value should be an integer or string");
+            })
+            .transpose()?
+            .unwrap_or(1);
+        if memory < min_memory {
+            bail!(
+                "task requires at least {min_memory} byte{s} of memory, but only {memory} \
+                 byte{s2} are available",
```
claymcleod, line 160: This has to be "gigabytes" right? And same comment as above on the "is" vs. "are".

wdl-engine/src/backend/local.rs
```diff
@@ -0,0 +1,285 @@
...
+                        .and_then(|v| v.try_into().ok())
+                        .with_context(|| {
+                            format!("task specifies an invalid `memory` requirement `{s}`")
+                        });
+                }
+
+                unreachable!("value should be an integer or string");
+            })
+            .transpose()?
+            .unwrap_or(1);
```
claymcleod, line 156: Should the default be 2 GiB?

wdl-engine/src/eval.rs
```diff
@@ -149,3 +201,155 @@ impl<'a> ScopeRef<'a> {
...
+                    }
+                }
+                Value::Compound(CompoundValue::Array(codes)) => {
+                    error = !codes.as_slice().iter().any(|v| {
+                        v.as_integer()
+                            .map(|i| i32::try_from(i).unwrap_or_default() == self.status_code)
+                            .unwrap_or(false)
+                    });
+                }
+                _ => unreachable!("unexpected return codes value"),
```
claymcleod, line 338: I'm assuming `Value` implements `Display`, in which case, would it be wise to print out the exact text that isn't as expected here?

wdl-engine/src/eval.rs
```diff
@@ -149,3 +201,155 @@ impl<'a> ScopeRef<'a> {
...
+    }
+
+    /// Handles the exit of a task execution.
+    ///
+    /// Returns an error if the task failed.
+    fn handle_exit(&self, requirements: &HashMap<String, Value>) -> anyhow::Result<()> {
+        let mut error = true;
+        if let Some(return_codes) = requirements
+            .get("return_codes")
+            .or_else(|| requirements.get("returnCodes"))
```
claymcleod, line 317: IMO, these kinds of keys should be `const`-ed.

wdl-engine/src/backend/local.rs
peterhuene: Yes, [it should be](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#memory). I'll fix.

wdl-engine/src/eval/v1/task.rs
```diff
@@ -0,0 +1,929 @@
...
+                }
+
+                info!(
+                    "evaluating task `{task}` in `{uri}`",
+                    task = task.name(),
+                    uri = document.uri()
+                );
+
+                // Tasks only have a root scope (0), an output scope (1), and a `task` variable
+                // scope (2)
```
claymcleod, line 257: Needs completing.

wdl-engine/src/backend/local.rs
peterhuene: It's converted to bytes before this and I didn't bother converting the bytes to a "min unit greater than 1" for the error message, though.

I'll see what I can do to improve the error message.

wdl-engine/src/eval/v1/task.rs
```diff
@@ -0,0 +1,929 @@
...
+                                )
+                                .with_context(|| {
+                                    format!("failed to execute task `{task}`", task = task.name())
+                                })?;
+
+                            // Introduce the task variable at this point; valid for both the command
+                            // section and the outputs section
+                            if version >= SupportedVersion::V1(V1::Two) {
+                                let task =
+                                    TaskValue::new_v1(task.name(), "bar", &definition, constraints);
```
claymcleod, line 313: "bar"?

wdl-engine/src/eval.rs
peterhuene: We can print it out in the panic, but this should be `unreachable!` due to requiring an error-free static analysis for the task evaluation to even occur.

wdl-engine/src/eval/v1/task.rs
peterhuene: Oh whoops, I forgot to fix the placeholder for the task id; I'll correct that.

wdl-engine/src/stdlib/quote.rs
```diff
@@ -79,17 +77,21 @@ mod test {
...
             .as_array()
             .unwrap()
-            .elements()
+            .as_slice()
             .iter()
             .map(|v| v.as_string().unwrap().as_str())
             .collect();
-        assert_eq!(elements, [r#""1.0""#, r#""1.1""#, r#""1.2""#]);
+        assert_eq!(elements, [
+            r#""1.000000""#,
```
claymcleod, line 85: Is this desirable?

peterhuene: The precision? The spec requires six decimal places for displaying floats.

wdl-engine/src/value.rs
```diff
@@ -821,7 +1024,7 @@ impl PrimitiveValue {
                 match self.0 {
                     PrimitiveValue::Boolean(v) => write!(f, "{v}"),
                     PrimitiveValue::Integer(v) => write!(f, "{v}"),
-                    PrimitiveValue::Float(v) => write!(f, "{v:?}"),
+                    PrimitiveValue::Float(v) => write!(f, "{v:.6?}"),
```
claymcleod, line 1027: Ah, I see it's intentional here. Could you share the thought process behind being explicit here?

wdl-engine/src/value.rs
```diff
@@ -833,14 +1036,62 @@ impl PrimitiveValue {
...
+    }
 }
 
 impl fmt::Display for PrimitiveValue {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Self::Boolean(v) => write!(f, "{v}"),
             Self::Integer(v) => write!(f, "{v}"),
-            Self::Float(v) => write!(f, "{v:?}"),
+            Self::Float(v) => write!(f, "{v:.6?}"),
```
claymcleod, line 1094: Ah, I see it's intentional here. Could you share the thought process behind being explicit here?

wdl/src/lib.rs
```diff
@@ -78,6 +78,9 @@ pub use wdl_analysis as analysis;
 #[cfg(feature = "ast")]
 #[doc(inline)]
 pub use wdl_ast as ast;
+#[cfg(feature = "engine")]
+#[doc(inline)]
+pub use wdl_engine as engine;
 // TODO: uncomment this when wdl-engine is ready for release.
 // #[cfg(feature = "engine")]
```
claymcleod, line 85: This needs to be deleted.

claymcleod (changes requested):
I saw in an earlier version that you were evaluating and comparing the output of the commands explicitly too. I liked that being checked is well, is there a reason you had to move away from that?

wdl-engine/src/stdlib/quote.rs
claymcleod: That feels bad. Perhaps there is a good reason for it?

wdl-engine/src/eval.rs
claymcleod: Oh I see. Okay I think I glossed over that this was unreachable. No need to print it out then.

wdl-engine/src/backend/local.rs
claymcleod: Okay, I suppose this is tied in with the `unwrap_or()` above then—I was under the impression that represented 1 GiB, so if that represents 1 byte (and you're going to update to 2GiB), I get it.

It's _slightly_ better to report it as GiB if possible, but I don't think it's a show-stopper.

wdl-engine/src/stdlib/quote.rs
peterhuene: It's mentioned in the section covering how floats are [coerced to string in a placeholder](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#expression-placeholder-coercion).

Unfortunately no justification given.

peterhuene:
The commands ended up with temp file names in them sometimes; I can look at what `tempfile` does for the file name and do a regex replacement on it to something consistent.

peterhuene:
Whoops, need to strip more paths from the command files.

wdl-engine/src/eval/v1/task.rs
```diff
@@ -0,0 +1,929 @@
...
+
+                let requirements = requirements.unwrap_or_default();
+                let hints = hints.unwrap_or_default();
+
+                // TODO: check call cache for a hit. if so, skip task execution and use cache
+                // paths for output evaluation
+
+                let status_code = execution.spawn(command, &requirements, &hints)?.await?;
+
+                // TODO: support retrying the task if it fails
```
a-frantz, line 391: Ah I was wondering about this. I thought maybe I missed something. Glad to see I've been following along correctly 😅

wdl-engine/src/eval/v1/task.rs
```diff
@@ -0,0 +1,929 @@
...
+                            }
+                            TaskGraphNode::Decl(decl) => {
+                                self.evaluate_decl(
+                                    document,
+                                    execution.as_ref(),
+                                    &mut scopes,
+                                    task,
+                                    decl,
+                                )?;
+                            }
```
a-frantz, line 438: How can an input/decl follow execution? Shouldn't these have been evaluated already?

wdl-engine/src/value.rs
```diff
@@ -2052,6 +2590,339 @@ impl From<Struct> for CompoundValue {
...
+    /// The task's return code.
+    ///
+    /// Initially set to `None`, but set after task execution completes.
+    return_code: Option<i64>,
+    /// The task's `meta` section as an object.
+    meta: Object,
+    /// The tasks's `parameter_meta` section as an object.
+    parameter_meta: Object,
+    /// The task's extension metadata.
+    ext: Object,
```
a-frantz, line 2643: What's this? Something added in 1.2 that I'm not familiar with?

wdl-engine/src/stdlib/quote.rs
claymcleod: Okay, I'm satisfied at any rate.

wdl-engine/tests/tasks.rs
```diff
@@ -0,0 +1,486 @@
...
+//!
+//! This test looks for directories in `tests/tasks`.
+//!
+//! Each directory is expected to contain:
+//!
+//! * `source.wdl` - the test input source to evaluate; the file is expected to
+//!   contain no static analysis errors, but may fail at evaluation time.
+//! * `error.txt` - the expected evaluation error, if any.
+//! * `inputs.json` - the inputs to the task.
+//! * `outputs.json` - the expected outputs from the task, if the task run
```
a-frantz, line 11:
```suggestion
//! * `outputs.json` - the expected outputs from the task, if the task runs
```

wdl-engine/tests/tasks.rs
```diff
@@ -0,0 +1,486 @@
...
+            "result from `{path}` is not as expected:\n{diff}",
+            path = path.display(),
+            diff = StrComparison::new(&expected, &result),
+        );
+    }
+
+    Ok(())
+}
+
+/// Runts the test given the provided analysis result.
```
a-frantz, line 156:
```suggestion
/// Runs the test given the provided analysis result.
```

a-frantz (commented):
Haven't gone through all the tests yet. Or checked out the branch and kicked the tires of the new `run` command in the wdl bin, but here's my comments so far. Will come back with some more feedback later.

wdl-engine/src/eval/v1/task.rs
peterhuene: Not if they're either:

1. unused (i.e. have no outgoing dependency edges)
2. have only outgoing edges to outputs (i.e. the command and requirements/hints/runtime sections do not reference the declaration

In this case the input or declaration would be evaluated after the command in topological ordering.

wdl-engine/src/value.rs
peterhuene: It's from 1.2's section on [runtime access to task information](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#runtime-access-to-requirements-hints-and-metadata).

wdl-engine/src/eval.rs
```diff
@@ -149,3 +201,155 @@ impl<'a> ScopeRef<'a> {
...
+                _ => unreachable!("unexpected return codes value"),
+            }
+        } else {
+            error = self.status_code != 0;
+        }
+
+        if error {
+            bail!(
+                "task process has terminated with status code {code}; see standard error output \
+                 `{stderr}`",
```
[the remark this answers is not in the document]
peterhuene: Perhaps just include the path to the test's execution directory and mention that it contains the stdout, stderr, and working directory for the task? Otherwise it might be path overload.

wdl-engine/src/eval.rs
```diff
@@ -149,3 +201,155 @@ impl<'a> ScopeRef<'a> {
...
+        if let Some(return_codes) = requirements
+            .get("return_codes")
+            .or_else(|| requirements.get("returnCodes"))
+        {
+            match return_codes {
+                Value::Primitive(PrimitiveValue::String(s)) if s.as_ref() == "*" => {
+                    error = false;
+                }
+                Value::Primitive(PrimitiveValue::String(s)) => {
+                    bail!("invalid return code value `{s}`; only `*` is accepted");
```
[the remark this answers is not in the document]
peterhuene: How about:

`invalid return code value `{s}`: only `*` is accepted when the return code is specified as a string"

wdl-engine/src/eval.rs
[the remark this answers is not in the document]
a-frantz: Path overload is a valid concern, but on the other hand, I like being able to quickly doubleclick to select each path and paste it into my terminal. Maybe just the execution dir is enough though... I'm not sure.

Is stderr, stdout, work dir, and temp dir, all at the root of the execution dir? If so, I think probably just the exec dir is enough. If `ls <exec dir>` immediately gives me everything I'm looking for, it would be ok.

wdl-engine/src/eval.rs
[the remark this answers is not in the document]
a-frantz: Yea that seems good 👍

wdl-engine/src/eval.rs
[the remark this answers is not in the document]
peterhuene: They are all at the root of the execution dir (`stdout`, `stderr`, `command`, `work_dir`, and `temp_dir`).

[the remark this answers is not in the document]
a-frantz: Very nice. That's as it should be IMO, but there are unspecified runners out there that make tracking those down difficult by being nested/in weird directories, etc.

In this case, I would change the message to just print the execution dir path 🚀

wdl-engine/src/backend/local.rs
```diff
@@ -0,0 +1,285 @@
...
+    /// directory to use.
+    pub fn new(root: &Path) -> Result<Self> {
+        let root = absolute(root).with_context(|| {
+            format!(
+                "failed to determine absolute path of `{path}`",
+                path = root.display()
+            )
+        })?;
+
+        // Recreate the temp directory as it may be needed for task evaluation
```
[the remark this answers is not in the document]
peterhuene: We can skip recreating it and implement incrementing it on retry when retries are implemented. I'll remove for now.

wdl-engine/src/eval/v1/task.rs
a-frantz: Do they even need to be evaluated if they aren't used? Is that just wasted computation?

peterhuene: Technically, the unused ones don't, but those referenced only by the outputs definitely do need to be evaluated; I figure we should probably evaluate the unused ones anyway for errors; we already emit a warning informing users that they can be removed with the justification that it would, in theory, reduce evaluation time.

a-frantz: That makes sense. I suppose a decent compromise is what you have implemented here. If one of these unused decls causes an error but the task otherwise completed successfully, the results could be salvaged from the run dir.

Q: What about "silently used" inputs like `.bam.bai` files? Those might not be referenced anywhere in the command/output, but certainly need to be localized for successful task completion.

claymcleod (approved):
Looks great.

wdl-engine/src/eval/v1/task.rs
peterhuene: I think I'll need to revisit path mapping in the future as right now files need to be transitively referenced by the command for them to be mapped into a container environment; so while we have a heuristic to suppress an unused input declaration on an input named `foo_bai`, we won't actually mount the file into the container unless the command references it.

Note, though, that it'll work now if you use a struct value to group the related paths together, since we map all the struct members if the struct value is referenced by the command.

This will all need to be tweaked a bit once we're doing more than local execution and mounting of files is required.

peterhuene: Also note that initial workflow/task inputs that reference URLs isn't yet implemented, either. I can open a feature issue for it.

a-frantz: > I think I'll need to revisit path mapping in the future as right now files need to be transitively referenced by the command for them to be mapped into a container environment; so while we have a heuristic to suppress an unused input declaration on an input named `foo_bai`, we won't actually mount the file into the container unless the command references it.

Yea, I think we can move forward with this PR without that, but it will most certainly need to be addressed in a follow up. I think the answer here is to just always localize/mount `File` inputs? I suppose it could be a wasted computation, but it's what I would expect as a user.

> Note, though, that it'll work now if you use a struct value to group the related paths together, since we map all the struct members if the struct value is referenced by the command.

This is an interesting caveat. Thank you for sharing! With this snippet of info, I actually think this corner case will be quite rare in the wild. With the WDLs I've reviewed, I think it's most common to "package" a "main" file and their indices/accessory files in a Struct. `workflows` is in the minority in that we do _not_ use structs this way. We have separate inputs for "main" files and their accessory files, _but_ we always localize them with `ln -s` in the command block as a sanity check. That should be unnecessary, but we've run into issues with some runners not handling file localization properly.

> This will all need to be tweaked a bit once we're doing more than local execution and mounting of files is required.

Agree we can come back 👍 
In the meantime, can you open an issue to track this? Don't want this lost in the shuffle.

peterhuene: Opened https://github.com/stjude-rust-labs/wdl/issues/267.

wdl-analysis/CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Refactored the `AnalysisResult` and `Document` types to move properties of
+  the former into the latter; this will assist in evaluation of documents in
+  that the `Document` along can be passed into evaluation ([#265](https://github.com/stjude-rust-labs/wdl/pull/265)).
```
a-frantz, line 20: GH is hiding the `suggestion` button from me, but I think `along` should be `alone`

a-frantz:
```suggestion
  that the `Document` alone can be passed into evaluation ([#265](https://github.com/stjude-rust-labs/wdl/pull/265)).
```
Not sure why I couldn't do this before

a-frantz (approved):

PR #272, opened 2024-12-17
fix: conditionally strip chars from first line

a-frantz:

Applies the #271 fix to multiline strings as well as command sections.

      changes (when appropriate).

### DIFF
--- a/wdl-ast/src/v1/expr.rs
+++ b/wdl-ast/src/v1/expr.rs
@@ -2218,10 +2218,12 @@ impl LiteralString {
         }
 
         // Trim the first line
+        let mut whole_first_line_trimmed = false;
         if let Some(StrippedStringPart::Text(text)) = result.first_mut() {
-            let end = text.find('\n').map(|p| p + 1).unwrap_or(text.len());
-            let line = &text[..end];
+            let end_of_first_line = text.find('\n').map(|p| p + 1).unwrap_or(text.len());
+            let line = &text[..end_of_first_line];
             let len = line.len() - line.trim_start().len();
+            whole_first_line_trimmed = len == line.len();
             text.replace_range(..len, "");
         }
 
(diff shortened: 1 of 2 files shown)


wdl-ast/src/v1/expr.rs
```diff
@@ -7837,4 +7839,38 @@ task test {
...
+
+        let decls: Vec<_> = tasks[0].declarations().collect();
+        assert_eq!(decls.len(), 1);
+
+        let expr = decls[0].expr().unwrap_literal().unwrap_string();
+        let stripped = expr.strip_whitespace().unwrap();
+        assert_eq!(stripped.len(), 1);
+        match &stripped[0] {
+            StrippedStringPart::Text(text) => {
+                assert_eq!(text.as_str(), "hello world\n    my name is Jeff.")
```
a-frantz, line 7871: @peterhuene is this correct, or should it be `"hello world\nmy name is Jeff."`?

peterhuene (approved):

a-frantz (commented):
oops my comment didn't go through

wdl-ast/src/v1/expr.rs
peterhuene: From my reading of the spec, I _think_ this is the expected output.

According to the spec the order of operations is:

1. Remove all line continuations and subsequent white space.
2. Remove all whitespace following the opening <<<, up to and including a newline (if any).
3. Remove all whitespace preceeding the closing >>>, up to and including a newline (if any).
4. Use all remaining non-blank lines to determine the common leading whitespace. 
5. Strip leading whitespace.

I think we're now faithfully following these steps; we remove line continuations first, then the `<<<` and `>>>` whitespace, then we calculate the minimum leading whitespace from every non-blank line, then we strip the leading whitespace.

In this case, as we remove the whitespace on the first line in step 2, we determine the minimum leading whitespace to be 0 because `hello world` is the first non-blank line; as a result, the leading whitespace remains.

PR #277, opened 2024-12-26
refactor: remove `Types` collection from `wdl-analysis`. 

peterhuene:
This commit removes the `Types` collection from `wdl-analysis`.

The collection served as an arena to allocate types into which also served as providing type recursion via arena identifiers. The hope was that this would prevent a lot of small allocations from occurring during analysis when calculating expression types.

However, each document had a different `Types` collection, requiring "importing" types between collections. Further complicating things, an evaluation engine had its own `Types` collection which needed importing of types from other document collections. This was a source of bugs in the implementation of the evaluation engine.

Worse, the types collection served as a bottleneck for sharing data across asynchronous tasks in the upcoming implementation of workflow evaluation, requiring a lock.

The alternative is to use an `Arc` in appropriate places where type recursion occurs. After running some simple benchmarks, the removal of the types arena doesn't appear to have any real world impact on performance.

Additionally, the removal of `Types` allows `Type` to implement `PartialEq` rather than a custom trait that uses the arena; likewise for `Display`.

Similarly in `wdl-engine`, `Value` can now directly implement serialization and deserialization traits.

      changes (when appropriate).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,7 +35,6 @@ faster-hex = "0.9.0"
 futures = "0.3.30"
 git2 = "0.18.3"
 glob = "0.3.1"
-id-arena = "2.2.1"
 indexmap = { version = "2.2.6", features = ["serde"] }
 indicatif = "0.17.8"
 itertools = "0.13.0"
--- a/Gauntlet.toml
+++ b/Gauntlet.toml
@@ -1313,11 +1313,6 @@ permalink = "https://github.com/broadinstitute/palantir-workflows/blob/58f72308b
 
 [[diagnostics]]
 document = "broadinstitute/palantir-workflows:/BenchmarkSVs/BenchmarkSVs.wdl"
-message = "BenchmarkSVs.wdl:242:26: error: type mismatch: expected type `Array[String]`, but found type `Array[File]?`"
-permalink = "https://github.com/broadinstitute/palantir-workflows/blob/58f72308bdf511a760dea92dbf2ac21eb2c7dc0e/BenchmarkSVs/BenchmarkSVs.wdl/#L242"
-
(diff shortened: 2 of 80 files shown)


peterhuene:
No rush on this review as I'm out this week and next, but the workflow evaluation implementation will be relying on these changes so that we can more easily spawn asynchronous tasks that do WDL evaluation internally (i.e. we no longer have to synchronize a singular `Types` collection for things to work).

peterhuene:
Looks like gauntlet caught a bug for me to track down too.

peterhuene:
So the gauntlet diff is actually removing false positives fixed by the PR.

The actual bug was in the now-deleted `import` function responsible for importing a type between two different type collections (e.g. referencing an input for a task/workflow in an imported document). The bug was that the imported type accidentally dropped the optional specifier for the type, resulting in type coercion failures in task/workflow call statements.

For example:

bar.wdl:

```wdl
version 1.1

task foo {
    input {
        Array[String]? bar
    }

    command <<<>>>
}
```

foo.wdl:

```wdl
version 1.1

import "bar.wdl"

workflow bar {
    input {
        Array[String]? bar
    }
    
    call bar.foo { input: bar = bar }
}
```

Results in this false positive diagnostic:

```
error: type mismatch: expected type `Array[String]`, but found type `Array[String]?`
  ┌─ foo.wdl:9:33
  │
9 │     call bar.foo { input: bar = bar }
  │                           ---   ^^^ this is type `Array[String]?`
  │                           │      
  │                           this expects type `Array[String]`
```

I'll push a bless of the gauntlet diagnostics to rid the false positives and add a test case to ensure we don't accidentally drop the optional qualifier like this in the future.

PR #283, opened 2025-01-07
feat: simpler `wdl` entrypoints

a-frantz:

Adds functions to the `wdl` crate to simplify the API. These new functions are used by the `wdl` binary and will also be used in the upcoming `sprocket` release.

      changes (when appropriate).


wdl-analysis/src/lib.rs
```diff
@@ -20,5 +20,90 @@ mod rules;
...
 pub use analyzer::*;
+use anyhow::bail;
+use indicatif::ProgressBar;
+use indicatif::ProgressStyle;
 pub use rules::*;
+use tokio::fs;
+use url::Url;
+
+/// The delay in showing the progress bar.
+const PROGRESS_BAR_DELAY: Duration = Duration::from_secs(2);
```
claymcleod, line 34: I'd probably do every second here.

wdl-engine/CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added utility functions `run` and `validate_inputs` for a simpler API ([#283](https://github.com/stjude-rust-labs/wdl/pull/283)).
```
claymcleod, line 16: Wrap this.

wdl-engine/Cargo.toml
```diff
@@ -41,10 +42,6 @@ rayon = { workspace = true }
 [lints]
 workspace = true
 
-[features]
```
claymcleod, line 44: Why was this removed?

wdl-engine/src/lib.rs
```diff
@@ -10,10 +10,267 @@ mod stdlib;
...
+    stream: &mut codespan_reporting::term::termcolor::StandardStream,
+    config: &codespan_reporting::term::Config,
+) -> anyhow::Result<()> {
+    if Path::new(&document).is_dir() {
+        bail!("expected a WDL document, found a directory");
+    }
+
+    let results = analyze(&document, vec![], false, false).await?;
+
+    let uri = if let Ok(uri) = Url::parse(&document) {
```
claymcleod, line 52: This can be simplified as `Url::parse(&document).ok_or_else(...)`

wdl-analysis/src/lib.rs
a-frantz: Maybe this comment should be clarified. This is adding a 2 second delay _before_ the progress bar appears.

I found that we were analyzing files/directories so quickly that the inclusion of the progress bar actually looked pretty crummy because it would pop up and then instantly be dropped. So this essentially prevents the progress bar from showing up unless the analysis is taking a long time (most likely due to a slow network connection. That's really the only bottleneck here).

wdl-engine/src/lib.rs
```diff
@@ -10,10 +10,267 @@ mod stdlib;
...
+                .evaluate(document, task, &inputs, &output, &name)
+                .await
+            {
+                Ok(evaluated) => {
+                    match evaluated.into_result() {
+                        Ok(outputs) => {
+                            // Buffer the entire output before writing it out in case there are
+                            // errors during serialization.
+                            let mut buffer = Vec::new();
+                            let mut serializer = serde_json::Serializer::pretty(&mut buffer);
```
claymcleod, line 221: Can you just use [`to_string_pretty()`](https://docs.rs/serde_json/latest/serde_json/fn.to_string_pretty.html) here?

wdl-analysis/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+* Added utility function `analyze` to simplify the API ([#283](https://github.com/stjude-rust-labs/wdl/pull/283)).
```
claymcleod, line 12:
```suggestion
* Added `analyze` function to simplify the API ([#283](https://github.com/stjude-rust-labs/wdl/pull/283)).
```

claymcleod: In general, I find calling things "utility" functions or making "utility" modules to be not the best route. I was in favor of you pulling this out of the `util` module, but I think we should also just not refer to it as a utility function.

claymcleod (changes requested):
Here's my thoughts.

wdl-analysis/src/lib.rs
claymcleod: Okay, I see. That sounds good then.

claymcleod: Perhaps then `PROGRESS_BAR_DELAY_BEFORE_RENDER`.

claymcleod:
Additionally, when you merge this, please squash everything and remove references to things being "utilities".

wdl-engine/src/lib.rs
```diff
@@ -10,10 +10,267 @@ mod stdlib;
...
+            None => bail!(
+                "inputs file `{path}` is empty; use the `--name` option to specify the name of \
+                 the task or workflow to run",
+                path = path.display()
+            ),
+        }
+    } else if let Some(name) = name {
+        if document.task_by_name(&name).is_some() {
+            (None, name, Inputs::Task(Default::default()))
+        } else if document.workflow().is_some() {
```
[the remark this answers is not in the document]
a-frantz: That gets checked at [line 261](https://github.com/stjude-rust-labs/wdl/pull/283/files#diff-385deb0f21857ca2ae2a00a3c2fe3a916b03f49a0f0c1228ff90f728cc9e0c23R261-R263)

But it would probably be clearer to do that check here. I'll move that.

wdl-engine/Cargo.toml
a-frantz: I needed `codespan-reporting` to be included as a regular dependency for the `run()` and `validate_inputs()` functions, so after that change there's no need for a separate `codespan` `feature`.

wdl-engine/src/lib.rs
peterhuene: Should be able to now that we no longer have to do custom serialization like we did before the last refactoring.

wdl-engine/Cargo.toml
claymcleod: Sure, but these are not equivalent—the codespan feature also includes crate-specific functionality for `wdl-ast` (and, in turn, `wdl-grammar`). I'd just add the dependency without removing this feature.

claymcleod (approved):
After the above is added back in, it looks good to me.

claymcleod:
Make sure to squash the commits!

wdl-analysis/Cargo.toml
```diff
@@ -12,6 +12,7 @@ documentation = "https://docs.rs/wdl-analysis"
 
 [dependencies]
 wdl-ast = { path = "../wdl-ast", version = "0.9.0" }
+wdl-lint = { path = "../wdl-lint", version = "0.8.0" }
```
peterhuene, line 15: So this dependency will prevent us from basing `wdl-lint` on top of `wdl-analysis` in the future.

I was hoping that we could expose these functions in the `wdl` crate itself so that the upstream crates didn't have potentially unnecessary inter-dependencies like this.

wdl-analysis/Cargo.toml
```diff
@@ -25,6 +26,7 @@ petgraph = { workspace = true }
 futures = { workspace = true }
 path-clean = { workspace = true }
 indexmap = { workspace = true }
+indicatif = { workspace = true }
```
peterhuene, line 29: Similarly, this crate has been UX-agnostic until now, leaving the `wdl` crate (albeit locked away in the `bin` and not the `lib`) to implement and expose the UX.

wdl-engine/src/lib.rs
```diff
@@ -10,10 +10,251 @@ mod stdlib;
...
+use v1::TaskEvaluator;
 pub use value::*;
+use wdl_analysis::analyze;
+use wdl_analysis::path_to_uri;
+use wdl_grammar::Diagnostic;
+use wdl_grammar::Severity;
+
+/// Validates the inputs for a task or workflow.
+pub async fn validate_inputs(
+    document: String,
```
peterhuene, line 42: This can probably be `&str`.

wdl-engine/src/lib.rs
```diff
@@ -10,10 +10,251 @@ mod stdlib;
...
+    inputs: PathBuf,
```
peterhuene, line 43: Likewise, this can be `&Path` or `impl AsRef<Path>`.

wdl-engine/src/outputs.rs
```diff
@@ -8,7 +8,7 @@ use crate::Scope;
 use crate::Value;
 
 /// Represents outputs of a WDL workflow or task.
-#[derive(Default, Debug, Clone)]
+#[derive(Default, Debug, Clone, serde::Serialize)]
```
claymcleod, line 11: Let's just import `Serialize` here.

wdl/CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added an `--open` flag to `wdl doc` subcommand ([#269](https://github.com/stjude-rust-labs/wdl/pull/269)).
 * Added the `engine` module containing the implementation of `wdl-engine` ([#265](https://github.com/stjude-rust-labs/wdl/pull/265)).
 * Implemented the `wdl run` subcommand for running tasks ([#265](https://github.com/stjude-rust-labs/wdl/pull/265)).
+* Added a `validate` subcommand for validating input JSONs ([#283](https://github.com/stjude-rust-labs/wdl/pull/283)).
+* Added `analyze()`, `validate_inputs()`, and `run()` API entrypoints ([#283](https://github.com/stjude-rust-labs/wdl/pull/283)).
```
claymcleod, line 18:
```suggestion
* Added `analyze()`, `validate_inputs()`, and `run()` entrypoints ([#283](https://github.com/stjude-rust-labs/wdl/pull/283)).
```

claymcleod (approved):

wdl-engine/src/outputs.rs
```diff
@@ -3,20 +3,21 @@
...
 
 use indexmap::IndexMap;
+use serde::Serialize;
 
 use crate::Scope;
 use crate::Value;
 
 /// Represents outputs of a WDL workflow or task.
-#[derive(Default, Debug, Clone)]
+#[derive(Default, Debug, Clone, Serialize)]
```
peterhuene, line 12: Rather than implement `Serialize` here, I think we should move the `serialize` function which is no longer needed since #277 and turn it into an explicit `Serialize` implementation.

wdl-engine/src/outputs.rs
```diff
@@ -3,20 +3,21 @@
...
 pub struct Outputs {
     /// The name of the outputs.
     ///
     /// This may be set to the name of the call in a workflow or the task name
     /// for a direct task execution.
     name: Option<String>,
     /// The map of output name to value.
-    values: IndexMap<String, Value>,
+    pub values: IndexMap<String, Value>,
```
peterhuene, line 20: If we make `Outputs` properly serializable with what's being done in the `serialize` function, I don't think this will need to be public.

peterhuene (commented):
Just two comments relating to `Outputs` serialization, otherwise looks great!

peterhuene (approved):

peterhuene:
Looks like we're missing a `use` for the `Serialize` trait in `cli.rs`, but otherwise 👍

Issue #276, opened 2024-12-20
improve `fix` messages


a-frantz:
This issue is the successor of #193 

#260 ensured that every lint diagnostic includes a `fix` message, however some of them are still lacking. We should take another pass and include "corrected" code snippets wherever possible.

#264 (adds `ShellCheck`) uses the same `fix` message for every diagnostic. SC does provide its own "fix" for _some_ diagnostics (but not all). It would be great if we could "translate" the SC fixes into our own fix messages.

thatRichman:
I'd be happy to take a stab at this one, in the interest of finishing what I started in #264.

a-frantz:
Sure, go right ahead! If you want to just focus on the `ShellCheck` fix messages, that would be fine. But we would certainly appreciate a more thorough PR that improves the other lint rule messages as well! You're call :)

I think the goal is to supply "corrected" code whenever possible. That is more straightforward for some rules than others, so the scope of this PR can be up to you.

Gyan-max:
@a-frantz Hi! I noticed this issue is still marked as open. Is it still available to work on?

PR #284, opened 2025-01-10
feat: improve shellcheck fix messages

thatRichman:

Partially addresses #276 .

This PR improves ShellCheck lint messages by parsing, when available, the `fix` field and applying the suggested replacements. To enable this, the `fix` module was added which is closely modeled after shellcheck's own [Fixer](https://github.com/koalaman/shellcheck/blob/master/src/ShellCheck/Fixer.hs). The intent for this module is for it be usable by other lint rules to suggest and apply their own replacements using a consistent framework, but this is out-of-scope for this PR. 

      changes (when appropriate).


wdl-lint/tests/lints/shellcheck-warn/source.errors
```diff
@@ -7,7 +7,7 @@ note[ShellCheck]: Double quote to prevent globbing and word splitting.
    │                      SC2086[info]: Double quote to prevent globbing and word splitting.
    │                      more info: https://www.shellcheck.net/wiki/SC2086
    │
-   = fix: address the diagnostic as recommended in the message
+   = fix: did you mean `"$line17"`?
```
thatRichman, line 10: ShellCheck includes the entire line for context with its 'did you mean' messages, but this feels redundant given the highlight above.

thatRichman:
There's opportunity here to take some of the warnings that shellcheck does not currently emit `fix`es for, but I think this may not be a good idea for a couple of reasons.

1. It means that if shellcheck is updated to begin emitting fixes, we're overriding that.
2. Much of the low-hanging fruit is actually fairly complicated because shellcheck will emit multiple warnings for the same line, (e.g.  `[[foo]]` emits three conflicting suggestions), of which we would have to decide priority.

wdl-lint/src/fix.rs
```diff
@@ -0,0 +1,247 @@
...
+
+use std::ops::Range;
+use std::ops::RangeInclusive;
+
+use ftree::FenwickTree;
+use serde::Deserialize;
+
+#[derive(Copy, Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[non_exhaustive]
```
a-frantz, line 11: I don't think this is necessary. How would this potentially be expanded in the future?

wdl-lint/tests/lints/shellcheck-warn/source.errors
a-frantz: I think you made the right call to chop away the excess text.

wdl-lint/tests/lints/shellcheck-warn/source.errors
a-frantz, line 1: Can you update this test to include errors other than double quoting? I want to see some more examples of the constructed fix messages.

wdl-lint/src/fix.rs
thatRichman: Good catch. Generally I avoid assuming externally-derived enums are exhaustive but I also can't imagine there's other potential values.

wdl-lint/Cargo.toml
```diff
@@ -15,6 +15,7 @@ readme = "../README.md"
 wdl-ast = { path = "../wdl-ast", version = "0.9.0" }
 anyhow = { workspace = true }
 convert_case = { workspace = true }
+ftree = "1.2.0"
```
a-frantz, line 18: Please add this to the workspace Cargo instead

wdl-lint/src/fix.rs
```diff
@@ -0,0 +1,246 @@
...
+    /// A Fenwick tree for tracking modifications.
+    tree: FenwickTree<i32>,
+}
+
+#[allow(unused)]
+impl Fixer {
+    /// Create a new Fixer from a String.
+    pub fn new(value: String) -> Self {
+        Fixer {
+            tree: FenwickTree::from_iter(vec![0; value.len() + 1]),
```
a-frantz, line 100: Can you add a comment explaining the `+1`?

wdl-lint/src/fix.rs
```diff
@@ -0,0 +1,246 @@
...
+            value,
+        }
+    }
+
+    /// Apply a `Replacement` to the value contained in the Fixer.
```
a-frantz, line 105:
```suggestion
    /// Apply a [`Replacement`] to the value contained in the Fixer.
```

wdl-lint/src/fix.rs
```diff
@@ -0,0 +1,246 @@
...
+    pub fn apply_replacement(&mut self, rep: &Replacement) {
```
a-frantz, line 106:
```suggestion
    pub fn apply_replacement(&mut self, replacement: &Replacement) {
```
When balancing verbosity VS clarity, our team almost always picks clarity. Long variable names are acceptable round these parts (within reason of course).

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -64,27 +68,87 @@ static SHELLCHECK_EXISTS: OnceLock<bool> = OnceLock::new();
...
+/// Suggested fix for a ShellCheck diagnostic.
+struct ShellCheckFix {
+    /// The replacements to perform.
+    pub replacements: Vec<ShellCheckReplacement>,
+}
+
+#[derive(Clone, Debug, Deserialize)]
+/// A ShellCheck replacement.
+///
+/// This differs from a `Replacement` in that
```
a-frantz, line 81:
```suggestion
/// This differs from a [`Replacement`] in that
```

a-frantz (commented):
Mostly just nits related to variable naming and getting doc comments to link to definitions.

wdl-lint/src/rules/shellcheck.rs
```diff
@@ -64,27 +68,88 @@ static SHELLCHECK_EXISTS: OnceLock<bool> = OnceLock::new();
...
+    /// An `InsertionPoint`.
+    #[serde(rename = "insertionPoint")]
+    pub insertion_point: InsertionPoint,
+    /// Column replacement occurs on.
+    pub column: usize,
+    /// Column replacements ends on.
+    #[serde(rename = "endColumn")]
+    pub end_column: usize,
+    /// Replacement text.
+    pub replacement: String,
```
a-frantz, line 104:
```suggestion
    pub value: String,
```
to match the core `Replacement` struct

wdl-lint/src/fix.rs
```diff
@@ -0,0 +1,272 @@
...
+
+use std::ops::Range;
+use std::ops::RangeInclusive;
+
+use ftree::FenwickTree;
+use serde::Deserialize;
+
+#[derive(Copy, Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+/// An insertion point.
```
peterhuene, line 11: Nit: for consistency, can we move the doc comment above the derive macros?

wdl-lint/src/fix.rs
```diff
@@ -0,0 +1,272 @@
...
+pub enum InsertionPoint {
+    /// Insert immediately before a specified region.
+    BeforeStart,
+    /// Insert immediately after a specified region.
+    AfterEnd,
+}
+
+#[derive(Clone, Debug)]
+/// A replacement to be applied to a String.
```
peterhuene, line 20: Nit: for consistency, can we move the doc comment above the derive macros?

wdl-lint/src/fix.rs
```diff
@@ -0,0 +1,272 @@
...
+    end: usize,
+    /// where to insert the replacement.
+    insertion_point: InsertionPoint,
+    /// Value to be inserted.
+    value: String,
+    /// Precedence for replacement. Higher precedences are applied first.
+    precedence: usize,
+}
+
+#[allow(unused)]
```
peterhuene, line 34: Which member is not used?

wdl-lint/src/fix.rs
```diff
@@ -0,0 +1,272 @@
...
+impl Fixer {
+    /// Create a new Fixer from a String.
+    pub fn new(value: String) -> Self {
+        Fixer {
+            tree: FenwickTree::from_iter(vec![0; value.len() + 1]),
+            value,
+        }
+    }
+
+    /// Apply a [`Replacement`] to the value contained in the Fixer.
```
peterhuene, line 113: Can we add a `# Panic` section mentioning that this will panic if the replacement is out of bounds?

peterhuene (commented):
This looks really good and something we might build upon for a general purpose fixing framework for lint diagnostics in the future.

Just a few nits and minor comments.

a-frantz (approved):
LGTM after Peter's comments are addressed

wdl-lint/src/fix.rs
thatRichman: `adjust_range_inc` is not used, but since I implemented an `adjust_range` convenience method I figured having one for RangeInclusive would also be good.

peterhuene: Can we either apply it to `adjust_range_inclusive` directory or remove it if we don't need it? I would prefer not to have the directive on the `impl`.

thatRichman: I went ahead and removed it, can always add it back if it's needed.

peterhuene (approved):
Looks good (pending fix to the CI).

Thanks again for this work as it'll definitely improve the UX with shellcheck installed!

PR #298, opened 2025-01-17
ci: bump to right version

a-frantz:

### DIFF
--- a/ci/src/main.rs
+++ b/ci/src/main.rs
@@ -327,10 +327,8 @@ fn bump(version: &str, patch_bump: bool) -> String {
     }
     if major != 0 {
         format!("{}.0.0", major + 1)
-    } else if minor != 0 {
-        format!("0.{}.0", minor + 1)
     } else {
-        format!("0.0.{}", patch + 1)
+        format!("0.{}.0", minor + 1)
     }
 }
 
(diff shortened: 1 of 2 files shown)


ci/src/main.rs
```diff
@@ -327,10 +327,8 @@ fn bump(version: &str, patch_bump: bool) -> String {
     }
     if major != 0 {
         format!("{}.0.0", major + 1)
-    } else if minor != 0 {
-        format!("0.{}.0", minor + 1)
     } else {
-        format!("0.0.{}", patch + 1)
+        format!("0.{}.0", minor + 1)
```
peterhuene, line 331: Would this not preclude us from making patch releases until we have a 1.0?

Could you describe what problem this is addressing so I can understand the change?

peterhuene: Oh never mind, I saw the `patch_bump` conditional above after expanding the diff further.

I still don't understand what this fixes, exactly.

a-frantz: We want our "first" releases to be `v0.1.0`, not `v0.0.1` don't we?

peterhuene: Oh right, for `wdl-engine` I thought we'd just manually bump it to `0.1.0`, but this works too!

peterhuene (approved):

PR #302, opened 2025-01-21
fix: `wdl run` should join input paths correctly.

### LINKED ISSUE #301: Files aren't being localized to the `work` directory when they are workflow inputs.
When using a `File` as the input to a workflow in `wdl`, it doesn't appear that the files are properly being localized to the `work` directory. It seems as though the files _are_ localized if they aren't inputs to the workflow.

For example,

```wdl
version 1.2

task count_lines {
    input {
        File file
    }

    command <<<
      wc -l ~{file}
    >>>
}

workflow run {
    input {
        File file
    }

    call count_lines { input: file }
}
```

This will fail because `file` is not localized to the `count_lines` working directory.

peterhuene:
This PR fixes the `wdl run` command to properly join input paths relative to the inputs file.

The code to make the path of the inputs file absolute was incorrectly done, resulting in the later `join_paths` call to effectively be a no-op when the inputs file path on the command line was relative to CWD (e.g. `./foo.json`).

As a result, task execution would fail when using the path as the path would then be treated as relative to the task's working directory as it was not made absolute.

The fix is to simply return the absolute path correctly from `parse_inputs`.

Fixes #301.

      changes (when appropriate).

### DIFF
--- a/wdl/CHANGELOG.md
+++ b/wdl/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Fixed
+
+* Fixed `wdl run` not correctly updating file/directory paths in an inputs file ([#302](https://github.com/stjude-rust-labs/wdl/pull/302)).
+
 ## 0.11.0 - 01-17-2025
 
 ### Added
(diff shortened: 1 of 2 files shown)


a-frantz (commented):
Changes look good! But can we add some comments to the body of `parse_inputs()` to clarify what's happening? It's a bit messy and hard to follow IMO.

peterhuene:
Done.

a-frantz (approved):

adthrasher (approved):
Appreciate the extra comments!

PR #306, opened 2025-01-22
chore: update gauntlet

a-frantz:


a-frantz:
Went through the changes and didn't see anything problematic. Just a routine update here.

adthrasher (approved):
I wonder if we should somehow separate the checksum-only changes to facilitate easier review.

a-frantz:
> I wonder if we should somehow separate the checksum-only changes to facilitate easier review.

I've had the same thought, but I'm not sure how to accomplish that in practice.

PR #310, opened 2025-01-23
feat: report progress during `wdl run`.

peterhuene:
This PR implements reporting progress during `wdl run`. The `WorkflowEvaluator` now accepts a callback that is called when a task or workflow is evaluated and when it is completed.

The progress bar displayed by `wdl run` has been updated to print out the number of completed tasks, the number of in-process tasks, including up to 10 names of tasks currently being executed.

Additionally, this changes the progress bars displayed by `wdl` to target stdout so that it isn't interleaved with log output on stderr.

The default log level for `wdl` is now `warn` rather than `error` (one less `-v` option required to get to useful log output).

`wdl run` now prefers to run the workflow in a document containing a single workflow and a single task, when not provided either a path to an inputs file or the name of the task/workflow to run. Previously it preferred the task over the workflow.

      changes (when appropriate).

### DIFF
--- a/wdl/src/cli.rs
+++ b/wdl/src/cli.rs
@@ -1,9 +1,9 @@
 //! Entry point functions for the command-line interface.
 
+use std::fmt;
 use std::path::Path;
 use std::path::PathBuf;
 use std::path::absolute;
-use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::Context;
@@ -11,7 +11,9 @@ use anyhow::Result;
 use anyhow::anyhow;
 use anyhow::bail;
 use colored::Colorize;
+use indexmap::IndexSet;
 use indicatif::ProgressBar;
+use indicatif::ProgressDrawTarget;
 use indicatif::ProgressStyle;
 use serde_json::to_string_pretty;
 use tokio::fs;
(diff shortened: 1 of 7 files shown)


wdl/src/cli.rs
```diff
@@ -152,22 +157,26 @@ pub fn parse_inputs(
...
-                    .workflow()
-                    .map(|w| (w.name().to_string(), Inputs::Workflow(Default::default())))
-            })
-            .context("inputs file is empty and the WDL document contains no tasks or workflow")?;
-
-        if iter.next().is_some() {
-            bail!("inputs file is empty and the WDL document contains more than one task");
-        }
+        // Neither an inputs file or name was provided, look for a workflow in the
+        // document; failing that, find at most one task in the document
```
a-frantz, line 161: Do we really need to support this use case? I feel like if you are providing neither a name nor an inputs file, it's fine for us to just fail. I guess there isn't any harm in supporting this case, it just strikes me as a bit odd.

a-frantz (commented):
Code looks good! Gonna checkout the branch and kick the tires a bit, but didn't see any issues during review.

wdl/src/cli.rs
peterhuene: It's useful for when there's a workflow you want to run that either has no inputs or all inputs have default values, so one doesn't need to specify an inputs file or explicitly give it a name.

claymcleod (approved):
This all seems good. Sometime earlier this year, I was thinking about a more general event driven system that subscribers could register to within WDL. This would support more than just the progress types here but any major event that was executed. The design I had in mind was similar to this implementation but just a broad definition for events—something to consider in the future.

a-frantz (approved):

PR #316, opened 2025-01-26
fix: use the correct type for scatter output.

### LINKED ISSUE #315: Deferred gathering after scatter
Consider this workflow:

```wdl
version 1.2

struct Sample {
  String kind
  String name
}

workflow run {
  Array[Sample] samples = [
    Sample {
      kind: "normal",
      name: "sample_one",
    },
    Sample {
      kind: "tumor",
      name: "sample_two",
    },
    Sample {
      kind: "normal",
     name: "sample_three",
    },
    Sample {
      kind: "tumor",
      name: "sample_four",
    },
  ]

  scatter (sample in samples) {
    if (sample.kind == "normal") {
      String name = sample.name
    }
  }

  output {
    Array[String] names = select_all(name)
  }
}
```

This fails to run with the following diagnostic:

```bash
error: type mismatch: argument to function `select_all` expects type `Array[X]`, but found type `String?`
   ┌─ foo.wdl:35:38
   │
35 │     Array[String] names = select_all(name)
   │                                      ^^^^ this is type `String?`

error: aborting due to previous diagnostic
```

However, in the section on [the `scatter` statement](https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#scatter-statement), the specification makes it clear that `name` should be gathered and immediately available as an `Array[String?]`:

> After evaluation has completed for all iterations of a scatter, each declaration or call output in the scatter body (except for the scatter variable) is collected into an array, and those array declarations are exposed in the enclosing context. In other words, for a declaration or call output T <name> within a scatter body, a declaration Array[T] <name> is implicitly available outside of the scatter body.

It appears that, in the current implementation, we defer the gathering until it is assigned to a type of that same value.

peterhuene:
This commit fixes an incorrect type used when creating the output of a scatter statement into the parent scope.

It should be an array type, but the array's element type was being used instead.

The fix is to use the proper type and assert, at a minimum, that the `new_unchecked` functions at least get the expected compound type.

Fixes #315.

      changes (when appropriate).


a-frantz (approved):

claymcleod:
Great, thanks!

PR #320, opened 2025-01-30
refactor: change workflow evaluation to use `tokio::spawn`.

peterhuene:
This refactoring primarily affects `wdl-engine` and how workflows are evaluated. Previously, workflows were evaluated without spawning any tokio tasks. This meant that subgraph evaluation was not concurrent, although WDL tasks were spawned concurrently.

With the upcoming support for integrating Crankshaft, the `Backend` trait was refactored such that the backend itself can maintain state and notify callers of when a WDL task is spawned and when it completes. This allows for the local execution backend to be rewritten in a way that enables it to respect the `cpu` and `memory` requirements of a WDL task.

To support making the evaluator futures `Send`, we now utilize `rowan::AstPtr` to send AST nodes between threads performing evaluation. The evaluation graphs are mapped to be `Send` so that evaluation can keep a shared reference to the graph across concurrently executing tokio tasks.

Also included with these changes:

* Use `tracing-indicatif` to prevent mixing of tracing output and progress bars; the progress bar now consistently shows at the end of tracing output and can also be quieted.
* Made `wdl_analysis::Document` trivially clonable.
* Minor changes to internal names in `wdl-analysis`.
* Fixed a missing constraint on the `basename` function definition for an overload that's only required for WDL 1.2.
* Added a `Config` type to `wdl-engine` that controls how evaluation is performed; in the future this can be used to support evaluation configuration files.
* Expanded the `ProgressKind` enumeration to allow for tracking how many tasks are "ready" (i.e. waiting to be spawned by the backend) vs. "executing" (i.e. currently running).
* Removed the incorrect mapping of host paths to guest paths; this will soon be reimplemented as part of the initial support for a Docker backend.
* Added retry logic to task execution.

      changes (when appropriate).


wdl-engine/src/eval/v1/task.rs
```diff
@@ -251,186 +409,197 @@ impl<'a> TaskEvaluator<'a> {
...
-            let mut outputs: Outputs = mem::take(&mut state.scopes[OUTPUT_SCOPE_INDEX.0]).into();
-            if let Some(section) = definition.output() {
-                let indexes: HashMap<_, _> = section
-                    .declarations()
-                    .enumerate()
-                    .map(|(i, d)| (TokenStrHash::new(d.name()), i))
-                    .collect();
-                outputs.sort_by(move |a, b| indexes[a].cmp(&indexes[b]))
+            if let Err(e) = evaluated.handle_exit(state.request.requirements()) {
+                if retry >= self.config.task.retries.unwrap_or(0) {
```
a-frantz, line 562: Is the zero here a magic number? Should this refer to a CONST somewhere?

wdl-engine/src/eval/v1/task.rs
```diff
@@ -487,15 +665,60 @@ impl<'a> TaskEvaluator<'a> {
...
+                value
+                    .as_primitive()
+                    .expect("value should be primitive")
+                    .raw()
+                    .to_string(),
+            );
+        }
+
+        // If the backend uses containers, add a mapping entry for any paths the value
+        // might contain Initially, it will be an empty value, but will be
```
a-frantz, line 710:
```suggestion
        // might contain. Initially, it will be an empty value, but will be
```

wdl-engine/src/backend/local.rs
```diff
@@ -1,325 +1,472 @@
...
+        total_memory: u64,
+    ) {
+        let mut state = State::new(total_cpu, total_memory);
+
+        loop {
+            // If there aren't any spawned tasks, wait for a spawn request only
+            if state.spawned.is_empty() {
+                assert!(
+                    state.parked.is_empty(),
+                    "there can't be any parked requests if there are no spawned tasks"
```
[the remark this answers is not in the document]
peterhuene: Only tasks that _could_ run will ever get parked, otherwise the task fails if the host doesn't have enough resources to at least guarantee enough CPUs/memory.

wdl-engine/src/config.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+/// Represents task evaluation configuration.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct TaskConfig {
+    /// The maximum number of retries to attempt if a task fails.
+    ///
+    /// Defaults to 0 (no retries).
```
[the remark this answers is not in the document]
peterhuene: The spec does define the default for `max_retries` to be 0, which we aren't respecting yet (I can fix as part of this PR).

This setting should become the _default_ maximum retries if a task doesn't specify `max_retries`, which makes sense to keep at 0.

wdl-engine/src/eval/v1/task.rs
peterhuene: It's magic and should have a default constant like we use for cpu/memory.

wdl-engine/src/config.rs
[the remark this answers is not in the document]
a-frantz: Alright. If it's defined by the spec we should respect that. On the workflows repo we've hardcoded every task to attempt at least one retry, so the default doesn't really impact us I guess

wdl-engine/src/config.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// By default, the value is the parallelism supported by the task
+    /// execution backend.
+    ///
+    /// A value of `0` is invalid.
+    ///
+    /// Lower values use less memory for evaluation and higher values may better
+    /// saturate the task execution backend with tasks to execute.
+    ///
+    /// <div class="warning">
+    /// Warning: nested scatter statements cause exponential memory usage based
```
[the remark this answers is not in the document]
peterhuene: This setting is more about how many scatter array elements get processed concurrently per scatter statement encountered, rather than how many tasks may be executed concurrently (which is ultimately up to the execution backend since they're resource constrained). It's really just a dial so we don't explode memory when evaluating nested scatter statements (one of the reason that `wdl run` shows a far less amount of memory used to evaluate that scatter stress test).

The default of using the backend's maximum task execution concurrency means that, by default, we scatter about as much as the backend will execute tasks concurrently, assuming the tasks don't haven't constraints resources that would further restrict it.

Take for example this:

```wdl
scatter (i in range(16)) {
    call foo
}

scatter (j in range(16)) {
    call bar
}
```

On a system with 16 cores (like my laptop), the default local backend supports up to 16 concurrent tasks (assumes that generally a task wants a 1.0 cores). Therefore, by default, we will evaluate up to 16 array elements per concurrent scatter.

This would immediately send 16 `foo` tasks and 16 `bar` tasks to the backend, 16 of those would immediately execute and 16 would be parked waiting on the others to finish.

Contrast this with:

```wdl
scatter (i in range(16)) {
    scatter (j in range(16)) {
        call foo
    }
}
```

Which would immediately send up 16*16 `foo` tasks to the backend, 16 of those would start, the other 240 would wait.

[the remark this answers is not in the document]
a-frantz: Ah ok. I was misunderstanding. Really good explanation here! Can you add a bit more to the code comment? I don't think you need to go into as much detail as this comment, but I appreciate the explanation!

wdl-engine/src/backend/local.rs
```diff
@@ -1,325 +1,472 @@
...
+    /// Spawns the requested task.
+    ///
+    /// Returns the status code of the process when it has exited.
+    async fn spawn_task(request: Arc<TaskSpawnRequest>) -> Result<i32> {
         // Recreate the working directory
-        if self.work_dir.exists() {
-            fs::remove_dir_all(&self.work_dir).with_context(|| {
+        let work_dir = request.root.work_dir();
+        if work_dir.exists() {
+            fs::remove_dir_all(work_dir).with_context(|| {
```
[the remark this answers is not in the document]
claymcleod: I would also put in for an immutable working directory structure with some sort of indication of what try you are on baked into the name.

wdl-engine/src/config.rs
```diff
@@ -0,0 +1,221 @@
...
+    }
+}
+
+/// Represents scatter statement evaluation configuration.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub struct ScatterConfig {
+    /// The number of scatter array elements to process concurrently.
+    ///
+    /// By default, the value is the parallelism supported by the task
```
claymcleod, line 73: I'm wondering (a) if this is necessary and (b) if it makes a difference in speed. If I'm understanding correctly (which I may not be), this would also limit the number of concurrent scatters that don't actually call a task, which seems a pity.

claymcleod: I now see the other comment that explains this a bit. I think it's fine, though I am still curious about the speed (does it come out to roughly the same speed as executing them all concurrently?).

wdl-engine/src/config.rs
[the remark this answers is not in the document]
claymcleod: Plus, I think the point here is that we can simply tune this knob based on your configuration of Sprocket, so it should be easy to setup a production-ready configuration.

wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -69,12 +71,48 @@ use crate::ScopeRef;
...
 use crate::v1::ExprEvaluator;
 use crate::v1::TaskEvaluator;
 
+/// Helper for formatting a workflow or task identifier for a call statement.
+fn id(namespace: Option<&str>, target: &str, alias: &str, scatter_index: &str) -> String {
+    if alias != target {
+        match namespace {
+            Some(ns) => {
+                format!(
+                    "{ns}-{target}-{alias}{sep}{scatter_index}",
```
claymcleod, line 87: I'm uncertain, but perhaps the retry counter should be in here?

claymcleod (approved):

wdl-engine/src/backend/local.rs
```diff
@@ -1,325 +1,472 @@
...
+                        None => break,
+                    }
                 }
+                Some(Ok(response)) = state.spawned.join_next() => {
+                    state.cpu += response.request.cpu;
+                    state.memory += response.request.memory;
+                    response.request.tx.send(response.result).ok();
+
+                    // Look for tasks to unpark
+                    while let Some(pos) = state.parked.iter().position(|r| r.cpu <= state.cpu && r.memory <= state.memory) {
```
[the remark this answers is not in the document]
peterhuene: We definitely could, I figured that since this isn't going to be a backend used in production (or at least really shouldn't be), then I wasn't going to do anything terribly fancy here.

wdl-engine/src/eval/v1/workflow.rs
peterhuene: The retry is internal to the task's evaluation, so not something visible to the workflow call statement site.

wdl-engine/src/config.rs
```diff
@@ -0,0 +1,221 @@
...
+    /// Defaults to the total amount of memory for the host.
+    ///
+    /// The value cannot be zero or exceed the host's total amount of memory.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub memory: Option<String>,
+}
+
+impl LocalBackendConfig {
+    /// Validates the local task execution backend configuration.
+    pub fn validate(&self) -> Result<()> {
```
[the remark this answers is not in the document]
peterhuene: Agreed that this needs some test coverage.

At worst, we could set some values that are unlikely to be realistically achieved on modern hardware (and certainly not in a CI environment).

wdl-engine/src/eval/v1/workflow.rs
claymcleod: Okay, that's fine then.

wdl-engine/src/config.rs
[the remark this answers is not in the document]
claymcleod: Sounds good. It just strikes me as a _critically_ important section for future PRs, as I imagine the community is going to want to add configuration settings, and we need to support that by showing how we want tests to be run. Configuration is always the _most_ annoying when it doesn't work and is hard to figure out why.

wdl-engine/src/config.rs
peterhuene: The speed is roughly the same, but evaluation isn't usually the bottleneck for actually running a workflow, it's task execution since that will (soon) involve spinning up containers; evaluation will be dominated by waits.

wdl-engine/src/backend/local.rs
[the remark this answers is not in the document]
claymcleod: Oh that's totally fine—okay, I did not remember that context when I was reviewing. So let this serve as a comment of "we _should_ support fair sharing in backend implementations". You can mark as resolved when you've seen this.

wdl-engine/src/backend/local.rs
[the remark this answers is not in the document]
peterhuene: We'll go with that approach and leave working directories untouched.

There's a setting in miniwdl that defaults to off regarding cleaning up working directories, which we can replicate in the future.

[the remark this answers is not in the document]
claymcleod: That sounds good.

wdl-engine/src/eval/v1/task.rs
```diff
@@ -554,16 +583,29 @@ impl TaskEvaluator {
...
+                        "too many attempts were made to run the task `{task}`",
+                        task = state.task.name()
+                    )
+                })?);
                 task.set_return_code(evaluated.status_code);
             }
 
             if let Err(e) = evaluated.handle_exit(state.request.requirements()) {
-                if retry >= self.config.task.retries.unwrap_or(0) {
+                if attempt >= max_retries {
```
a-frantz, line 596: Is this right? If we start with `attempt = 0` and max_retries is also `0`, this will error. Should it be `attempt > max_retries`?

a-frantz (commented):
Q: what about the temp dir on retries? And stderr/stdout files? Are those overwritten or preserved?

wdl-engine/src/eval/v1/task.rs
peterhuene: It should error when `attempt` is `0` (i.e. first attempt) and `max_retries` is also `0` (i.e. no retries), as we increment the counter after this check, right?

If `max_retries` was `1` (i.e. retry at most one time), then we want to bail when `attempt` was also `1` (i.e. second attempt).

Unfortunately `attempt` has to start at `0` and not what one would logically expect (i.e. `1`) per the WDL spec where `attempt` is visible from the `task` variable in 1.2.

peterhuene: Although the language is a little vague in the spec too.

I view `max_retries` as the limit to the _retries_ and not how many attempts were made, as there will always be at least one attempt made to execute the task.

> The max_retries attribute specifies the maximum number of times a task should be retried in the event of failure. The execution engine must retry the task at least once and up to (but not exceeding) the specified number of attempts.

Here it's saying up to the specified number of _attempts_, but that's not the same thing as the number of retries.

If I say "don't retry this thing more than 5 times", it means that it'll run 6 times at most. 🤷

peterhuene: FYI, I'm adding a test for the retry logic to task evaluation; this is missing reevaluating the command on the retry so that `task.attempt` reflects the bump.

I've now realized that it's not just the work directory we'd want to preserve, but the entire execution root (the command that was run, stdout/stderr, etc).

Let me address that now.

a-frantz: Would the `command` ever change between attempts? I suppose in WDL 1.2 you can reference the attempt number, so in that case it would make sense to isolate `command` files by attempt.

peterhuene: Yeah, I basically wrote the test such that it used `task.attempt` to fail up to some specific attempt and then succeed, but it wasn't working as we only evaluated it once as 0.

a-frantz: (Although that kind of "meta" reference inside your command sounds like a _terrible_ idea for reproducibility)

claymcleod: Just a note confirming that `task` can be used inside the command (see the first example here: https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#runtime-access-to-requirements-hints-and-metadata), so each attempt will need it's own copy of the command.

peterhuene: As command evaluation can create files via the `write_*` standard library functions, I think we'll need a task directory structure like this:

```
<root>/
├─ tmp/                # Where files are created by the stdlib for evaluation *before* the command
├─ attempt/            # Stores the attempts
│  ├─ 0/               # First attempt 
│  │  ├─ work/         # Working directory for the task's first execution
│  │  ├─ tmp/          # Where files are created by the stdlib during command evaluation
│  │  ├─ command       # The evaluated command for the first execution
│  │  ├─ stdout        # The standard output of the first execution
│  │  ├─ stderr        # The standard error of the first execution
│  ├─ 1/
│  │  ├─ ...
```

This also exposes another bug specific to command evaluation wherein calls to `write_*` functions within command placeholders are getting host paths instead of expected guest paths. Right now the placeholder evaluation for commands remaps host paths to guest paths, but only for files _not_ created by the command itself.

I can fix that here too.

peterhuene: I'm going to punt on fixing the path mapping into the implementation I am working on for the docker backend. I did push up a fix so that we layout the directory structure as above.

peterhuene:
Looks like a new nightly was released that changed formatting, so need to bump my local nightly and reformat.

peterhuene:
> Q: what about the temp dir on retries? And stderr/stdout files? Are those overwritten or preserved?

There are now two temp directories in the directory layout as described in a previous comment: one is where files are created _before_ and _after_ we evaluate the command and one is for the evaluation of the current attempt's command.

So effectively they are preserved from attempt to attempt.

claymcleod (approved):

PR #323, opened 2025-02-06
feat: implement mount points for task execution.

### LINKED ISSUE #267: feature: properly map file/directory paths for task evaluation
With the initial implementation of task evaluations in #265, only `File`/`Directory` values that are transitively referenced by the command get mapped into container environments.

This means that auxiliary inputs to a task, such as specifying a `bai` input that isn't referenced by the command directly, will _not_ be mapped into the container environment; note that in `wdl-analysis` we have a heuristic to suppress unused inputs with specific suffixes (`bai` being one of them) for this reason as well.

While referencing a struct value that contains multiple files should map all of the structs paths, we should probably map, at the very least, all inputs along with any private declarations explicitly referenced by the command; alternatively, just map all inputs and all private declarations regardless of being referenced by the command.

peterhuene:
This PR implements the calculation of mount points for task execution backends that use containers.

Paths from `File` and `Directory` values that are visible prior to evaluating the command section of a task are added to a `PathTrie`.

The `PathTrie` is used to determine the minimum number of mounts in the guest environment to be able to send input files to the container.

Command evaluation does a translation from host paths to guest paths so that the generated command script uses only guest paths.

Similarly, evaluation of output values does a translation from guest paths to host paths so that tasks always output the host path.

Fixes #267.

      changes (when appropriate).

### DIFF
--- a/wdl-engine/src/backend.rs
+++ b/wdl-engine/src/backend.rs
@@ -289,8 +289,8 @@ pub trait TaskExecutionBackend: Send + Sync {
     ///
     /// Containerized execution uses the following guest paths:
     ///
-    /// * `<root>/inputs` - where inputs are mounted; this is mounted read-only.
-    /// * `<root>/work` - the task working directory; this is mounted
+    /// * `<root>/inputs/` - where inputs are mounted; this is mounted read-only.
+    /// * `<root>/work/` - the task working directory; this is mounted
     ///   read-write.
     /// * `<root>/command` - the command to execute; this is mounted read-only.
     ///
--- a/wdl-engine/src/backend.rs
+++ b/wdl-engine/src/backend.rs
@@ -96,11 +96,11 @@ pub struct TaskExecutionRoot {
 
 impl TaskExecutionRoot {
     /// Creates a task execution root for the given path and execution attempt.
-    pub fn new(path2: &Path, attempt: u64) -> Result<Self> {
-        let root_dir = absolute(path2).with_context(|| {
+    pub fn new(path: &Path, attempt: u64) -> Result<Self> {
+        let root_dir = absolute(path).with_context(|| {
             format!(
                 "failed to determine absolute path of `{path}`",
-                path = path2.display()
+                path = path.display()
             )
         })?;
 
--- a/wdl-engine/src/backend.rs
+++ b/wdl-engine/src/backend.rs
@@ -289,7 +289,8 @@ pub trait TaskExecutionBackend: Send + Sync {
     ///
     /// Containerized execution uses the following guest paths:
     ///
-    /// * `<root>/inputs/` - where inputs are mounted; this is mounted read-only.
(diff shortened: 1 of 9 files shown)


peterhuene:
I forgot to mention that I need to test that new unit test on Windows as the paths obviously are for unix, so I expected the Windows CI to fail; I'll get a similar test implemented to test Windows paths.

claymcleod (approved):
LGTM.

wdl-engine/src/backend.rs
```diff
@@ -93,44 +96,50 @@ pub struct TaskExecutionRoot {
 
 impl TaskExecutionRoot {
     /// Creates a task execution root for the given path and execution attempt.
-    pub fn new(path: &Path, attempt: u64) -> Result<Self> {
-        let path = absolute(path).with_context(|| {
+    pub fn new(path2: &Path, attempt: u64) -> Result<Self> {
```
a-frantz, line 99: Maybe I'm missing something obvious. Why is this `path2`? What's the `2` mean?

wdl-engine/src/eval.rs
```diff
@@ -370,3 +388,280 @@ impl EvaluatedTask {
         Ok(())
     }
 }
+
+/// Represents mount points for mapping host and guest paths for task execution
+/// backends that use containers.
+#[derive(Debug, Default)]
+pub struct MountPoints(Vec<(PathBuf, PathBuf, bool)>);
```
a-frantz, line 395: Can you add to this doc comment what the contents refer to? I can see from `insert` it's `(host_path, guest_path, read_only)` but would be good to go in this comment

wdl-engine/src/backend.rs
peterhuene: Whoops, that was a rename to see where it was being used for corrections, it should be `path`.

wdl-engine/src/backend.rs
```diff
@@ -294,13 +284,21 @@ pub trait TaskExecutionBackend: Send + Sync {
...
-    /// Gets the container root directory for the backend (e.g. `/mnt/task`)
+    /// Gets the container (guest) root directory for the backend (e.g.
+    /// `/mnt/task`).
+    ///
+    /// Containerized execution uses the following guest paths:
+    ///
+    /// * `<root>/inputs` - where inputs are mounted; this is mounted read-only.
+    /// * `<root>/work` - the task working directory; this is mounted
+    ///   read-write.
+    /// * `<root>/command` - the command to execute; this is mounted read-only.
```
a-frantz, line 295:
```suggestion
    /// * `<root>/inputs/` - where inputs are mounted; this is mounted read-only.
    /// * `<root>/work/` - the task working directory; this is mounted
    ///   read-write.
    /// * `<root>/command` - the command to execute; this is mounted read-only.
```
nit to clarify that the first two are directories and the last is a file

a-frantz (approved):

PR #350, opened 2025-03-11
revise: allow `validate_inputs` to accept input content directly

### LINKED ISSUE #347: Enhancement request: Allow validate_inputs to accept input content directly
Currently, the `wdl::cli::validate_inputs` function requires a file path to the input file. When working with input formats that need conversion (like YAML → JSON), this requires creating a temporary file, which adds unnecessary I/O overhead.

It would be more efficient if `validate_inputs` had an overload that accepted the input content directly as a string or as a parsed serde_json::Value, in addition to the existing file path version.

Possible API signature:
```rust
pub async fn validate_inputs_content(
    document: &str,
    input_json: &serde_json::Value,
) -> Result<Option<Diagnostic>> {
    // Implementation here
}
```

This would allow tools like Sprocket to avoid writing temporary files when converting between formats or when the input is generated programmatically.

fahdfady:
…validate_json_file
resolves #347 

### DIFF
--- a/wdl-engine/src/inputs.rs
+++ b/wdl-engine/src/inputs.rs
@@ -627,8 +627,37 @@ impl Inputs {
                 })?,
         );
 
+        // ? should we call parse_json here after turning our Path to a  JSON value?
+
         Self::parse_object(document, object)
             .with_context(|| format!("failed to parse input file `{path}`", path = path.display()))
+    }
+
+    /// Parses a JSON inputs file from the given JSON.
+    ///
+    /// The parse uses the provided document to validate the input keys within
+    /// the file.
+    ///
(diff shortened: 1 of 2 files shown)


claymcleod:
Hi @fahdfady, there are some CI issues if you could look at addressing those. Thank you!

peterhuene:
That's correct, we have a PR in-progress on `wdl-engine` to put support for YAML into `Inputs`, which we can then utilize in the helper for validating inputs and also update sprocket to support YAML from the configuration file (which at that point will be just updating the command line option help text).

peterhuene:
Hi @fahdfady. I appreciate this PR, but we decided to fix this as part of #352 which makes this PR unnecessary.

If you disagree, please let me know. For now, we'll close it.

PR #355, opened 2025-03-17
refactor: refactor `wdl-ast` to allow for alternative tree representation.

peterhuene:
This PR refactors `wdl-ast` so that all of the AST types take a type parameter for either a node or a token.

This change allows for the AST to operate off of a different representation of `SyntaxNode` and `SyntaxToken` for evaluation.

In the case of `wdl-engine`, evaluation is performed asynchronously and so the node and token types must be `Send`; however, the implementation from `rowan` is not `Send`. With this refactoring, the engine can "morph" the AST returned from analysis to an AST with its own implementation of `SyntaxNode` and `SyntaxToken` that are `Send`.

This cleans up quite a few things in the engine, most notably the engine no longer has to rebuild task and workflow evaluation graphs with `AstPtr`-based graph nodes.

It will also allow the engine's expression evaluator to become fully async in an upcoming change to allow for the stdlib functions to be asynchronous, which is require for the file functions that may fetch files from remote storage.

Many of the changes here stem from a decision to rename two methods in `wdl-ast`: `inner` instead of `syntax` and `text` instead of `as_str`. The rational for using `inner` is that since the underlying types are generic and do not have to be `SyntaxNode` or `SyntaxToken`, calling the method `syntax` no longer makes sense. Additionally, we need a method on the abstract node type for getting its text and `as_str` doesn't work since it must return `impl fmt::Display`; therefore, I decided to name it `text` on both the node and token traits.

This also fixes a bug where we weren't detecting duplicate hints sections in 1.2 documents.

      changes (when appropriate).

### DIFF
--- a/wdl-ast/src/v1/decls.rs
+++ b/wdl-ast/src/v1/decls.rs
@@ -818,7 +818,7 @@ impl<N: TreeNode> Decl<N> {
         kind == SyntaxKind::BoundDeclNode || kind == SyntaxKind::UnboundDeclNode
     }
 
-    /// Casts the given node to [`StructItem`].
+    /// Casts the given node to [`Decl`].
     ///
     /// Returns `None` if the node cannot be cast.
     pub fn cast(inner: N) -> Option<Self> {
--- a/wdl-ast/src/v1/decls.rs
+++ b/wdl-ast/src/v1/decls.rs
@@ -4,22 +4,23 @@ use std::fmt;
 
 use super::EnvKeyword;
 use super::Expr;
+use super::Plus;
+use super::QuestionMark;
 use crate::AstNode;
 use crate::AstToken;
 use crate::Ident;
 use crate::SyntaxKind;
 use crate::SyntaxNode;
-use crate::WorkflowDescriptionLanguage;
-use crate::support;
-use crate::token;
+use crate::TreeNode;
+use crate::TreeToken;
 
 /// Represents a `Map` type.
 #[derive(Clone, Debug, Eq)]
-pub struct MapType(SyntaxNode);
+pub struct MapType<N: TreeNode = SyntaxNode>(N);
 
-impl MapType {
+impl<N: TreeNode> MapType<N> {
     /// Gets the key and value types of the `Map`.
-    pub fn types(&self) -> (PrimitiveType, Type) {
+    pub fn types(&self) -> (PrimitiveType<N>, Type<N>) {
         let mut children = self.0.children().filter_map(Type::cast);
         let key = children
             .next()
@@ -38,33 +39,25 @@ impl MapType {
     }
 }
 
-impl PartialEq for MapType {
+impl<N: TreeNode> PartialEq for MapType<N> {
     fn eq(&self, other: &Self) -> bool {
         self.is_optional() == other.is_optional() && self.types() == other.types()
     }
 }
 
-impl AstNode for MapType {
-    type Language = WorkflowDescriptionLanguage;
-
-    fn can_cast(kind: SyntaxKind) -> bool
-    where
-        Self: Sized,
-    {
+impl<N: TreeNode> AstNode<N> for MapType<N> {
+    fn can_cast(kind: SyntaxKind) -> bool {
         kind == SyntaxKind::MapTypeNode
     }
 
-    fn cast(syntax: SyntaxNode) -> Option<Self>
-    where
-        Self: Sized,
-    {
-        match syntax.kind() {
-            SyntaxKind::MapTypeNode => Some(Self(syntax)),
+    fn cast(inner: N) -> Option<Self> {
+        match inner.kind() {
+            SyntaxKind::MapTypeNode => Some(Self(inner)),
             _ => None,
         }
     }
 
-    fn syntax(&self) -> &SyntaxNode {
+    fn inner(&self) -> &N {
         &self.0
     }
 }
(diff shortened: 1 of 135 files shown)


wdl-ast/src/v1/decls.rs
```diff
@@ -761,171 +717,151 @@ impl fmt::Display for Type {
...
     }
 
-    /// Attempts to cast the [`SyntaxNode`] to any of the underlying members
-    /// within the [`Decl`].
-    pub fn cast(syntax: SyntaxNode) -> Option<Self>
-    where
-        Self: Sized,
-    {
-        match syntax.kind() {
+    /// Casts the given node to [`StructItem`].
```
a-frantz, line 821:
```suggestion
    /// Casts the given node to [`Decl`].
```
?

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
+//! Implementation of syntax tree elements that are `Send`+`Sync`.
```
claymcleod, line 1:
```suggestion
//! Implementation of syntax tree elements that are `Send + Sync`.
```

claymcleod: Or `Send` and `Sync`

claymcleod (approved):
Looks good to me.

A note: I had to use a heuristic to review this many changed files + code. If the file had the same number of additions/deletions _and_ was in a file that I was relatively confident while scanning that the file just contained updated method calls, I just signed off on the file without reviewing it in detail. That being said, this should have caught all of the new/important changes.

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+    /// Returns `None` if there are no children.
+    pub fn first_child_or_token(&self) -> Option<SyntaxElement> {
+        self.children_with_tokens().next()
+    }
+
+    /// Gets the next sibling node.
+    ///
+    /// Returns `None` if there is no sibling node.
+    pub fn next_sibling(&self) -> Option<SyntaxNode> {
+        // This should also be a constant-time access rather than having to iterate
```
a-frantz, line 115:
```suggestion
        // This should also be a constant-time access rather than having to iterate.
```

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+        }
+
+        None
+    }
+
+    /// Gets the next sibling node or token.
+    ///
+    /// Returns `None` if there is no sibling node or token.
+    pub fn next_sibling_or_token(&self) -> Option<SyntaxElement> {
+        // This should also be a constant-time access rather than having to iterate
```
a-frantz, line 135:
```suggestion
        // This should also be a constant-time access rather than having to iterate.
```

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+                    NodeOrToken::Node(n) => self.new_child_node(n.to_owned(), index, start).into(),
+                    NodeOrToken::Token(t) => {
+                        self.new_child_token(t.to_owned(), index, start).into()
+                    }
+                }
+            })
+    }
+
+    fn last_token(&self) -> Option<Self::Token> {
+        // Unfortunately `rowan` does not expose the relative offset of each green child
```
a-frantz, line 229:
```suggestion
        // Unfortunately `rowan` does not expose the relative offset of each green child.
```

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+        // If it did, we could easily just look at the last child here instead of
+        // iterating to find the last child's start
```
a-frantz, line 231:
```suggestion
        // iterating to find the last child's start.
```

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+        None
+    }
+
+    /// Gets the next sibling node or token.
+    ///
+    /// Returns `None` if there is no sibling node or token.
+    pub fn next_sibling_or_token(&self) -> Option<SyntaxElement> {
+        // This should also be a constant-time access rather than having to iterate
+        // We need the offset relative to the start of the parent from the green node to
+        // do that; currently that information is private in `rowan`
```
a-frantz, line 137:
```suggestion
        // do that; currently that information is private in `rowan`.
```

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+        self.children_with_tokens().next()
+    }
+
+    /// Gets the next sibling node.
+    ///
+    /// Returns `None` if there is no sibling node.
+    pub fn next_sibling(&self) -> Option<SyntaxNode> {
+        // This should also be a constant-time access rather than having to iterate
+        // We need the offset relative to the start of the parent from the green node to
+        // do that; currently that information is private in `rowan`
```
a-frantz, line 117:
```suggestion
        // do that; currently that information is private in `rowan`.
```

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+pub struct SyntaxToken(Arc<ElementData>);
+
+impl SyntaxToken {
+    /// Gets the next sibling node or token.
+    ///
+    /// Returns `None` if there is no next sibling node or token.
+    pub fn next_sibling_or_token(&self) -> Option<SyntaxElement> {
+        // This should also be a constant-time access rather than having to iterate
+        // We need the offset relative to the start of the parent from the green node to
+        // do that
```
a-frantz, line 327:
```suggestion
        // This should also be a constant-time access rather than having to iterate.
        // We need the offset relative to the start of the parent from the green node to
        // do that.
```

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+    }
+}
+
+impl From<SyntaxNode> for SyntaxElement {
+    fn from(value: SyntaxNode) -> Self {
+        Self::Node(value)
+    }
+}
+
+/// Represents a syntax token that is `Send`+`Sync`.
```
a-frantz, line 316:
```suggestion
/// Represents a syntax token that is `Send + Sync`.
```

wdl-engine/src/tree.rs
```diff
@@ -0,0 +1,543 @@
...
+            index: 0,
+            offset: 0,
+        }
+    }
+}
+
+/// Represents an element in a syntax tree.
+pub type SyntaxElement = NodeOrToken<SyntaxNode, SyntaxToken>;
+
+/// Represents a syntax node that is `Send`+`Sync`.
```
a-frantz, line 71:
```suggestion
/// Represents a syntax node that is `Send + Sync`.
```

a-frantz (commented):
changelog entries?

peterhuene:
Yep, adding those shortly.

wdl-format/CHANGELOG.md
```diff
@@ -9,6 +9,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* Updated to use new `wdl-ast` API ([#355](https://github.com/stjude-rust-labs/wdl/pull/355)).
+* Refactored analysis API to support different syntax tree element
+  representations ([#355](https://github.com/stjude-rust-labs/wdl/pull/355)).
```
a-frantz, line 14:
```suggestion
```

a-frantz (approved):

PR #362, opened 2025-03-20
fix: lsp errors out when a file uri is received instead of directory

dead8309:
This pull request fixes the current way of handling URLs in `textDocument/didOpen` event

If i open a single `wdl` file the analyzer can't _add_ the file as currently the analyzer doesn't check if the `URL` it got is a file or a directory, it just assumes its a directory and calls `analyzer.add_directory()`  instead of `analyzer.add_document()`

Error log:

```
2026-03-20T01:10:20.308154Z ERROR wdl_lsp::server: failed to add document file:///home/kiit/dev/stjude-rust-labs/sprocket-vscode/syntaxes/example.wdl: `/home/kiit/dev/stjude-rust-labs/sprocket-vscode/syntaxes/example.wdl` is a file, not a directory
```

      changes (when appropriate).

### DIFF
--- a/wdl-lsp/src/server.rs
+++ b/wdl-lsp/src/server.rs
@@ -438,29 +438,27 @@ impl LanguageServer for Server {
 
         debug!("received `textDocument/didOpen` request: {params:#?}");
 
-        if let Ok(path) = params.text_document.uri.to_file_path() {
-            if let Err(e) = self
-                .analyzer
-                .add_document(path_to_uri(&path).expect("should convert to uri"))
-                .await
-            {
-                error!(
-                    "failed to add document {uri}: {e}",
-                    uri = params.text_document.uri
-                );
-                return;
-            }
+        if let Err(e) = self
+            .analyzer
+            .add_document(params.text_document.uri.clone())
+            .await
+        {
+            error!(
+                "failed to add document {uri}: {e}",
+                uri = params.text_document.uri
+            );
+            return;
+        }
 
-            if let Err(e) = self.analyzer.notify_incremental_change(
-                params.text_document.uri,
-                IncrementalChange {
-                    version: params.text_document.version,
-                    start: Some(params.text_document.text),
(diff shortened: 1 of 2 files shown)


wdl-lsp/src/server.rs
```diff
@@ -439,9 +439,31 @@ impl LanguageServer for Server {
...
+                    .await
+                {
+                    error!(
+                        "failed to add document {uri}: {e}",
+                        uri = params.text_document.uri
+                    );
+                    return;
+                }
+            } else if path.is_dir() {
+                // Handle individual directory
```
dead8309, line 456: Checked the [SPEC](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentItem) once again, do we really need to check for directories as only document uris will be received  ?

peterhuene: I'm fairly certain only files will be received and we can error in the case of a directory.

I think we can simply change the original call to `add_document` without having to read the file's metadata and let it error in analysis when opening it if it's not a file.

wdl-lsp/src/server.rs
```diff
@@ -439,7 +439,11 @@ impl LanguageServer for Server {
         debug!("received `textDocument/didOpen` request: {params:#?}");
 
         if let Ok(path) = params.text_document.uri.to_file_path() {
-            if let Err(e) = self.analyzer.add_directory(path).await {
+            if let Err(e) = self
+                .analyzer
+                .add_document(path_to_uri(&path).expect("should convert to uri"))
```
a-frantz, line 444: Should this just use the `text_document.uri` directly? We convert to a path then back to a URI.

dead8309: fixed!

a-frantz (approved):

peterhuene (approved):
@dead8309 thanks for the contribution!

PR #369, opened 2025-03-26
fix: improve handling of URLs in file functions.

peterhuene:
* allow `glob` to operate off of `file://` URLs.
* allow `size` to operate off of `file://` URLs.
* allow `basename` to operate off of any URLs.

This PR also removes custom diagnostics from the stdlib functions so that all of the standard library functions now use `function_called_failed` diagnostics for errors. It gives a more consistent diagnostic for function calls.

      changes (when appropriate).

### DIFF
--- a/wdl-engine/src/stdlib/find.rs
+++ b/wdl-engine/src/stdlib/find.rs
@@ -14,6 +14,9 @@ use crate::PrimitiveValue;
 use crate::Value;
 use crate::diagnostics::function_call_failed;
 
+/// The name of the function defined in this file for use in diagnostics.
+const FUNCTION_NAME: &str = "find";
+
 /// Given two String parameters `input` and `pattern`, searches for the
 /// occurrence of `pattern` within `input` and returns the first match or `None`
 /// if there are no matches.
@@ -31,7 +34,7 @@ fn find(context: CallContext<'_>) -> Result<Value, Diagnostic> {
         .unwrap_string();
 
     let regex = Regex::new(pattern.as_str())
-        .map_err(|e| function_call_failed("find", &e, context.arguments[1].span))?;
+        .map_err(|e| function_call_failed(FUNCTION_NAME, &e, context.arguments[1].span))?;
 
     match regex.find(input.as_str()) {
         Some(m) => Ok(PrimitiveValue::new_string(m.as_str()).into()),
--- a/wdl-engine/src/stdlib/find.rs
+++ b/wdl-engine/src/stdlib/find.rs
@@ -12,7 +12,7 @@ use super::Function;
 use super::Signature;
 use crate::PrimitiveValue;
 use crate::Value;
-use crate::diagnostics::invalid_regex;
+use crate::diagnostics::function_call_failed;
 
 /// Given two String parameters `input` and `pattern`, searches for the
 /// occurrence of `pattern` within `input` and returns the first match or `None`
@@ -30,8 +30,8 @@ fn find(context: CallContext<'_>) -> Result<Value, Diagnostic> {
         .coerce_argument(1, PrimitiveType::String)
         .unwrap_string();
 
-    let regex =
-        Regex::new(pattern.as_str()).map_err(|e| invalid_regex(&e, context.arguments[1].span))?;
+    let regex = Regex::new(pattern.as_str())
+        .map_err(|e| function_call_failed("find", &e, context.arguments[1].span))?;
 
     match regex.find(input.as_str()) {
         Some(m) => Ok(PrimitiveValue::new_string(m.as_str()).into()),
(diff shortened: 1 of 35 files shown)


wdl-engine/src/stdlib/size.rs
```diff
@@ -40,8 +61,17 @@ fn size(context: CallContext<'_>) -> BoxFuture<'_, Result<Value, Diagnostic>> {
...
 
-            unit.parse()
-                .map_err(|_| invalid_storage_unit(&unit, context.arguments[1].span))?
+            unit.parse().map_err(|_| {
+                function_call_failed(
+                    "size",
+                    format!(
+                        "invalid storage unit `{unit}`: supported units are `B`, `KB`, `K`, `MB`, \
+                         `M`, `GB`, `G`, `TB`, `T`, `KiB`, `Ki`, `MiB`, `Mi`, `GiB`, `Gi`, `TiB`, \
+                         and `Ti`",
```
a-frantz, line 70: should the supported units in this message be derived from something?

wdl-engine/src/stdlib/join_paths.rs
```diff
@@ -98,6 +134,42 @@ fn join_paths(context: CallContext<'_>) -> Result<Value, Diagnostic> {
...
 
+    if let Ok(mut url) = first.parse::<Url>() {
+        for (i, element) in array
+            .as_slice()
+            .iter()
+            .enumerate()
+            .skip(if skip { 1 } else { 0 })
+        {
+            let next = element.as_string().expect("element should be string");
+            if next.starts_with('/') || next.starts_with("//") | next.contains(":") {
```
claymcleod, line 145: Is it necessary to check both `/` and `//` starting patterns here?

wdl-engine/src/stdlib/find.rs
```diff
@@ -30,8 +30,8 @@ fn find(context: CallContext<'_>) -> Result<Value, Diagnostic> {
         .coerce_argument(1, PrimitiveType::String)
         .unwrap_string();
 
-    let regex =
-        Regex::new(pattern.as_str()).map_err(|e| invalid_regex(&e, context.arguments[1].span))?;
+    let regex = Regex::new(pattern.as_str())
+        .map_err(|e| function_call_failed("find", &e, context.arguments[1].span))?;
```
a-frantz, line 34: Could we replace all the `function_call_failed(<hardcoded str>,...` with something like `Self::function_name()`?

a-frantz (approved):
Two questions about using fewer hardcoded strings in the error messages. I'm not sure how feasible either request is, so maybe they can be ignored. Just trying to think of easing future maintenance.

Otherwise, LGTM

wdl-engine/src/stdlib/find.rs
peterhuene: It's a free function, so there's no `Self` context. At best perhaps a constant. I can do that.

wdl-engine/src/stdlib/size.rs
peterhuene: At best I think we could have an iterator on `StorageUnit` that iterates over this list, but then we'd have to format it. As this message is just moved from the now-removed diagnostic, I don't think this is necessary to be honest.

a-frantz: Ok. I'm just looking at a long hardcoded list and thinking "that could be annoying if the units ever change", but in practice that seems unlikely to happen. I'm fine with this as it is 👍

peterhuene:
CI is failing due to docker hub rate limiting.

We need to fix the pull image functionality in Crankshaft to gate to a single request if there are a number of concurrent tasks all requesting to pull the same image.

peterhuene:
As a workaround, I think we could do what bollard does in CI and use a local docker hub container that we've pushed any images we need for the tests, which will limit the pull requests to Docker Hub to only the number of images needed to run the tests.

a-frantz:
all that sounds good, but a perhaps simpler workaround is to pull from quay.io instead of docker hub to avoid rate limits entirely 🤷

peterhuene:
CI is failing on windows for the `join_paths` tests as absolute file paths on Windows are being successfully parsed as URLs (sigh).

peterhuene:
> all that sounds good, but a perhaps simpler workaround is to pull from quay.io instead of docker hub to avoid rate limits entirely 🤷

As far as I know, there's no way to tell the local docker daemon to use an alternative default namespace for the image names, so we'd have to run the workflow/task evaluation tests with an engine config that overrides the default container and also update any explicit container names in the test WDLs.

Not terribly difficult to do. Alternatively, I think if we pulled `ubuntu:latest` at the start of CI, that'd also do the trick (the daemon would already see it locally and not fetch it from hub).

claymcleod (approved):

PR #371, opened 2025-03-27
feat: Enable linking between related lint rules

dead8309:
This pull request adds support for linking related lint rules, which would allow `sprocket explain` to provide better guidance to users.

This change adds a `related_rules`  to the `Rule` trait.

- added debug assertions to validate related rule IDs during development builds.
- I've tried to add rules which seem related as a first pass

Before submitting this PR, please make sure:

- [x] You have added a few sentences describing the PR here.
- [x] You have added yourself or the appropriate individual as the assignee.
- [x] You have added at least one relevant code reviewer to the PR.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.
- [x] Your PR title follows the [conventional commit] style.

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.1.0/

PR #372, opened 2025-03-28
docs: more guidance on exceptable nodes

a-frantz:

Inspired by the documentation additions in (the currently under review) #371 

Adds some more guidance to CONTRIBUTING about how to define the `Rule::exceptable_nodes()` method of a new lint rule.

      changes (when appropriate).
- [ ] Your PR title follows the [conventional commit] style.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,13 +77,29 @@ There are a handful of reasons the CI may have turned red. Try the following fix
 4. review `source.errors` to see if it matches our expectations
     - this isn't exactly true, but I conceptualize `source.errors` as the output if a user ran a `lint` command on `source.wdl`
     - while reviewing, ask yourself if the printed diagnostics are clear and informative
+5. consider which nodes should be "exceptable" for this rule
+    - See "[How should I decide which nodes to add to `exceptable_nodes()`?](#how-should-i-decide-which-nodes-to-add-to-exceptable_nodes)" for more information on this process
 5. repeat
 
 ### Can you explain how rules use `exceptable_nodes()` and `exceptable_add()`?
 
 Every lint rule has an ID which can be used in lint directives (comments beginning with `#@ except:`) to prevent them from emitting diagnostics for portions of a WDL document. Rules "excepted" during the preamble (comments which are before the version statement) will be turned off for the entire document; otherwise, lint directives will shut off a rule while processing the children of whatever node it comes before, but only if that node is in the rule's `exceptable_nodes()` list. `exceptable_add()` will check all the ancestors of `element` for nodes that match the `exceptable_nodes()` list and see if they have a lint directive disabling the current rule; if so, the diagnostic will not be added to the validation output.
 
-#### Further reading
+### How should I decide which nodes to add to `exceptable_nodes()`?
+
+Returning `None` from `exceptable_nodes()` will enable lint directives to work anywhere in the document. This should be used sparingly, and is generally not the desired behavior. The most common case for returning `None` is if the lint rule pertains to "trivia" (whitespace and comments) which are permitted to appear almost anywhere in a WDL document.
+
+_Every_ rule that returns `Some(_)` for `exceptable_nodes()` should include `SyntaxKind::VersionStatementNode` as the first entry. This is to ensure the rule can be disabled for an entire document.
+
+The other `SyntaxKind` nodes that should be returned should be _intuitive_, _minimal_, and _comprehensive_.
+
+Intuitive: users shouldn't need to know anything about the internal CST or AST structure to determine where to add a lint directive. However, we can reasonably expect users to intuit the nested structure of a WDL document. There are top level items (workflow, task, and struct definitions), and then each of those top level items has a variety of sections contained within, which in turn may have further items contained within them. Excepting a rule somewhere "higher" in the nested structure will disable the rule for everything "beneath".
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -77,13 +77,29 @@ There are a handful of reasons the CI may have turned red. Try the following fix
...
+
+Minimal: We want to have as few exceptable nodes as possible for each rule, while still enabling the most intuitive locations to be excepted. If two nodes would cover the same set of potential diagnostics, _always_ prefer the more specific node. e.g. for a rule relating to `input` sections, the `InputSectionNode`, `TaskDefinitionNode`, and `WorkflowDefinitionNode` would all cover the exact same set of diagnostics (as there can be at most one input section for each task or workflow definition); so we would prefer to have `InputSectionNode` returned by `exceptable_nodes()` rather than either `TaskDefinitionNode` or `WorkflowDefinitionNode`.
+
+Comprehensive: Try to cover as many _unique_ sets of potential diagnostics as possible with the nodes returned by `exceptable_nodes()`. An individual diagnostic should be able to targeted by a lint directive, and so should any unique group of diagnostics.
```
adthrasher, line 100: > An individual diagnostic should be able to targeted by a lint directive, and so should any unique group of diagnostics.

I think this could be expanded. I wasn't clear on the first reading (and may not be now).

CONTRIBUTING.md
```diff
@@ -77,13 +77,29 @@ There are a handful of reasons the CI may have turned red. Try the following fix
...
+
+Intuitive: users shouldn't need to know anything about the internal CST or AST structure to determine where to add a lint directive. However, we can reasonably expect users to intuit the nested structure of a WDL document. There are top level items (workflow, task, and struct definitions), and then each of those top level items has a variety of sections contained within, which in turn may have further items contained within them. Excepting a rule somewhere "higher" in the nested structure will disable the rule for everything "beneath".
+
+Minimal: We want to have as few exceptable nodes as possible for each rule, while still enabling the most intuitive locations to be excepted. If two nodes would cover the same set of potential diagnostics, _always_ prefer the more specific node. e.g. for a rule relating to `input` sections, the `InputSectionNode`, `TaskDefinitionNode`, and `WorkflowDefinitionNode` would all cover the exact same set of diagnostics (as there can be at most one input section for each task or workflow definition); so we would prefer to have `InputSectionNode` returned by `exceptable_nodes()` rather than either `TaskDefinitionNode` or `WorkflowDefinitionNode`.
```
adthrasher, line 98: When I read this initially, I thought you meant that the two locations are equivalent. After writing out the example below to ask a question, I think you mean that `TaskDefinitionNode` and `WorkflowDefinitionNode` aren't places we want to except `input` rules, since there is only one `input` section, so we **prefer** to require users to except on it. I'm not sure if that is completely in-line with _intuitive_, but I'm OK with it. This block of text might need a little refining to ensure we're communicating that.

```
#@ except: InputSorted <-- TaskDefinitionNode
task foo {
   #@ except: InputSorted <-- InputSectionNode
   input {}
}

a-frantz: Can you be more specific about what you found confusing? It reads fine to me, but I did write it so I'm not a good judge 😅 

> I think you mean that `TaskDefinitionNode` and `WorkflowDefinitionNode` aren't places we want to except `input` rules, since there is only one `input` section, so we **prefer** to require users to except on it.

this is what I was trying to convey

> I'm not sure if that is completely in-line with _intuitive_, but I'm OK with it. This block of text might need a little refining to ensure we're communicating that.

Yea, maybe I should add a sentence that at times these 3 (intuitive, minimal, and comprehensive) are at odds with each other? I suppose they are, but I'm not sure how to communicate that without making it more confusing.

CONTRIBUTING.md
a-frantz: Should I have an example here? The one coming to mind are the rules related to declarations, which can appear in a lot of places, and thus can have a decent amount of unique ways to "except" different groups of decls

CONTRIBUTING.md
adthrasher: I've reread this now a number of times and I can't decide exactly what I found confusing initially. I considered suggesting something like "Note: minimal may conflict with intuitive", but I'm not convinced that is better.

CONTRIBUTING.md
adthrasher: > Try to cover as many _unique_ sets of potential diagnostics as possible with the nodes returned by `exceptable_nodes()`.

Can you clarify what you mean by "unique sets"?  I think the second sentence is clear enough and an example would be noisy.

a-frantz: e.g. https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/rules/disallowed_declaration_name.rs#L57

All of the nodes returned here are potentially unique. The task and workflow definition nodes are a superset ofthe input node, output node, and private decls.

CONTRIBUTING.md
a-frantz: > so we **prefer** to require users to except on it

re-reading your comment, and need to point it it's actually a _requirement_, not just a preference. Placing a lint directive on a node that isn't returned by `exceptable_nodes()` results in it being ignored (unless `exceptable_nodes()` returns `None` for that rule). It will also fire off a `MisplacedLintDirective` diagnostic, so it's not silently ignored.

PR #381, opened 2025-04-01
Fix: add chown workaround after task execution

dead8309:
This pr ensures files created by tasks using `crankshaft` backend are owned by the correct user by chowning the output directory after execution.

      changes (when appropriate).


wdl-engine/tests/workflows/chown_cleanup_docker/source.wdl
dead8309, line 1: Maybe having them doesn't make sense ? They can never pass on mac and windows

a-frantz:
@dead8309 does this only `chown` things marked `output`? We need to `chown` everything potentially created by the container (which should be limited to the work directory, as that's the only directory mounted with write permissions).

dead8309:
> @dead8309 does this only `chown` things marked `output`? .

yea After execution we pass the `output dir` and it `chown` that recursively 

> We need to `chown` everything potentially created by the container (which should be limited to the work directory, as that's the only directory mounted with write permissions)

So both `workdir` and `output` directory

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -365,4 +371,153 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
+        let inner_backend = self.inner.clone();
+        let generator = self.generator.clone();
+        let output_path = output_dir.to_path_buf();
+
+        tokio::spawn(async move {
+            let result = async {
+                let (uid, gid) = get_uid_gid();
+                let ownership = format!("{uid}:{gid}");
+                info!(
+                    "Cleanup target: '{}', Attempting to set ownership to: {}",
```
claymcleod, line 392:
```suggestion
                    "cleanup target: '{}', attempting to set ownership to: {}",
```

wdl-engine/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Added retry logic for task execution ([#320](https://github.com/stjude-rust-labs/wdl/pull/320)).
 * Added a `Config` type for specifying evaluation configuration ([#320](https://github.com/stjude-rust-labs/wdl/pull/320)).
 * Added progress callback to `WorkflowEvaluator` ([#310](https://github.com/stjude-rust-labs/wdl/pull/310)).
+* Added `chown` workaround to `docker backend` after task execution ([#379](https://github.com/stjude-rust-labs/wdl/pull/379)).
```
adthrasher, line 29: This should be in the `unreleased` section. I'm assuming this was an artifact of a merge or rebase.

dead8309:
signed all commits

dead8309:
If I'm not wrong, i think it covers what was discussed in the slack channel - `chown'ing the whole output directory`. Are there any changes that needs to be made ?

@a-frantz @peterhuene

wdl-engine/Cargo.toml
```diff
@@ -41,6 +41,9 @@ reqwest = { workspace = true }
 reqwest-middleware = { workspace = true }
 dirs = { workspace =  true }
 
+[target.'cfg(target_os = "linux")'.dependencies]
```
a-frantz, line 44: @peterhuene is this right? I think in other places we have conditional compilation we do `windows` and `not(windows)`. Is this issue not present on MacOS?

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -458,4 +464,153 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
+                    generator
+                        .lock()
+                        .expect("generator should always acquire")
+                        .next()
+                        .expect("generator should never be exhausted")
+                );
+
+                let cleanup_resources = Resources::builder().cpu(0.1).ram(0.05).build();
+
+                let cleanup_task = Task::builder()
```
adthrasher, line 522: Should this be defined by the inner backend? Can we think of any other backend-specific cleanup steps that would need to be run?

dead8309: Can do. But shouldn't we have as slim resources as possible for cleanup

adthrasher: Ah, the comment is meant to be for the definition of the `cleanup_task`. The recursive `chown` is specific to Docker. My question was more general for the group as to whether we expect a necessary cleanup for other backends.

a-frantz: We may one day want to add a "remove intermediate files" config option that on success of a workflow purges the filesystem of anything not saved as an `output`. That seems a long way off, but is in theory a use for this cleanup func?

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -458,4 +464,153 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
+                let cleanup_task_name = format!(
+                    "wdl-engine-chown-cleanup-{}",
+                    generator
+                        .lock()
+                        .expect("generator should always acquire")
+                        .next()
+                        .expect("generator should never be exhausted")
+                );
+
+                let cleanup_resources = Resources::builder().cpu(0.1).ram(0.05).build();
```
a-frantz, line 520: Can you make these `const` instead of the magic numbers here?

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -471,4 +486,154 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
+                        "output directory `{path}` is not a directory",
+                        path = output_path.display()
+                    );
+                }
+
+                let output_mount = Input::builder()
+                    .path(GUEST_OUT_DIR)
+                    .contents(Contents::Path(output_path.clone()))
+                    .ty(Type::Directory)
+                    // need write acces
```
adthrasher, line 527:
```suggestion
                    // need write access
```

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -66,6 +67,15 @@ const GUEST_WORK_DIR: &str = "/mnt/work";
 /// The guest path for the command file.
 const GUEST_COMMAND_PATH: &str = "/mnt/command";
 
+/// The guest path for the output directory.
+const GUEST_OUT_DIR: &str = "/workflow_output";
```
adthrasher, line 71: @peterhuene  - Are these paths not defined somewhere already that can be accessed here?

wdl-engine/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Added
 
 * Implemented remote file localization for task execution (#[386](https://github.com/stjude-rust-labs/wdl/pull/386)).
+* Added `chown` workaround to `docker backend` after task execution ([#379](https://github.com/stjude-rust-labs/wdl/pull/379)).
```
a-frantz, line 13: This should be under `Fixed` and should mention the actual issue being addressed (users not having permissions for files created by a root docker user). I don't think `chown` needs to be mentioned at all in the CHANGELOG (implementation detail that users reading the changelog don't need to care about).

dead8309:
rebased

a-frantz (approved):
Looks good to me, but I'd want @peterhuene 's blessing on these changes before we merge.

wdl-engine/Cargo.toml
peterhuene: Yes, it is also a macOS issue as well.

wdl-engine/src/backend.rs
```diff
@@ -319,6 +319,9 @@ pub trait TaskExecutionBackend: Send + Sync {
         request: TaskSpawnRequest,
         token: CancellationToken,
     ) -> Result<TaskExecutionEvents>;
+
+    /// Performs cleanup operations after workflow/task evaluation completes.
+    fn cleanup(&self, output_dir: &Path) -> Result<Receiver<Result<()>>>;
```
peterhuene, line 324: I think `cleanup` should be "best-effort" and as a result infallible; if the individual cleanup operations want to internally log a failure, I think that'd be fine, but I don't think we ever want to fail evaluation just because `cleanup` did.

I think we can probably just have this method be:

```rust
/// Performs cleanup operations after top-level workflow or task evaluation completes.
///
/// Returns `None` if no cleanup is required.
fn cleanup<'a, 'b, 'c>(&'a self, output_dir: &'b Path) -> Option<BoxFuture<'c, Output = ()>> where 'a: 'c, 'b: 'c, Self: 'c { None }
```

And then for the Crankshift backend we can override this (but still return `None` if not using docker) to return 

```rust
Some(async move {
   // .. cleanup here
}.boxed())
```

Then there'd also be no need to spawn, we'd just:

```rust
if let Some(fut) = backend.cleanup() {
    fut.await;
}
```

wdl-engine/src/backend/crankshaft.rs
peterhuene: This particular path is the guest path specific to the cleanup operation, so it probably belongs here and can be its own value.

wdl-engine/src/backend/crankshaft.rs
peterhuene: I think we can punt on this as my support for TES will actually break apart the Crankshaft backend into multiple `wdl-engine` (`docker`, `tes`, etc.) backends as they're different enough in terms of their behavior to require it.

Regarding its potential use to clean up "intermediate" files, I think we could build upon this to accept cleanup options that the backends can use to figure out what needs to be cleaned up (in the case of Docker, it can delete files in the local `work` directory; for TES that might mean making HTTP requests to delete cloud storage files, etc.).

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -471,4 +486,154 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
+
+                info!(
+                    "running cleanup task '{}' to chown '{}' to '{}'",
+                    cleanup_task_name,
+                    output_path.display(),
+                    ownership
+                );
+
+                let (spawned_tx, _) = oneshot::channel();
+                let token = CancellationToken::new();
```
peterhuene, line 571: We should probably be passing in the same cancellation token that we're using for evaluation rather than this new one, as there would be no way for the caller to signal the token as cancelled. 

That is to say we should pass the token in as an argument to this function.

wdl-engine/src/backend/local.rs
```diff
@@ -337,4 +338,10 @@ impl TaskExecutionBackend for LocalTaskExecutionBackend {
             completed: completed_rx,
         })
     }
+
+    fn cleanup(&self, _output_dir: &Path) -> Result<oneshot::Receiver<Result<()>>> {
```
peterhuene, line 342: If we take my suggestion for a default implementation of `cleanup` to return `None`, we can delete this change.

wdl-engine/src/eval/v1/workflow.rs
```diff
@@ -804,28 +807,71 @@ impl WorkflowEvaluator {
...
+                subgraph,
+                max_concurrency,
+                Arc::new(id.to_string()),
+                progress,
+            )
+            .await
+        };
 
-        Ok(outputs)
+        let cleanup_result = match self.backend.cleanup(&effective_output_dir) {
```
peterhuene, line 822: I think a lot of the logging logic here can ultimately be removed and we would just call:

```rust
let evaluation_result = ...;

if let Some(cleanup) = self.backend.cleanup(&effective_output_dir) {
    cleanup.await;
}

let mut outputs = ...;
Ok(outputs)
```

The `cleanup` implementation could potentially log any problems, but I don't want `cleanup` to cause `sprocket run` to fail if it ran into a problem.

peterhuene (commented):
Hi @dead8309.

This looks great, just a few feedback comments.

wdl-engine/Cargo.toml
dead8309: using `libc` now directly 👍

peterhuene (approved):
Looks good. Thanks @dead8309 again for another contribution!

Issue #253, opened 2024-11-12
Implement an analysis check that statically validates a string literal `pattern` argument to `find`, `matches`, and `sub` is a valid regex.


peterhuene:
If the `pattern` argument to the `find`, `matches`, and `sub` WDL standard library functions are string literals, we can easily check to see that the pattern compiles as a regex and, if not, emit a diagnostic. It's quite common to use a string literal for the pattern too.

I'm leaning towards doing that in a lint rather than analysis, although a lint warning seems wrong as it could lead to a runtime evaluation error.

peterhuene:
Another reason I don't want this in analysis itself is that compiling regular expressions can be expensive depending on their complexity; as we _must_ compile it for evaluation, I don't want to compile it twice either (once for analysis and then once again for evaluation).

a-frantz:
This makes more sense in analysis to me. IMO evaluating it twice is an acceptable redundancy. Main reason for me is that I would want it emitted during a run of `sprocket check` when lints aren't active.

peterhuene:
I guess I could implement a cache in each analyzed document where you can ask it for an already-compiled regex for a particular expression.

Ok, makes sense to stick this in analysis then.

dead8309:
hey @Pabsthegreat if you are not currently working on this, i'd like to get assigned @peterhuene @a-frantz

a-frantz:
@dead8309 assigned!

PR #404, opened 2025-04-08
feat: validate regular expressions on the go

dead8309:
Adds static validation of regex and show diagnostics if its invalid.

closes: #253 

      changes (when appropriate).

### DIFF
--- a/wdl-analysis/src/diagnostics.rs
+++ b/wdl-analysis/src/diagnostics.rs
@@ -794,14 +794,9 @@ pub fn invalid_placeholder_option<N: TreeNode>(
 }
 
 /// Creates an invalid regex pattern diagnostic.
-pub fn invalid_regex_pattern(
-    function_name: &str,
-    pattern: &str,
-    error: &regex::Error,
-    span: Span,
-) -> Diagnostic {
+pub fn invalid_regex_pattern(function: &str, pattern: &str, error: &str, span: Span) -> Diagnostic {
     Diagnostic::error(format!(
-        "invalid regex pattern \"{pattern}\" for function `{function_name}`: {error}"
+        "invalid regex pattern `{function}`: `{pattern}` is not a valid regular expression"
     ))
-    .with_label("this pattern is not a valid regular expression", span)
+    .with_label(format!("error: {error}"), span)
 }
--- a/wdl-analysis/src/diagnostics.rs
+++ b/wdl-analysis/src/diagnostics.rs
@@ -792,3 +792,16 @@ pub fn invalid_placeholder_option<N: TreeNode>(
 
     Diagnostic::error(message).with_label(format!("this is type `{ty}`"), span)
 }
+
+/// Creates an invalid regex pattern diagnostic.
+pub fn invalid_regex_pattern(
+    function_name: &str,
+    pattern: &str,
+    error: &regex::Error,
+    span: Span,
+) -> Diagnostic {
+    Diagnostic::error(format!(
+        "invalid regex pattern \"{pattern}\" for function `{function_name}`: {error}"
+    ))
+    .with_label("this pattern is not a valid regular expression", span)
+}
--- a/wdl-analysis/src/diagnostics.rs
+++ b/wdl-analysis/src/diagnostics.rs
@@ -794,9 +794,14 @@ pub fn invalid_placeholder_option<N: TreeNode>(
 }
 
 /// Creates an invalid regex pattern diagnostic.
-pub fn invalid_regex_pattern(function: &str, pattern: &str, error: &str, span: Span) -> Diagnostic {
+pub fn invalid_regex_pattern(
+    function: &str,
+    pattern: &str,
+    error: &regex::Error,
+    span: Span,
+) -> Diagnostic {
     Diagnostic::error(format!(
-        "invalid regex pattern `{function}`: `{pattern}` is not a valid regular expression"
+        "invalid regular expression `{pattern}` used in function `{function}`: {error}"
     ))
-    .with_label(format!("error: {error}"), span)
+    .with_label("invalid regular expression", span)
 }
--- a/wdl-analysis/src/types/v1.rs
+++ b/wdl-analysis/src/types/v1.rs
@@ -1,6 +1,5 @@
 //! Type conversion helpers for a V1 AST.
 
-use std::collections::HashMap;
 use std::fmt;
 use std::fmt::Write;
 use std::sync::Arc;
@@ -533,14 +532,6 @@ pub struct ExprTypeEvaluator<'a, C> {
     /// If the count is non-zero, special evaluation behavior is enabled for
     /// string interpolation.
     placeholders: usize,
-
-    /// A cache of regexes that have been evaluated.
-    ///
-    /// This is used to avoid re-evaluating the same regex multiple times.
-    ///
-    /// The key is the regex string and the value is either an error or
-    /// a compied regex.
-    regex_cache: HashMap<String, Result<regex::Regex, String>>,
 }
 
 impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
@@ -549,7 +540,6 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
         Self {
             context,
             placeholders: 0,
-            regex_cache: HashMap::new(),
         }
     }
 
@@ -1443,10 +1433,8 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
                 // Evaluate the argument expressions
                 let mut count = 0;
                 let mut arguments = [const { Type::Union }; MAX_PARAMETERS];
-                let mut arguments_node = Vec::with_capacity(expr.arguments().count());
 
                 for arg in expr.arguments() {
-                    arguments_node.push(arg.clone());
                     if count < MAX_PARAMETERS {
                         arguments[count] = self.evaluate_expr(&arg).unwrap_or(Type::Union);
                     }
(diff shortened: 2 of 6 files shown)


a-frantz:
@dead8309 please look into this panic in gauntlet
```
thread '<unnamed>' panicked at wdl-analysis/src/types/v1.rs:1712:22:
type should be an array
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Rayon: detected unexpected panic; aborting

thread '<unnamed>' panicked at /home/runner/work/wdl/wdl/wdl-analysis/src/rayon.rs:48:34:
failed to receive from oneshot channel: RecvError(())
/home/runner/work/_temp/add27558-c568-4527-8320-782066c7ce0d.sh: line 1:  1[917](https://github.com/stjude-rust-labs/wdl/actions/runs/14342715511/job/40205721825?pr=404#step:5:918) Aborted                 (core dumped) cargo run --release --bin gauntlet
Error: Process completed with exit code 134.
```
start a thread on slack if you get stuck!

dead8309:
> please look into this panic in gauntlet

done @a-frantz

wdl-analysis/src/types/v1.rs
```diff
@@ -531,6 +533,14 @@ pub struct ExprTypeEvaluator<'a, C> {
...
     /// string interpolation.
     placeholders: usize,
+
+    /// A cache of regexes that have been evaluated.
+    ///
+    /// This is used to avoid re-evaluating the same regex multiple times.
+    ///
+    /// The key is the regex string and the value is either an error or
+    /// `Ok(())`.
+    regex_cache: HashMap<String, Result<(), String>>,
```
a-frantz, line 543: this doesn't actually save much compute time. Instead of caching `Ok(())`, we should cache the _compiled regex_ (IDK what that Rust type is, but it's something offered by the `regex` crate).

The expensive operation here is (potentially) compiling the regex (if it's sufficiently complicated, which is admittedly a pretty niche case). Once it's compiled, using the RE to search text is a relatively light operation. So we want to compile once to ensure it's valid, and then cache the compiled RE for re-use during `run` operations.

If someone reviewing this knows more than me and something I said here is wrong, please correct :)

dead8309:
Screenshot_20250410-065703_Chrome
github decided to not process my commit for three hours straight. force pushing now

a-frantz (approved):
LGTM 👍

adthrasher (approved):

wdl-analysis/src/diagnostics.rs
```diff
@@ -792,3 +792,11 @@ pub fn invalid_placeholder_option<N: TreeNode>(
 
     Diagnostic::error(message).with_label(format!("this is type `{ty}`"), span)
 }
+
+/// Creates an invalid regex pattern diagnostic.
+pub fn invalid_regex_pattern(function: &str, pattern: &str, error: &str, span: Span) -> Diagnostic {
+    Diagnostic::error(format!(
+        "invalid regex pattern `{function}`: `{pattern}` is not a valid regular expression"
```
peterhuene, line 799: This will require updating the test baseline:

```suggestion
        "invalid regular expression `{pattern}` used in function `{function}`: {error}"
```

wdl-analysis/src/diagnostics.rs
```diff
@@ -792,3 +792,11 @@ pub fn invalid_placeholder_option<N: TreeNode>(
...
+    ))
+    .with_label(format!("error: {error}"), span)
```
peterhuene, line 801:
```suggestion
    .with_label("invalid regular expression", span)
```

wdl-analysis/src/types/v1.rs
```diff
@@ -531,6 +533,14 @@ pub struct ExprTypeEvaluator<'a, C> {
...
     /// string interpolation.
     placeholders: usize,
+
+    /// A cache of regexes that have been evaluated.
+    ///
+    /// This is used to avoid re-evaluating the same regex multiple times.
+    ///
+    /// The key is the regex string and the value is either an error or
+    /// a compied regex.
+    regex_cache: HashMap<String, Result<regex::Regex, String>>,
```
peterhuene, line 543: The lifetime of `ExprTypeEvaluator` is typically pretty short so this cache won't live for very long.

I wonder if we could keep the cache in `populate_document` as a `HashMap<&str, Result<regex::Regex, regex::Error>>` and pass that on down in places where we construct an `ExprTypeEvaluator`.

I'm also wondering if it's truly worth having a cache as it only saves us on regular expressions that are repeatedly used throughout a single WDL document, and I'm unsure if it's worth the effort now, to be honest.

wdl-analysis/src/diagnostics.rs
```diff
@@ -792,3 +792,11 @@ pub fn invalid_placeholder_option<N: TreeNode>(
 
     Diagnostic::error(message).with_label(format!("this is type `{ty}`"), span)
 }
+
+/// Creates an invalid regex pattern diagnostic.
+pub fn invalid_regex_pattern(function: &str, pattern: &str, error: &str, span: Span) -> Diagnostic {
```
peterhuene, line 797: We can probably have `error` be `&regex::Error` as we don't need to create a string from it.

wdl-analysis/src/types/v1.rs
```diff
@@ -1432,14 +1443,47 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
                 // Evaluate the argument expressions
                 let mut count = 0;
                 let mut arguments = [const { Type::Union }; MAX_PARAMETERS];
+                let mut arguments_node = Vec::with_capacity(expr.arguments().count());
```
peterhuene, line 1446: I don't think we need this `Vec`.

We can replace it with a call to `if let Some(argument) = expr.arguments().nth(1)`.

wdl-analysis/src/types/v1.rs
```diff
@@ -1432,14 +1443,47 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
...
+
                 for arg in expr.arguments() {
+                    arguments_node.push(arg.clone());
                     if count < MAX_PARAMETERS {
-                        arguments[count] = self.evaluate_expr(&arg)?;
+                        arguments[count] = self.evaluate_expr(&arg).unwrap_or(Type::Union);
```
peterhuene, line 1451: Thanks for catching this; we should definitely be treating the argument as `Union`.

wdl-analysis/src/types/v1.rs
```diff
@@ -1478,20 +1522,22 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
...
-                            self.context.add_diagnostic(argument_type_mismatch(
-                                target.text(),
-                                &expected,
-                                &arguments[index],
-                                expr.arguments()
-                                    .nth(index)
-                                    .map(|e| e.span())
-                                    .expect("should have span"),
-                            ));
+                            if index < arguments.len() {
```
peterhuene, line 1530: It should be guaranteed that `index < arguments.len()`, so I would rather panic when that's not the case than silently not emit a diagnostic.

I think we can revert this check.

wdl-analysis/src/types/v1.rs
```diff
@@ -1643,11 +1689,11 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
         severity: Severity,
     ) {
         let (label, span, fix) = match target.text() {
-            "select_first" => {
-                let ty = &arguments[0]
-                    .as_array()
-                    .expect("type should be an array")
-                    .element_type;
+            "select_first" if !arguments.is_empty() => {
```
peterhuene, line 1692: `check_unnecessary_call` is only ever called upon a successful binding of a function, which means it can make assumptions about the number of arguments and their types here.

Like above, I'd rather panic if the invariant is broken rather than silently not emit a diagnostic.

Therefore, I think we can revert the changes to this function.

wdl-analysis/src/types/v1.rs
peterhuene: Note that the actual WDL stdlib function implementations in the engine also don't currently do any sort of interning of the regular expressions passed to it, so I am leaning more towards just removing the cache in analysis until such time that it proves necessary.

wdl-analysis/src/types/v1.rs
peterhuene: So I think the panics in the tests are due to this change, where we're now binding to a function even if we can't evaluate one of its arguments.

I think we'll need to change `check_unnecessary_call` to accept things if the argument's type is `Union`.

wdl-analysis/src/types/v1.rs
```diff
@@ -1644,10 +1670,11 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
     ) {
         let (label, span, fix) = match target.text() {
             "select_first" => {
-                let ty = &arguments[0]
+                let ty = arguments[0]
```
peterhuene, line 1673: Suggested fix is to just `if let Some(ty) = arguments[0].as_array().map(|a| a.element_type()) { ... }`

peterhuene: For the check for `defined` below, we'll also want to not figure the diagnostic if the argument is also `Union`, so ` arguments[0].is_optional() || arguments[0].is_union()`.

wdl-analysis/src/types/v1.rs
```diff
@@ -1644,10 +1670,11 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
...
                     .as_array()
                     .expect("type should be an array")
-                    .element_type;
+                    .element_type();
+
                 if ty.is_optional() {
```
peterhuene, line 1678:
```suggestion
                if ty.is_optional() || ty.is_union() {
```

wdl-analysis/src/types/v1.rs
```diff
@@ -1644,37 +1668,35 @@ impl<'a, C: EvaluationContext> ExprTypeEvaluator<'a, C> {
     ) {
         let (label, span, fix) = match target.text() {
             "select_first" => {
-                let ty = &arguments[0]
-                    .as_array()
-                    .expect("type should be an array")
-                    .element_type;
-                if ty.is_optional() {
+                if let Some(ty) = arguments[0].as_array().map(|a| a.element_type()) {
+                    if ty.is_optional() {
```
peterhuene, line 1672:
```suggestion
                    if ty.is_optional() || ty.is_union() {
```

peterhuene (approved):
Looks great @dead8309! Thanks for the contribution.

Issue #388, opened 2025-04-02
feat: implement concurrent file downloads for file localization


peterhuene:
With the changes in #386, task execution backends are now requested to localize remote files if necessary.

Currently the backends download one input file at a time.

We should change this to `tokio::spawn` as many downloads as required to localize all of the remote input files and await them all to complete.

As part of this feature, we should add an engine configuration option under `HttpConfig` for the maximum number of concurrent downloads (a reasonable default might be `10`). The `HttpDownloader` implementation should internally keep a semaphore that is based on this configuration value and request a permit for each call to `download` that reaches the `get` method.

MrigankoSarkar:
Hi! 👋

I’m interested in contributing to this issue "feat: implement concurrent file downloads for file localization #388." I’d love to help implement concurrent file downloads using tokio::spawn and optimize the localization process. Additionally, I can work on adding the engine configuration option under HttpConfig to set a maximum number of concurrent downloads, along with implementing a semaphore in HttpDownloader to manage the concurrency limit effectively.

Please let me know if this issue is still open for contribution and if there are any specific guidelines or expectations I should follow. Looking forward to collaborating!

Best,
Mriganko (GSOC 2025 Contributor)

a-frantz:
You're all set to start working on this! Please make sure you have read the [contributing guide](https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md) and follow **all* instructions in the PR template when you open one!

Gyan-max:
Hi @peterhuene @a-frantz 
I saw this issue is already assigned but I'm really interested in working on this too. 
Please consider me as backup.

MrigankoSarkar:
> Hi [@peterhuene](https://github.com/peterhuene) [@a-frantz](https://github.com/a-frantz) I saw this issue is already assigned but I'm really interested in working on this too. Please consider me as backup.

@Gyan-max 
I’ve already started contributing to this issue, so it might be best if you explore one of the other open issues—there are plenty of great problems to work on and contribute to.

whitewolf2000ani:
Hey @MrigankoSarkar @peterhuene  
Is this issue still open to be solved, If not would like to collaborate and work on the issue.

whitewolf2000ani:
Great, good to hear that you are working on the issue.
Please Link the PR to clear out any confusion.

Thank you.

adthrasher:
Hi @MrigankoSarkar,

[...]

PR #424, opened 2025-04-18
feat: concurrent file downloads for localization

dead8309:
closes: #388 

For external contributors:

For all contributors:

      changes (when appropriate).


dead8309:
test runs with 50 remote files

#### Before

- Uncached

```bash
➜  wdl git:(feat/concurrent-downloads) time cargo test --package wdl-engine --features="codespan" --test workflows -- -- concurrent-localization
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.19s
     Running tests/workflows.rs (target/debug/deps/workflows-719bfd71ea48623c)

running 1 tests

test concurrent-localization ... ok

test result: ok. 1 passed

cargo test --package wdl-engine --features="codespan" --test workflows -- --   0.86s user 0.25s system 15% cpu 7.383 total
```

- Cached

```bash
➜  wdl git:(feat/concurrent-downloads) time cargo test --package wdl-engine --features="codespan" --test workflows -- -- concurrent-localization
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.20s
     Running tests/workflows.rs (target/debug/deps/workflows-719bfd71ea48623c)

running 1 tests

test concurrent-localization ... ok

test result: ok. 1 passed

cargo test --package wdl-engine --features="codespan" --test workflows -- --   0.62s user 0.16s system 35% cpu 2.196 total
```

---

### After

- Uncached
```bash

➜  wdl git:(feat/concurrent-downloads) ✗ time cargo test --package wdl-engine --features="codespan" --test workflows -- -- concurrent-localization
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.18s
     Running tests/workflows.rs (target/debug/deps/workflows-719bfd71ea48623c)

running 1 tests

test concurrent-localization ... ok

test result: ok. 1 passed

cargo test --package wdl-engine --features="codespan" --test workflows -- --   0.93s user 0.31s system 38% cpu 3.265 total
```

- Cached

```bash
➜  wdl git:(feat/concurrent-downloads) ✗ time cargo test --package wdl-engine --features="codespan" --test workflows -- -- concurrent-localization
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.18s
     Running tests/workflows.rs (target/debug/deps/workflows-719bfd71ea48623c)

running 1 tests

test concurrent-localization ... ok

test result: ok. 1 passed

cargo test --package wdl-engine --features="codespan" --test workflows -- --   0.67s user 0.16s system 45% cpu 1.832 total
```

dead8309:
tldr;
| Type | Uncached Time | Cached Time |
| -----|----------------|--------------|
| Before | 7.383s | 2.196s |
| After | 3.265s | 1.832s |

wdl-engine/tests/workflows/concurrent-localization/source.wdl
dead8309, line 1: Should i remove this test workflow, considering `wdl-engine/tests/localization` already exists (but doesn't test on large number of files)

claymcleod:
I think the remaining CI issue is going to be fixed by rebasing—would you please do that at your convenience?

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -380,41 +382,79 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
             for input in inputs.iter() {
                 trie.insert(input)?;
             }
 
             for (index, guest_path) in trie.calculate_guest_paths(GUEST_INPUTS_DIR)? {
-                inputs[index].set_guest_path(guest_path);
+                if let Some(input) = inputs.get_mut(index) {
+                    input.set_guest_path(guest_path);
+                } else {
+                    return Err(anyhow!("invalid index {} returned from trie", index));
```
claymcleod, line 406: You should `bail!()` here, as it's a bit more idiomatic.

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -397,41 +399,79 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
-                let location = match input.path() {
-                    EvaluationPath::Local(path) => Location::Path(path.into()),
-                    EvaluationPath::Remote(url) => downloader
-                        .download(url)
-                        .await
-                        .map_err(|e| anyhow!("failed to localize `{url}`: {e:?}"))?,
-                };
+            let mut download_futs = JoinSet::new();
+
+            let urls_to_download: Vec<(usize, Url)> = inputs
```
claymcleod, line 431: Here, I would recommend just iterating through the inputs mutably once and matching on the `EvaluationPath`:

- If it's an `EvaluationPath::Remote`, you can spawn the task to your `JoinSet` and collect the result later.
- If it's an `EvaluationPath::Local`, you set the location in place.

Upon completion of the first pass through, you can then do what you're doing in line 258. In other words, I these first three sections can be condensed down into one pass.

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -397,41 +399,79 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
+            while let Some(result) = download_futs.join_next().await {
+                match result {
+                    Ok(Ok((idx, location))) => {
+                        inputs
+                            .get_mut(idx)
+                            .expect("index from should be valid")
+                            .set_location(location);
+                    }
+                    Ok(Err(e)) => {
+                        download_futs.abort_all();
```
claymcleod, line 467: Something where we might need to pick a lane on—a within the docs on `abort_all()` says this.

> This does not remove the tasks from the `JoinSet`. To wait for the tasks to complete cancellation, you should call `join_next` in a loop until the JoinSet is empty.

So `abort_all()` is a bit of a no-op without actually going through and emptying the `JoinSet`. On the other hand, the docs for `JoinSet` itself say this:

> When the JoinSet is dropped, all tasks in the JoinSet are immediately aborted.

I kind of view what we have on the page now as in-between these two realities of not caring about cleaning up the aborted futures and cleaning them up. I see no need for the latter, so I'd honest just recommend removing the call to `abort_all()` and maybe documenting explicitly that the other tasks are aborted when `download_futs` is dropped.

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -397,41 +399,78 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
+            while let Some(result) = download_futs.join_next().await {
+                match result {
+                    Ok(Ok((idx, location))) => {
+                        inputs
+                            .get_mut(idx)
+                            .expect("index from should be valid")
+                            .set_location(location);
+                    }
+                    Ok(Err(e)) => {
+                        // futures are aborted when the JoinSet is dropped
```
claymcleod, line 467:
```suggestion
                        // Futures are aborted when the `JoinSet` is dropped.
```

Should also go in the next block for `Err(e)`.

wdl-engine/src/backend/local.rs
```diff
@@ -269,39 +271,78 @@ impl TaskExecutionBackend for LocalTaskExecutionBackend {
...
-                    .with_context(|| {
-                        format!("path `{path}` is not UTF-8", path = location.display())
-                    })?
-                    .to_string();
-
-                // Set the guest path to the download location for path translation
-                let location = location.into_owned();
-                input.set_guest_path(guest_path);
-                input.set_location(location);
+        async move {
```
claymcleod, line 284: We should obviously come back and get this one too once we settle on a solution in `crankshaft.rs`.

wdl-engine/src/config.rs
```diff
@@ -88,11 +91,21 @@ pub struct HttpConfig {
...
+    /// Defaults to 10.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub max_concurrent_downloads: Option<u64>,
 }
 
 impl HttpConfig {
     /// Validates the HTTP configuration.
     pub fn validate(&self) -> Result<()> {
+        if let Some(limit) = self.max_concurrent_downloads {
+            if limit == 0 {
```
claymcleod, line 105: This needs a test to ensure it fails if `limit` is `0`.

claymcleod: And that it is valid above that.

wdl-engine/src/http.rs
```diff
@@ -338,6 +350,13 @@ impl Downloader for HttpDownloader {
                 continue;
             }
 
+            let permit = self
```
claymcleod, line 353:
```rust
let permit = self
    .semaphore
    .acquire()
    .await
    .expect("semaphore should not be closed");
```

This should be sufficient here, you don't need anything more than a standard semaphore permit.

claymcleod:
With this change, I think it's also prudent to go ahead and handle some rudimentary concept of retries within the `Downloader`. This is mainly because I think it's more likely that these are going to fail due to network bandwidth issues. My suggestion is to acquire the permit _within_ the loop (getting a new permit on each retry) so that you fairly share downloading bandwidth amongst the files (and, perhaps give the particular files a second to breathe if the download has just failed. It's unclear how often that will make a difference, but it can't hurt I would say).

wdl-engine/tests/workflows/concurrent-localization/source.wdl
claymcleod: I like keeping it in, provided we implement the retrying logic to keep us from flaking on tests.

claymcleod:
You could also do a `tokio::time::sleep` with a backoff that increments on each retry if you really wanted to be fancy 😄.

wdl-engine/src/eval/v1/task.rs
```diff
@@ -1117,7 +1117,7 @@ impl TaskEvaluator {
 
         // Localize the inputs
         self.backend
-            .localize_inputs(&self.downloader, &mut inputs)
+            .localize_inputs(Arc::new(self.downloader.clone()), &mut inputs)
```
claymcleod, line 1120: You actually aren't going to need this `Arc` I don't believe. The reason is that the `HttpDownloader` can be cheaply cloned, as it uses a bunch of `Arc`s internally. We should rearchitect this to not use the outer `Arc` if possible.

wdl-engine/src/http.rs
```diff
@@ -300,59 +321,104 @@ impl Downloader for HttpDownloader {
...
+                            break;
+                        }
+
+                        // backoff and retry
+                        let delay_secs =
+                            INITIAL_RETRY_DELAY.as_secs_f64() * 2.0f64.powi(attempt as i32);
+                        let delay = Duration::from_secs_f64(delay_secs);
+
+                        drop(permit);
+                        sleep(delay).await;
```
claymcleod, line 396: Let's log the sleep time.

wdl-engine/src/eval/v1/task.rs
dead8309: Replaced `dyn Downloader` with `HttpDownloader`  but I wanted to confirm if it's fine in case we need to extend the downloader later for mocks?

claymcleod: That's a good thought. For now, I think it's fine to keep it as is. If we do need to use mocks in the future, I'd probably reach for using generics rather than `&dyn Downloader` or `Arc<dyn Downloader>`.

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -397,41 +399,69 @@ impl TaskExecutionBackend for CrankshaftBackend {
...
-                inputs[index].set_guest_path(guest_path);
+                if let Some(input) = inputs.get_mut(index) {
+                    input.set_guest_path(guest_path);
+                } else {
+                    bail!("invalid index {} returned from trie", index);
+                }
             }
 
             // Localize all inputs
             // TODO: only do this for local task execution
```
claymcleod, line 428: I'm unsure what this comment means. @peterhuene, can this comment be deleted once this is merged in?

wdl-engine/src/backend/local.rs
```diff
@@ -269,39 +270,72 @@ impl TaskExecutionBackend for LocalTaskExecutionBackend {
...
+                        let input = inputs.get_mut(idx).expect("index should be valid");
+                        input.set_location(location);
+                        input.set_guest_path(guest_path);
+                    }
+                    Ok(Err(e)) => {
+                        // futures are aborted when the JoinSet is dropped
+                        bail!(e);
+                    }
+                    Err(e) => {
+                        bail!("download task panicked: {e}");
```
claymcleod, line 336:
```suggestion
                        bail!("download task failed: {e}");
```

wdl-engine/src/backend/local.rs
```diff
@@ -269,39 +270,72 @@ impl TaskExecutionBackend for LocalTaskExecutionBackend {
...
+                                )
+                            })?
+                            .to_string();
+
+                        let input = inputs.get_mut(idx).expect("index should be valid");
+                        input.set_location(location);
+                        input.set_guest_path(guest_path);
+                    }
+                    Ok(Err(e)) => {
+                        // futures are aborted when the JoinSet is dropped
```
claymcleod, line 332: Also, please make these comments the same as above.

wdl-engine/src/backend/local.rs
claymcleod, line 336: And add the comment here.

wdl-engine/src/http.rs
```diff
@@ -300,59 +321,111 @@ impl Downloader for HttpDownloader {
...
-                            );
-
-                            // Insert an entry and break out of the loop to download
+                            // not downloading, not downloaded. mark for retry/download
                             downloads.insert(
                                 url.clone().into_owned(),
-                                Status::Downloading(Arc::default()),
+                                Status::Downloading(Arc::new(Notify::new())),
                             );
-                            break;
```
claymcleod, line 331: Instead of this `Either`, you should just use an `Option`, as you use it exclusively as an option (logically speaking) below.

wdl-engine/src/http.rs
```diff
@@ -300,59 +321,111 @@ impl Downloader for HttpDownloader {
             // This loop exists so that all requests to download the same URL will block
             // waiting for a notification that the download has completed.
             // When the notification is received, the lookup into the downloads is retried
-            let mut retried = false;
+            let mut retried_wait = false;
```
claymcleod, line 324: This is probably just more aptly named `waited`.

wdl-engine/src/http.rs
```diff
@@ -300,59 +321,111 @@ impl Downloader for HttpDownloader {
...
+                    Ok(location) => {
+                        result = Ok(location);
+                        break; // permit is dropped here
+                    }
+                    Err(e) => {
+                        last_error = Some(Arc::new(e));
+                        // if it was the last attempt, return the error
+                        if attempt == MAX_DOWNLOAD_ATTEMPTS - 1 {
+                            result = Err(last_error.clone().unwrap_or_else(|| {
+                                Arc::new(anyhow!("unknown download error after max attempts"))
```
claymcleod, line 385: You can just `unwrap` with a `// SAFETY: ` comment here—you set `last_error` above, so it should never be `None` at this point.

wdl-engine/src/http.rs
```diff
@@ -300,59 +321,111 @@ impl Downloader for HttpDownloader {
...
+                        sleep(delay).await;
+                    }
+                }
+                // permit is implicitly dropped here
+            }
 
-                notify.notified().await;
-                retried = true;
-                continue;
+            if last_error.is_some() && result.is_ok() {
```
claymcleod, line 409: Won't this capture the case where we errored once or twice, but then on the last try the download succeeded? In that case, do we really want to return an error?

wdl-engine/src/http.rs
```diff
@@ -300,59 +321,111 @@ impl Downloader for HttpDownloader {
...
-            let res = self.get(&url).await.map_err(Arc::from);
             let notify = {
                 let mut downloads = self.downloads.lock().expect("failed to lock downloads");
-                match std::mem::replace(
-                    downloads.get_mut(&url).expect("should have status"),
-                    Status::Downloaded(res.clone()),
-                ) {
-                    Status::Downloading(notify) => notify,
-                    _ => panic!("file should be downloading"),
+                match downloads.insert(url.clone().into_owned(), Status::Downloaded(result.clone()))
```
claymcleod, line 417:
```suggestion
                match downloads.insert(url.into_owned(), Status::Downloaded(result.clone()))
```

wdl-engine/src/http.rs
dead8309: ah my bad i was being conservative here - removed thanks!

wdl-engine/src/http.rs
dead8309: revised to break out of loop with the error instead

wdl-engine/src/http.rs
```diff
@@ -41,6 +45,12 @@ mod s3;
 /// The default cache subdirectory that is appended to the system cache
 /// directory.
 const DEFAULT_CACHE_SUBDIR: &str = "wdl";
+
+/// Maximum number of download attempts
```
claymcleod, line 49:
```suggestion
/// Maximum number of download attempts.
```

wdl-engine/src/http.rs
```diff
@@ -141,13 +153,21 @@ impl HttpDownloader {
 
         let cache = Arc::new(Cache::new(DefaultCacheStorage::new(cache_dir)));
 
+        let max_downloads = config
+            .http
+            .max_concurrent_downloads
+            .unwrap_or(DEFAULT_MAX_CONCURRENT_DOWNLOADS) as usize;
+
+        info!("maximum concurrent downloads set to {max_downloads}");
```
claymcleod, line 161: I'm argue this is more `debug!()` information.

wdl-engine/src/http.rs
```diff
@@ -300,59 +320,107 @@ impl Downloader for HttpDownloader {
...
+                    }
+                    Err(e) => {
+                        let current_error = Arc::new(e.context(format!(
+                            "download attempt {} failed for `{url}`",
+                            attempt + 1
+                        )));
+
+                        // if it was the last attempt, return the error
+                        if attempt == MAX_DOWNLOAD_ATTEMPTS - 1 {
+                            info!(
```
claymcleod, line 380: This should probably be `error!()`.

claymcleod (commented):
After these changes, this looks right, and I'll merge it in. Thanks for sweating the details!

wdl-engine/src/backend/crankshaft.rs
claymcleod: I'm going to say, for now, let's keep it in.

claymcleod (approved):

PR #423, opened 2025-04-17
refactor: accept case insensitive `--except` args

dead8309:
> #@ except lint directives within WDL source should still require an exact case sensitive match, but I think we can relax that requirement on the CLI.

ref: https://github.com/stjude-rust-labs/sprocket/issues/95

Checking of rules were done inside wdl cli crate instead of sprocket.

I think the change should be inside this repository, but do let me know if i have to make changes in `sprocket` instead

Before submitting this PR, please make sure:

For external contributors:

- [x] You have read the [CONTRIBUTING](https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md) guide in its entirety.
- [x] You have not used AI on any parts of this pull request.

For all contributors:

- [x] You have added a few sentences describing the PR here.
- [x] Your code builds clean without any errors or warnings.
- [x] You have added tests (when appropriate).
- [x] You have updated the README or other documentation to account for these
      changes (when appropriate).
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.
- [x] Your PR title follows the [conventional commit] style.

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.1.0/

PR #430, opened 2025-04-21
feat: incremental command line parsing

claymcleod:
This commit started out as implementing incremental command line parsing. Incremental command line parsing is as follows:

- Parse both files and key-value pairs on the command line and combine the results into a single input object.
- Overwrite earlier values provided on the command line arguments with command line arguments that come later.

This commit then ballooned a bit, as it was more fitting for some of those pieces to be implemented over in the Sprocket code base (https://github.com/stjude-rust-labs/sprocket). This then snowballed into a larger discussion about moving functionality out of the `wdl` binary and into `sprocket` proper, which I went ahead and knocked out in this commit.

---

A few other notes:

- The idea of introducing `wdl-cli` is that it's a facade for common interactions with the `wdl-*` crates through an opinionated lens. Notice, however, that implementation specific behavior (progress bar styling, etc) has been moved out and will be included in `sprocket`. As such, `wdl-cli` still needs to be focused on a general purpose implementation that anyone could pick up and customize to their liking.
- This is one half of the set of changes needed to get the functionality into `sprocket`. I'll follow up with another PR in Sprocket.

---

For external contributors:

For all contributors:

      changes (when appropriate).


Cargo.toml
```diff
@@ -28,11 +29,11 @@ anyhow = "1.0.97"
 approx = "0.5.1"
 chrono = "0.4.40"
 clap = { version = "4.5.32", features = ["derive"] }
-clap-verbosity-flag = "3.0.2"
+clap-verbosity-flag = { version = "3.0.2", features = ["tracing"] }
 codespan-reporting = "0.11.1"
 colored = "3.0.0"
 convert_case = "0.8.0"
-crankshaft = "0.1.0"
+crankshaft = { path = "../crankshaft/crankshaft" }
```
claymcleod, line 36: Come back and replace this.

wdl-cli/src/analysis.rs
```diff
@@ -0,0 +1,175 @@
...
+mod source;
+
+pub use results::AnalysisResults;
+pub use source::Source;
+
+/// The type of the initialization callback.
+type InitCb = Box<dyn Fn() + 'static>;
+
+/// The type of the progress callback.
+type ProgressCb =
```
claymcleod, line 25: Get rid of `()` here.

wdl-cli/src/analysis.rs
```diff
@@ -0,0 +1,175 @@
...
+
+impl Default for Analysis {
+    fn default() -> Self {
+        Self {
+            sources: Default::default(),
+            exceptions: Default::default(),
+            lint: Default::default(),
+            shellcheck: Default::default(),
+            init: Box::new(|| {}),
+            progress: Box::new(|_, _, _, _| Box::pin(async {})),
```
claymcleod, line 143: Should these be options with a default for progress?

wdl-cli/src/analysis.rs
```diff
@@ -0,0 +1,175 @@
...
+            .filter(|rule| !exceptions.contains(rule.id())),
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::Analysis;
+
+    #[test]
+    fn default() {
```
claymcleod, line 172: This isn't needed anymore now that the default impl doesn't unwrap anything.

wdl-cli/src/eval.rs
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /// Runs a WDL task or workflow evaluation.
+    pub async fn run<R, P>(
+        mut self,
+        token: CancellationToken,
+        progress: P,
+    ) -> Result<EvaluationResult>
+    where
+        R: Future<Output = ()> + Send,
```
claymcleod, line 80: Can we simplify this by using `BoxFuture`?

wdl-cli/src/eval.rs
```diff
@@ -0,0 +1,150 @@
...
+        P: Fn(ProgressKind<'_>) -> R + Send + Sync + 'static,
+    {
+        let result = match self.inputs {
+            Inputs::Task(ref mut inputs) => {
+                // Make any paths specified in the inputs absolute.
```
claymcleod, line 85: This comment doesn't fit.

wdl-cli/src/eval.rs
```diff
@@ -0,0 +1,150 @@
...
+            }
+        };
+
+        match result {
+            Ok(outputs) => Ok(EvaluationResult::Success(outputs)),
+            Err(e) => match e {
+                EvaluationError::Source(diagnostic) => {
+                    let source = self.document.root().text().to_string();
+                    emit_diagnostics(&self.document.path(), &source, &[diagnostic])?;
+                    Ok(EvaluationResult::Failed)
```
claymcleod, line 144: I wonder if this is really needed? Can we get away with just returning the error and eliminating `EvaluationResult`?

wdl-cli/src/inputs.rs
```diff
@@ -0,0 +1,492 @@
...
+use wdl_engine::PrimitiveValue;
+use wdl_engine::Value;
+
+pub mod file;
+pub mod origin_paths;
+
+pub use file::InputFile;
+pub use origin_paths::OriginPaths;
+
+/// A regex that matches a valid identifier.
```
claymcleod, line 25: "This is useful when..."

wdl-cli/src/inputs.rs
```diff
@@ -0,0 +1,492 @@
...
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// An input parsed from the command line.
+#[derive(Clone, Debug, PartialEq)]
+pub enum Input {
+    /// A file.
+    File(
+        /// The path to the file.
+        ///
+        /// If this input is successfully created, the input is always
```
claymcleod, line 76:
```suggestion
        /// If this input is successfully created, the input is
```

wdl-cli/src/inputs/file.rs
```diff
@@ -0,0 +1,111 @@
...
+use wdl_engine::Object;
+use wdl_engine::Value;
+
+use crate::Inputs;
+
+/// An error related to a input file.
+#[derive(Error, Debug)]
+pub enum Error {
+    /// An input file cannot be read from a directory.
+    #[error("an input file cannot be read from the directory: `{0}`")]
```
claymcleod, line 17:
```suggestion
    #[error("an input file cannot be read from a directory: `{0}`")]
```

wdl-cli/src/inputs/file.rs
```diff
@@ -0,0 +1,111 @@
...
+    InvalidDir(PathBuf),
+
+    /// An I/O error.
+    #[error(transparent)]
+    Io(std::io::Error),
+
+    /// The input file did not contain a map at the root.
+    #[error("input file `{0}` did not contain a map at the root")]
```
claymcleod, line 25:
```suggestion
    #[error("input file `{0}` did not contain a map from strings to values at the root")]
```

wdl-cli/src/inputs/file.rs
```diff
@@ -0,0 +1,111 @@
...
+    NonMapRoot(PathBuf),
+
+    /// Neither JSON nor YAML could be parsed from the provided path.
+    #[error("file `{0}` is neither a valid JSON nor a valid YAML file")]
+    UnsupportedFormat(PathBuf),
```
claymcleod, line 30: Make this more general.

wdl-cli/src/inputs/file.rs
```diff
@@ -0,0 +1,111 @@
...
+}
+
+/// A [`Result`](std::result::Result) with an [`Error`].
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// An input file for a WDL run.
```
claymcleod, line 36:
```suggestion
/// An input file containing WDL values.
```

wdl-cli/src/inputs/file.rs
```diff
@@ -0,0 +1,111 @@
...
+pub struct InputFile;
+
+impl InputFile {
+    /// Reads a input file.
```
claymcleod, line 40:
```suggestion
    /// Reads an input file.
```

wdl-engine/src/backend/crankshaft.rs
```diff
@@ -542,6 +542,7 @@ impl TaskExecutionBackend for CrankshaftBackend {
             let inner_backend = self.inner.clone();
             let generator = self.generator.clone();
             let output_path = std::path::absolute(output_dir).expect("failed to get absolute path");
+
```
claymcleod, line 545: Remove.

wdl-engine/src/inputs.rs
```diff
@@ -656,22 +664,23 @@ impl Inputs {
 
         // Parse the JSON (should be an object)
         let reader = BufReader::new(file);
-        let object = mem::take(
```
claymcleod, line 659: All of this ended up being unnecessary—we can just deserialize directly.

wdl-engine/src/inputs.rs
```diff
@@ -694,28 +703,22 @@ impl Inputs {
 
         // Parse the YAML
         let reader = BufReader::new(file);
-        let yaml: YamlValue = serde_yaml_ng::from_reader(reader).with_context(|| {
```
claymcleod, line 697: Same here.

wdl-engine/src/outputs.rs
```diff
@@ -9,7 +9,7 @@ use crate::Scope;
 use crate::Value;
 
 /// Represents outputs of a WDL workflow or task.
-#[derive(Default, Debug)]
+#[derive(Default, Debug, PartialEq)]
```
claymcleod, line 12: All these are just to enable testing.

wdl-engine/src/value.rs
```diff
@@ -1155,6 +1178,21 @@ impl PrimitiveValue {
         }
     }
 
+    /// Performs common substitutions on file and directory paths.
+    pub(crate) fn substitute_path(&mut self) -> Result<()> {
```
claymcleod, line 1182: Let's rename this.

wdl-cli/src/analysis.rs
claymcleod: Nevermind, this needs to stay in.

wdl-cli/src/analysis.rs
claymcleod: After looking at this closer, I don't think so. I think it's fine.

wdl-cli/src/eval.rs
claymcleod: On second thought, nah.

wdl-cli/src/inputs/file.rs
claymcleod: On second thought, nevermind. I think it's not likely we will support other input file types any time soon.

Cargo.toml
```diff
@@ -28,11 +29,11 @@ anyhow = "1.0.98"
 approx = "0.5.1"
 chrono = "0.4.40"
 clap = { version = "4.5.37", features = ["derive"] }
-clap-verbosity-flag = "3.0.2"
+clap-verbosity-flag = { version = "3.0.2", features = ["tracing"] }
 codespan-reporting = "0.12.0"
 colored = "3.0.0"
 convert_case = "0.8.0"
-crankshaft = "0.1.0"
+crankshaft = { git = "https://github.com/stjude-rust-labs/crankshaft", rev = "4975725f4" }
```
claymcleod, line 36: We need a crankshaft release probably.

wdl-engine/tests/inputs/missing-file/error.txt
```diff
@@ -1,4 +1,5 @@
-failed to open input file `tests/inputs/missing-file/inputs.json`
+failed to parse input file `tests/inputs/missing-file/inputs.json`
```
claymcleod, line 1: This error changed because I moved a `context` call around to accommodate identifier key verification earlier in the process. I felt it was a fine side effect, even though it's a bit redundant.

wdl-engine/src/inputs.rs
claymcleod: Funnily enough, this turned out _not_ to be unnecessary :). I reverted these changes back.

wdl-engine/src/inputs.rs
claymcleod: And, again, same here.

wdl-cli/src/analysis.rs
```diff
@@ -0,0 +1,164 @@
...
+/// The type of the progress callback.
+type ProgressCb =
+    Box<dyn Fn((), ProgressKind, usize, usize) -> BoxFuture<'static, ()> + Send + 'static>;
+
+/// An analysis.
+pub struct Analysis {
+    /// The set of root nodes to analyze.
+    ///
+    /// Can be files, directories, or URLs.
+    sources: Vec<Source>,
```
a-frantz, line 33: Small nit: I find that if there's the ability to pass in directories, I don't ever have the need to pass in multiple "sources". I'd be ok replacing this `Vec` with a single `source` for the sake of simplifying things, but this doesn't add much complexity so ok to ignore this comment. Or maybe there's a use case I'm not thinking of where multiple sources is useful 🤷‍♀️

claymcleod: I think it makes sense generally to provide the ability to add multiple entrypoints to an analysis—even for use cases we're not thinking of right now (i.e., downstream users). I would say, off the top of my head, just adding two specific files (e.g., `sprocket check file_one.wdl file_two.wdl`) without globbing an entire directory is one such use case. Further, remember that most shells are going to expand globs before passing the arguments to the command line tool unless you surround it in quotes.

wdl-cli/src/inputs.rs
```diff
@@ -0,0 +1,507 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn identifier_regex() {
+        assert!(IDENTIFIER_REGEX.is_match("here_is-an.identifier"));
```
a-frantz, line 341: Are dashes allowed in WDL identifiers? I don't think they are, but I could be wrong.

wdl-cli/src/analysis/source.rs
```diff
@@ -0,0 +1,146 @@
...
+            Source::Directory(path) => write!(f, "{path}", path = path.display()),
+        }
+    }
+}
+
+impl std::str::FromStr for Source {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Some(url) = parse_url(s) {
```
claymcleod, line 62: This is deserving of being called out. @peterhuene, since there is no great way to tell if a string is a remote URL or a local file/directory, I've relied on your function in `wdl-engine` to check that. That's the best solution I can think of without purely duplicating the function, and at least it means that `wdl-engine` and `wdl-cli` will be in sync. That being said, it does require the `path` module of `wdl-engine` to be made public.

a-frantz:
There's a behavior regression here from PR #423

We should have case insensitive matching on rules passed by `--except`

wdl-cli/src/analysis/results.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl AnalysisResults {
+    /// Attempts to create a new set of analysis results.
+    ///
+    /// Returns an error if an error was encountered during analysis. That being
+    /// said, each analysis result may have diagnostics.
+    pub fn try_new(results: Vec<AnalysisResult>) -> Result<Self> {
+        for result in results.iter() {
+            if let Some(e) = result.error() {
+                bail!(e.to_owned());
```
a-frantz, line 29: I'm not sure this is the best behavior. We've done a full analysis at this point, so it's not like bailing on the _first_ error saves any compute. I think it would be better to return the full set of errors. We can leave it up to the caller whether the full set of errors is reported or just the first one.

wdl-cli/src/inputs.rs
```diff
@@ -0,0 +1,507 @@
...
+
+        if parts.len() != 2 {
+            return Err(Error::InvalidPair {
+                pair: s.to_string(),
+                reason: format!(
+                    "expected exactly one equal sign (`=`), found {}",
+                    parts.len() - 1,
+                ),
+            });
+        }
```
a-frantz, line 192: What about `sprocket run ... weird_param='Bash_hack=foo'`? Where `weird_param` is of type `String`?

wdl-cli/src/inputs.rs
```diff
@@ -0,0 +1,507 @@
...
+        check_string_value(&inputs, "foo", "bar");
+        check_boolean_value(&inputs, "baz", false);
+        check_string_value(&inputs, "quux", "jacks");
+        check_string_value(&inputs, "new", "foobarbaz");
+        check_string_value(&inputs, "new_two", "bazbarfoo");
+        check_integer_value(&inputs, "sandwich", -100);
+
+        // An invalid key-value pair.
+        let error =
+            Inputs::coalesce(["./tests/fixtures/inputs_one.json", "foo=baz#bar"]).unwrap_err();
```
a-frantz, line 489: would this work `baz#bar` were wrapped in quotes?

wdl-cli/src/inputs/origin_paths.rs
```diff
@@ -0,0 +1,50 @@
+//! The set of origin paths for each read input argument.
+//!
+//! Origin paths as the associated directory for which each parsed command line
```
a-frantz, line 3: typo here?

wdl-engine/src/inputs.rs
```diff
@@ -614,15 +624,16 @@ impl Inputs {
     /// Parses an inputs file from the given file path.
     ///
     /// The format (JSON or YAML) is determined by the file extension:
+    ///
     /// - `.json` for JSON format
     /// - `.yml` or `.yaml` for YAML format
     ///
     /// The parse uses the provided document to validate the input keys within
     /// the file.
```
a-frantz, line 632: The `wdl-cli` implementation doesn't care about the extension. Should this be updated to use the same logic? (Attempt to parse as both, ignoring the extension)

peterhuene (commented):
Looks like a promising change. I have a few comments.

wdl-cli/src/analysis/results.rs
claymcleod: FWIW, this is just copied from the current `wdl` binary does (https://github.com/stjude-rust-labs/wdl/blob/0682dca08692717361b23e864276feb562d1a57e/wdl/src/bin/wdl.rs#L108-L110 and https://github.com/stjude-rust-labs/wdl/blob/0682dca08692717361b23e864276feb562d1a57e/wdl/src/bin/wdl.rs#L357-L359). That being said, since I'm formalizing it, I don't mind making it a new return type.

wdl-engine/src/value.rs
```diff
@@ -59,7 +59,7 @@ pub trait Coercible: Sized {
 /// Represents a WDL runtime value.
 ///
 /// Values are cheap to clone.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
```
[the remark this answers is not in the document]
claymcleod: Okay, I wasn't aware of that. That makes sense, and I've removed it in my latest revision.

wdl-engine/src/inputs.rs
```diff
@@ -23,16 +24,14 @@ use wdl_analysis::types::v1::task_hint_types;
...
 
-/// A type alias to a JSON map (object).
-type JsonMap = serde_json::Map<String, JsonValue>;
-
 /// Helper for replacing input paths with a path derived from joining the
 /// specified path with the input path.
 fn join_paths(
     inputs: &mut HashMap<String, Value>,
-    path: &Path,
+    path: impl Fn(&str) -> Result<PathBuf>,
```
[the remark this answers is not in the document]
claymcleod: Good catch.

wdl-engine/src/inputs.rs
claymcleod: I agree with Peter, so I'll keep it as is.

wdl-cli/src/inputs.rs
claymcleod: Yes, it would.

wdl-cli/src/analysis/source.rs
```diff
@@ -0,0 +1,146 @@
...
+            return Ok(Self::Remote(url));
+        }
+
+        let path = shellexpand::full(s)
```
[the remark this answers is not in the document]
claymcleod: You're right, we don't need `shellexpand` here.

To be clear, the reason that we do need `shellexpand` in the dependencies is that, with our syntax of `key=value` on the command line, some shell symbols won't be expanded. For example,

```
$ echo ~/Desktop/
/Users/cmcleod/Desktop/
$ echo foo=~/Desktop
foo=~/Desktop
```

But the use of it can be constrained to just when paths are expanded.

wdl-engine/src/inputs.rs
```diff
@@ -656,32 +668,37 @@ impl Inputs {
...
+        .map(|(key, value)| {
+            let value: Value = serde_json::from_value(value)
+                .with_context(|| format!("invalid input key `{key}`"))?;
+            anyhow::Ok::<(String, Value)>((key, value))
+        })
+        .collect::<Result<IndexMap<_, _>, _>>()
+        .map(Object::from)?;
 
         Self::parse_object(document, object)
-            .with_context(|| format!("failed to parse input file `{path}`", path = path.display()))
```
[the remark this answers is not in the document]
claymcleod: This is wrapped up in this change: https://github.com/stjude-rust-labs/wdl/pull/430#discussion_r2059303228

wdl-engine/src/inputs.rs
```diff
@@ -656,32 +668,37 @@ impl Inputs {
...
         // Parse the JSON (should be an object)
         let reader = BufReader::new(file);
-        let object = mem::take(
-            serde_json::from_reader::<_, JsonValue>(reader)
-                .with_context(|| {
-                    format!("failed to parse input file `{path}`", path = path.display())
-                })?
+
+        let object = std::mem::take(
+            serde_json::from_reader::<_, JsonValue>(reader)?
```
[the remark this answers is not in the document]
claymcleod: This was moved higher up in the chain here: https://github.com/stjude-rust-labs/wdl/pull/430/files#diff-64962bf152af0751e85bb87ad487b49d50bad1457a39a9d589725ec641ffcf00R648. You can be sure it was faithfully moved, as all of the tests that rely on it still pass.

wdl-engine/src/inputs.rs
```diff
@@ -694,9 +711,7 @@ impl Inputs {
 
         // Parse the YAML
         let reader = BufReader::new(file);
-        let yaml: YamlValue = serde_yaml_ng::from_reader(reader).with_context(|| {
-            format!("failed to parse input file `{path}`", path = path.display())
-        })?;
+        let yaml: YamlValue = serde_yaml_ng::from_reader(reader)?;
```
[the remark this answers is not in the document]
claymcleod: Same response here for me as well.

wdl-cli/src/analysis/source.rs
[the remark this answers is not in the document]
peterhuene: Gotcha, okay that makes sense to me then.

wdl-engine/src/inputs.rs
```diff
@@ -707,15 +722,22 @@ impl Inputs {
...
+        .map(|(key, value)| {
+            let value: Value = serde_json::from_value(value)
+                .with_context(|| format!("invalid input key `{key}`"))?;
+            anyhow::Ok::<(String, Value)>((key, value))
+        })
+        .collect::<Result<IndexMap<_, _>, _>>()
+        .map(Object::from)?;
 
         Self::parse_object(document, object)
-            .with_context(|| format!("failed to parse input file `{path}`", path = path.display()))
```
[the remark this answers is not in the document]
claymcleod: Same response here.

wdl-cli/src/inputs/file.rs
```diff
@@ -0,0 +1,111 @@
...
+        let path = path.as_ref();
+
+        if path.is_dir() {
+            return Err(Error::InvalidDir(path.to_path_buf()));
+        }
+
+        // SAFETY: the check above ensures that the path is not a directory,
+        // which means that it can't be the root directory, which means that
+        // this call to `.parent()` cannot return `None`.
+        let parent = path.parent().unwrap();
```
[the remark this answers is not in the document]
claymcleod: The code was not that informative on what would happen, nor were the Windows developer docs. I'll just have to check it on a Windows machine here in a bit.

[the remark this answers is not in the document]
claymcleod: Appears to work as expected:

```rust
fn main() {
    let arg = std::env::args().nth(1).unwrap();
    let path = std::path::Path::new(&arg);

    println!("{path}", path = path.display());
    println!("  -> exists: {}", path.exists());
    println!("  -> is file: {}", path.is_file());
    println!("  -> is directory: {}", path.is_dir());
}
```

```sh
$ cargo run --release 'C:\'
```

```txt
C:\
  -> exists: true
  -> is file: false
  -> is directory: true
```

wdl-cli/src/analysis.rs
```diff
@@ -0,0 +1,202 @@
...
+    let mut unknown = exceptions
+        .iter()
+        .filter(|rule| !names.iter().any(|name| name.eq_ignore_ascii_case(rule)))
+        .map(|rule| format!("`{rule}`"))
+        .collect::<Vec<_>>();
+
+    if !unknown.is_empty() {
+        unknown.sort();
+
+        warn!(
```
a-frantz, line 177: This seems fine for `wdl-cli`, but in the sprocket bin we probably want to use `find_nearest_rule()` ([here](https://github.com/stjude-rust-labs/wdl/blob/main/wdl-lint/src/util.rs#L108)) to offer a suggestion of what they might have meant to type.
Just something to keep in mind while designing this API

wdl-cli/src/analysis/results.rs
```diff
@@ -0,0 +1,90 @@
...
+///
+/// If successfully created, the set of analysis results are guaranteed not to
+/// have any associated errors (but they may contain diagnostics).
+#[derive(Debug)]
+pub struct AnalysisResults(Vec<AnalysisResult>);
+
+impl AnalysisResults {
+    /// Attempts to create a new set of analysis results.
+    ///
+    /// Returns an error if an error was encountered during analysis. That being
```
a-frantz, line 24:
```suggestion
    /// Returns any errors encountered during analysis. That being
```

wdl-cli/src/analysis/source.rs
```diff
@@ -0,0 +1,142 @@
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if let Some(url) = parse_url(s) {
+            return Ok(Self::Remote(url));
+        }
+
+        let path = Path::new(s);
+
+        let path = std::path::absolute(path)
+            .map_err(|_| anyhow!("failed to convert `{path}` to a URI", path = path.display()))
```
a-frantz, line 69:
```suggestion
            .map_err(|_| anyhow!("failed to make `{path}` absolute", path = path.display()))
```

wdl-cli/src/inputs.rs
```diff
@@ -0,0 +1,499 @@
...
+        let err = r#"foo$="bar""#.parse::<Input>().unwrap_err();
+        assert!(matches!(
+                err,
+                Error::InvalidPair {
+                    pair,
+                    reason
+                } if pair == r#"foo$="bar""# &&
+                reason == r"key `foo$` did not match the identifier regex (`^([a-zA-Z][a-zA-Z0-9_.]*)$`)"));
+
+        // An value that is valid despite that value not being valid as a key.
```
a-frantz, line 393:
```suggestion
        // A value that is valid despite that value not being valid as a key.
```

wdl-cli/src/inputs/file.rs
```diff
@@ -0,0 +1,122 @@
...
+/// A [`Result`](std::result::Result) with an [`Error`].
+pub type Result<T> = std::result::Result<T, Error>;
+
+/// An input file containing WDL values.
+pub struct InputFile;
+
+impl InputFile {
+    /// Reads an input file.
+    ///
+    /// The file is attempted to be parsed as JSON and, if that fails, then
```
a-frantz, line 53: this doc string needs updating to reflect the extension logic

wdl-engine/src/value.rs
```diff
@@ -1155,6 +1170,21 @@ impl PrimitiveValue {
         }
     }
 
+    /// Performs expansions for file and directory paths.
+    pub(crate) fn expand_path(&mut self) -> Result<()> {
+        let path = match self {
+            PrimitiveValue::File(path) => path,
+            PrimitiveValue::Directory(path) => path,
+            _ => unreachable!("only file and directory values can be substituted"),
```
a-frantz, line 1178:
```suggestion
            _ => unreachable!("only file and directory values can be expanded"),
```

wdl-cli/src/analysis.rs
claymcleod: Okay, I will incorporate that into my Sprocket PR when I get around to working on that one again.

wdl-cli/src/analysis/source.rs
claymcleod: I actually did consider this one, and I think I like the more generic `failed to cover to a URI`. The reason for this is because, when a path is not able to be made absolute, there's probably something funkier going on (e.g., it's not a real path) in which case the error message is more concerning if it's that specific.

wdl-cli/src/inputs.rs
```diff
@@ -0,0 +1,499 @@
...
+            if !path.exists() {
+                return Err(Error::FileNotFound(path));
+            }
+
+            return Ok(Input::File(path));
+        }
+
+        // SAFETY: we just checked above to make sure the string contains an
+        // equals sign.
+        let (key, value) = s.split_once("=").unwrap();
```
peterhuene, line 184: Nit: could we just do the following to avoid the `contains` followed by `unwrap`:

```rust
match s.split_once("=") {
     Some((key, value)) => {
        // ...
     }
     None => {
        // Do the `Input::File` branch above
}
```

wdl-cli/src/inputs/file.rs
```diff
@@ -0,0 +1,126 @@
...
+    #[error(transparent)]
+    Io(std::io::Error),
+
+    /// The input file did not contain a map at the root.
+    #[error("input file `{0}` did not contain a map from strings to values at the root")]
+    NonMapRoot(PathBuf),
+
+    /// Neither JSON nor YAML could be parsed from the provided path.
+    #[error(
+        "unsupported file extension: `{0}`; the supported formats are JSON (`.json`) or YAML \
```
peterhuene, line 34:
```suggestion
        "unsupported file extension `{0}`: the supported formats are JSON (`.json`) or YAML \
```

wdl-cli/src/inputs/file.rs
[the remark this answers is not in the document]
peterhuene: To be clear, I meant `C:` and not `C:\`. The former is a path consisting of _just_ the drive prefix and the latter is the prefix *with* the root path, so it is the root directory of that drive.

peterhuene (commented):
Just a few more minor comments.

Looks like we need a formatting and clippy update for CI, but other than that I think this is ready.

peterhuene (approved):

a-frantz (approved):

PR #432, opened 2025-04-22
feat: improve diagnostic for task execution failure.

peterhuene:
The task execution failure message could be overly verbose and lacked links to the source WDL.

This commit changes the error returned from evaluation to include a "callstack" which is simply the locations of call statements; this can be used when reporting errors to properly display a WDL evaluation callstack.

Additionally, the stderr output from the diagnostic was always printing _all_ of the stderr contents; this now effectively tails the last 10 lines of the file in case the stderr stream was verbose.

Dependencies were also updated to latest.

For all contributors:

      changes (when appropriate).


wdl-engine/tests/tasks.rs
```diff
@@ -421,21 +415,40 @@ fn compare_evaluation_results(
...
-        &diagnostic.to_codespan(),
-    )
-    .expect("should emit");
-
-    String::from_utf8(buffer.into_inner()).expect("should be UTF-8")
+/// Creates a string from the given evaluation error.
+fn eval_error_to_string(e: &EvaluationError) -> String {
+    match e {
+        EvaluationError::Source(e) => {
+            let mut files = SimpleFiles::new();
```
claymcleod, line 422: Discussion point, but not necessarily for this PR: should we be keeping a `SimpleFiles` associated with the analysis itself rather than creating it on demand?

peterhuene: There's no dependency on `codespan_reporting` from `wdl-analysis` and I don't think it's necessary to introduce one here; given that we only construct a `SimpleFiles` once upon evaluation error and then immediately exit.

Generally the diagnostic representation from `wdl-grammar` and down into `wdl-engine` has been agnostic with regards to _how_ a diagnostic will be displayed to the user. I'd prefer the `codespan_reporting` dependency to be in the executable projects and not the libs (with the exception of a helper in `wdl-grammar` for converting to a `codespan_reporting::Diagnostic`, which is optional and enabled from the executables).

claymcleod (approved):

wdl-engine/tests/tasks.rs
claymcleod: Sure that makes sense.

PR #445, opened 2025-05-15
revise: clarity around "allow nested inputs"

a-frantz:

If a WDL >=v1.2 workflow has a non-boolean value for the ,"allow nested inputs" hint, the metadata will also be searched for `allowNestedInputs: <bool>` instead of prematurely dis-allowing nested inputs

For external contributors:

For all contributors:

      changes (when appropriate).

### DIFF
--- a/wdl-ast/CHANGELOG.md
+++ b/wdl-ast/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Changed
+
+* Values of `WORKFLOW_HINT_ALLOW_NESTED_INPUTS` and `WORKFLOW_HINT_ALLOW_NESTED_INPUTS_ALIAS` were swapped
+    and documentation updated to clarify behavior differences between WDL v1.1 and >v1.2
+    ([#445](https://github.com/stjude-rust-labs/wdl/pull/445)).
+
 ## 0.12.1 - 05-02-2025
 
 _A patch bump was required because an error was made during the release of `wdl` v0.13.0 regarding dependencies._
--- a/wdl-ast/CHANGELOG.md
+++ b/wdl-ast/CHANGELOG.md
@@ -7,12 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
-#### Changed
-
-* Values of `WORKFLOW_HINT_ALLOW_NESTED_INPUTS` and `WORKFLOW_HINT_ALLOW_NESTED_INPUTS_ALIAS` were swapped
-    and documentation updated to clarify behavior differences between WDL v1.1 and >=v1.2
-    ([#445](https://github.com/stjude-rust-labs/wdl/pull/445)).
-
 #### Fixed
 
 * If a WDL >=v1.2 workflow has a non-boolean value for the `allow_nested_inputs` hint, the metadata will also be
--- a/wdl-ast/CHANGELOG.md
+++ b/wdl-ast/CHANGELOG.md
@@ -10,7 +10,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 #### Changed
 
 * Values of `WORKFLOW_HINT_ALLOW_NESTED_INPUTS` and `WORKFLOW_HINT_ALLOW_NESTED_INPUTS_ALIAS` were swapped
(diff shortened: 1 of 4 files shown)


wdl-ast/src/v1/workflow.rs
```diff
@@ -22,12 +22,13 @@ use crate::SyntaxKind;
 use crate::SyntaxNode;
 use crate::TreeNode;
 
-/// The name of the `allow_nested_inputs` workflow hint.
-pub const WORKFLOW_HINT_ALLOW_NESTED_INPUTS: &str = "allow_nested_inputs";
+/// The name of the `allow_nested_inputs` workflow metadata hint (i.e.
+/// `allowNestedInputs`) for WDL v1.1 and later.
+pub const WORKFLOW_HINT_ALLOW_NESTED_INPUTS: &str = "allowNestedInputs";
```
adthrasher, line 27: https://github.com/openwdl/wdl/blob/wdl-1.2/SPEC.md#allow_nested_inputs
Based on my reading, this was correct and now it is reversed.

> allow_nested_inputs
Allowed type: Boolean
Alias: allowNestedInputs

a-frantz: The v1.2 spec is a bit revisionist... In WDL v1.1 `allow_nested_inputs` isn't accepted anywhere, and you must use the camelCase form. So while the 1.2 spec swaps them, it would be weird to only accept an "alias" for 1.1 and reject the "non-alias" form

adthrasher: I see. I still think it'll be a long-term complication that we have the `WORKFLOW_HINT_ALLOW_NESTED_INPUTS` and `WORKFLOW_HINT_ALLOW_NESTED_INPUTS_ALIAS` swapped compared to the spec going forward. I'd suggest we simply add a comment below where the alias would be used that states something like: pre-1.2 this was the only valid option, but now it is considered the alias.

peterhuene: It was made this way intentionally to follow the 1.2 spec and going forward.

`allowNestedInputs` is now the alias and not the preferred name; it may be deprecated in a future version of the spec.

wdl-ast/src/v1/workflow.rs
```diff
@@ -117,10 +119,10 @@ impl<N: TreeNode> WorkflowDefinition<N> {
         self.metadata()
             .and_then(|s| {
                 s.items().find_map(|i| {
-                    if i.name().text() == "allowNestedInputs" {
+                    if i.name().text() == WORKFLOW_HINT_ALLOW_NESTED_INPUTS_ALIAS {
```
a-frantz, line 122: @peterhuene I just skimmed the spec and couldn't find an answer. In WDL v1.2, are "hints" allowed to be in the meta section at all? Currently, if no form of the allow nested inputs hint is found _in the `hints` section_, it falls through and searches _only_ for the `allowNestedInputs` form. Either both forms should be usable in meta sections or neither. The current implementation is odd.

peterhuene: It's entirely a backwards compatibility fallback and as `allow_nested_inputs` was never respected in the meta in 1.1, we only look for `allowNestedInputs` there.

Plus, `allowNestedInputs` in the meta section is now deprecated by 1.2 and its use as a fallback will be removed in the future.

Where possible, we should be encouraging `allow_nested_inputs` in a `hints` section from 1.2+.

wdl-ast/CHANGELOG.md
```diff
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Fixed
+
+* If a WDL >=v1.2 workflow has a non-boolean value for the `allow_nested_inputs` hint, the metadata will also be
+    searched for `allow_nested_inputs: <bool>` instead of prematurely dis-allowing nested inputs
```
a-frantz, line 13:
```suggestion
* If a WDL >=v1.2 workflow has a non-boolean value for the "allow nested inputs" hint, the metadata will also be
    searched for `allowNestedInputs: <bool>` instead of prematurely dis-allowing nested inputs
```

a-frantz:
@peterhuene @adthrasher reverted most of the changes here. This PR is really just a doc string clarification now, and a small change in corner case behavior (which is a case not covered by the SPEC, so I think we would be fine to do whatever we want really... This just documents the behavior a bit better 🤷‍♀️ )

peterhuene (approved):

PR #454, opened 2025-05-21
feat: implement a TES task execution backend.

peterhuene:
This PR implements a task execution backend that uses the Crankshaft TES backend for sending tasks to a TES server.

The TES backend currently has these limitations:

* The backend fails a task execution if it has an input that is a non-empty directory; this may occur with explicit use of `Directory` type inputs, but also if files are written to a task's `tmp` directory (i.e. from a call to any of the `write_` stdlib functions). In the future, the backend will upload the directory contents to configured cloud storage and specify the input by URL.

* Local input file contents are sent to the TES server directly when creating the task. For large files, this may easily run up against limitations in input and request sizes. In the future, the backend will upload the file to cloud storage and specify the input by URL.

* The parallelism of the TES backend is currently hardcoded to 100 tasks. TES itself has no facility to discover the server's concurrency limitations and due to its polling nature, a large number of tasks may be very inefficient to monitor effectively. We'll revisit this going forward and perhaps implement a better monitoring solution for effective parallelism.

* The TES backend only supports inputs and outputs from cloud storage URLs; it will not yet pass local paths to the TES server, meaning it inherently doesn't support a TES server using a shared file system. In the future, it should be possible to configure the TES backend to use shared file system paths and the backend would then skip uploading inputs to cloud storage.

This new backend was exclusively tested with the now-defunct TES on Azure implementation from Microsoft. However, given the recent developments regarding that project, it's probably not a good idea to implement integration tests around it. Therefore, this isn't a well-tested backend at the moment. Once we have a replacement server implementation to use, I plan on implementing some integration tests for this backend.

For external contributors:

For all contributors:

      changes (when appropriate).

### DIFF
--- a/wdl-engine/src/backend/local.rs
+++ b/wdl-engine/src/backend/local.rs
@@ -131,7 +131,7 @@ impl TaskManagerRequest for LocalTaskRequest {
             .kill_on_drop(true);
 
         // Set the PATH variable for the child on Windows to get consistent PATH
-        // searching See: https://github.com/rust-lang/rust/issues/122660
+        // searching. See: https://github.com/rust-lang/rust/issues/122660
         #[cfg(windows)]
         if let Ok(path) = std::env::var("PATH") {
             command.env("PATH", path);
--- a/wdl-engine/src/backend/local.rs
+++ b/wdl-engine/src/backend/local.rs
@@ -130,10 +130,12 @@ impl TaskManagerRequest for LocalTaskRequest {
             )
             .kill_on_drop(true);
 
-        // Set an environment variable on Windows to get consistent PATH searching
-        // See: https://github.com/rust-lang/rust/issues/122660
+        // Set the PATH variable for the child on Windows to get consistent PATH
+        // searching See: https://github.com/rust-lang/rust/issues/122660
         #[cfg(windows)]
-        command.env("WDL_TASK_EVALUATION", "1");
+        if let Ok(path) = std::env::var("PATH") {
+            command.env("PATH", path);
+        }
 
         let mut child = command.spawn().context("failed to spawn `bash`")?;
 
--- a/wdl-engine/src/backend/tes.rs
+++ b/wdl-engine/src/backend/tes.rs
@@ -1,0 +1,537 @@
+//! Implementation of the TES backend.
+
+use std::collections::HashMap;
+use std::fs;
+use std::path::Path;
+use std::sync::Arc;
+use std::sync::Mutex;
+
+use anyhow::Context;
+use anyhow::Result;
+use anyhow::anyhow;
+use anyhow::bail;
+use base64::Engine as _;
+use base64::engine::general_purpose::STANDARD;
+use crankshaft::config::backend;
+use crankshaft::engine::Task;
+use crankshaft::engine::service::name::GeneratorIterator;
+use crankshaft::engine::service::name::UniqueAlphanumeric;
+use crankshaft::engine::service::runner::Backend;
+use crankshaft::engine::service::runner::backend::TaskRunError;
+use crankshaft::engine::service::runner::backend::tes;
+use crankshaft::engine::task::Execution;
+use crankshaft::engine::task::Input;
+use crankshaft::engine::task::Output;
+use crankshaft::engine::task::Resources;
+use crankshaft::engine::task::input::Contents;
+use crankshaft::engine::task::input::Type as InputType;
+use crankshaft::engine::task::output::Type as OutputType;
+use futures::FutureExt;
+use futures::future::BoxFuture;
+use nonempty::NonEmpty;
+use tokio::sync::oneshot;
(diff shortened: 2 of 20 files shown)


claymcleod:
> The parallelism of the TES backend is currently hardcoded to 100 tasks. TES itself has no facility to discover the server's concurrency limitations and due to its polling nature, a large number of tasks may be very inefficient to monitor effectively. We'll revisit this going forward and perhaps implement a better monitoring solution for effective parallelism.

Since we now have configuration, how easy would it be to just pass this through as a configuration object when we finally get to exposing it in Sprocket?

peterhuene:
Happy to make it a configuration setting for the backend for now.

Ideally we don't limit it at all within the WDL backend and it can efficiently monitor thousands of tasks submitted to the TES server; the TES API doesn't really allow for it other than short-polling each task individually.

wdl-engine/src/backend/tes.rs
```diff
@@ -0,0 +1,537 @@
...
+                    })
+                    .ty(input.kind())
+                    .read_only(true)
+                    .build(),
+            );
+        }
+
+        let outputs_url = self
+            .config
+            .outputs
```
claymcleod, line 237: Does this mean that this is going to unwrap if the key is not provided in the configuration object? If so, do we want to treat it (and all similar scenarios) in a bit more user-friendly way?

peterhuene: Currently the value is a validated as a required field of the backend's configuration and validation is performed upon the backend's construction.

Ideally it'll be a truly optional field in the future where if an output URL is not specified, we treat it as a shared file system path and treat them as local paths instead.

wdl-engine/src/config.rs
```diff
@@ -448,6 +455,136 @@ impl Default for DockerBackendConfig {
     }
 }
 
+/// Represents information for TES backend basic authentication.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case", deny_unknown_fields)]
+pub struct TesBackendBasicAuth {
```
claymcleod, line 461: Nit: does this really need to be qualified with `TesBackend`? Seems that struct for holding basic auth would be generally useful for other servers later on.

wdl-engine/src/backend/tes.rs
claymcleod: Right, I now am coming back from later in my review where I see that. I think it deserves a `SAFETY` comment to indicate that to the reader.

claymcleod (approved):
Looks good after these comments are addressed.

claymcleod:
> Happy to make it a configuration setting for the backend for now.
> 
> Ideally we don't limit it at all within the WDL backend and it can efficiently monitor thousands of tasks submitted to the TES server; the TES API doesn't really allow for it other than short-polling each task individually.

To this point, what if we just `u64::MAX`ed it and then left as an option configuration setting for limiting it?

peterhuene:
> To this point, what if we just u64::MAXed it and then left as an option configuration setting for limiting it?

I'm okay with that for now provided that we're comitted to improving the TES API itself for more efficiently monitoring lots of in-flight tasks than individual polling.

wdl-engine/src/eval.rs
```diff
@@ -459,46 +459,64 @@ impl EvaluatedTask {
...
+                 `{dir}`{stdout_header}{stdout}{stdout_trailer}{stderr_header}{stderr}{stderr_trailer}",
                 code = self.result.exit_code,
                 dir = self.attempt_dir().display(),
                 stdout_path = self.stdout().as_file().expect("must be file"),
                 stderr_path = self.stderr().as_file().expect("must be file"),
-                header = if stderr.is_empty() {
+                stdout_header = if stderr.is_empty() {
+                    Cow::Borrowed("")
+                } else {
+                    format!("\n\ntask stdout output (last {MAX_STDERR_LINES} lines):\n\n").into()
```
claymcleod, line 511: If we keep this, this should just be `MAX_LINES`.

peterhuene: Oh no, it was just a commit to get some diagnostics out of CI, no intent to keep it.

wdl-engine/src/backend/local.rs
```diff
@@ -130,10 +130,12 @@ impl TaskManagerRequest for LocalTaskRequest {
             )
             .kill_on_drop(true);
 
-        // Set an environment variable on Windows to get consistent PATH searching
-        // See: https://github.com/rust-lang/rust/issues/122660
+        // Set the PATH variable for the child on Windows to get consistent PATH
+        // searching See: https://github.com/rust-lang/rust/issues/122660
```
claymcleod, line 134:
```suggestion
        // searching. See: https://github.com/rust-lang/rust/issues/122660
```

claymcleod:
LGTM.

PR #456, opened 2025-05-23
feat: add `<sprocket-code>` component 

sgarcia-dev:
Added a new web componet called `<sprocket-code>` which when bundled into the docs, allows for code like this:
```
<code-highlight language="wdl">
version 1.0

workflow count_lines {
    input { File input_file }
    call Count { input: file = input_file }
    output { Int num_lines = Count.num_lines }
}

task Count {
    input { File file }
    command { wc -l \${file} | awk '{print $1}' }
    output { Int num_lines = read_int(stdout()) }
}
</sprocket-code>
```

To be automatically be syntax highlighted in WDL with Shiki. Usage example in provided `index.html` file. You must run the build first to generate the bundle. The only downside is the package size is HUGE (800kb) as Shiki relies on a large C based engine compiled to WASM
https://shiki.style/guide/regex-engines#regexp-engines 

I've been unable to find a lightweight alternative that is compatible with our WDL textmate grammar definition. This issue can be sidestepped by building this with a build tool, but since we're using rust, we don't have a JS build step here.


wdl-doc/sprocket-code-wc/sprocket-code.js
```diff
@@ -0,0 +1,90 @@
+import { createHighlighterCore } from 'shiki/core';
+import catppuccinMocha from '@shikijs/themes/catppuccin-mocha';
+import rust from '@shikijs/langs/rust';
+import wdlGrammar from './wdl.tmGrammar.json';
```
claymcleod, line 4: Can we import the grammar dynamically from https://raw.githubusercontent.com/stjude-rust-labs/sprocket-vscode/refs/heads/main/syntaxes/wdl.tmGrammar.json instead of including it verbatim here?

sgarcia-dev: Sure thing, I left it hard coded for now focusing on getting this working and to avoid zero dependencies/network requests so it could be dragged and dropped into sprocket docs's rust code with as little issues as possible. I didn't dig deeper as I wanted us to first look at where were we are and if this is the correct approach/direction we wish to take.

PR #465, opened 2025-06-02
feat: add sprocket web components

sgarcia-dev:
See new README.md for usage guide. Also, side note, I wonder if we should stop tracking generated/dist files?

It adds a lot of changes to PRs that will rarely (if ever) get reviewed, the files are auto-generated meaning versioning achieves nothing, on top of increasing our repository size. Also, I believe most out of the box gitignore configs tend to be configured to ignore generated files like `dist`. We might need to work together to find a list of all files that are generated to untrack them from git and add them to gitignore though


a-frantz:
> Also, side note, I wonder if we should stop tracking generated/dist files?

We need to track them with the current build system, but I can review how we build and perhaps avoid it. In short, the `dist/` files are what's packaged into the `wdl` library and used downstream if `--theme` isn't passed to the `sprocket` binary.

Now that we're adding these web components which need to be built, we'll need to change that approach anyway. So I'll go back to the drawing board and see how we can distribute these build files better

sgarcia-dev:
All styling is handled by Shiki based on the TM grammar file, not controlled by me 😅 I am not sure if Shiki allows us to control this as it's internal behavior, since Shiki's themes are limited to just color palette changes. I could investigate their docs or source code if we feel its a big enough issue though @a-frantz

PR #476, opened 2025-06-11
chore: add a Default implementation for Source

adthrasher:
Adds a `Default` implementation to `Source`. In Sprocket, we want to enable running without specifying a location (similar to `cargo`).

For external contributors:

For all contributors:

      changes (when appropriate).
- [ ] Your PR title follows the [conventional commit] style.

### DIFF
--- a/wdl-cli/src/analysis/source.rs
+++ b/wdl-cli/src/analysis/source.rs
@@ -85,6 +85,13 @@ impl std::str::FromStr for Source {
     }
 }
 
+impl Default for Source {
+    fn default() -> Self {
+        // Default to the current directory.
+        Source::Directory(std::env::current_dir().unwrap_or_else(|_| PathBuf::from(".")))
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;


adthrasher:
> Is the plan to use this in the downstream subcommands for Sprocket (say, in `clap`)?
> 
> I ask this because it might be a better strategy just to have a default implementation live there in the subcommands where it is relevant rather than implement a default implementation here (in a vacuum, I don't really think of there being a _default_ source).

That was my plan. I intended to wrap the source argument in an option downstream and let the default implementation handle cases where the user doesn't provide a source.

a-frantz:
I think whenever possible, we should support an "argument-less" behavior that defaults to the CWD (much like `cargo clippy` etc). Looking over the README for all the commands:

`Source` should default to `pwd`:
- **`sprocket check`** performs static analysis on a document or directory of documents.
- **`sprocket format`** formats a document or directory of documents.
- **`sprocket lint`** performs static analysis on a document or directory of documents with additional linting rules enabled (effectively a shortcut for `check --lint`).
- the future `sprocket lock`

`Source` must be specified:
- **`sprocket inputs`** writes template input file (JSON or YAML) for a task or workflow.
- **`sprocket run`** runs a task or workflow.
- **`sprocket validate`** validates a set of inputs read from files or on the command line against a task or workflow.

N/A:
- **`sprocket analyzer`** runs Sprocket as a LSP server, which is useful for IDE integration.
- **`sprocket completions`** generates shell completions for Sprocket.
- **`sprocket config`** prints configuration values.
- **`sprocket explain`** explains validation and lint rules supported by Sprocket.

acfoltzer:
Whoops, my Github tab only updated with @claymcleod's comments after submitting my review. My comment still stands if we proceed with this originally-proposed default (or implement something similar elsewhere), but my approval should not be interpreted as trying to discard the alternative suggestion.

a-frantz (approved):
I also don't feel strongly between this and https://github.com/stjude-rust-labs/sprocket/pull/127

To Clay's point that this default shouldn't be used for all (or most) cases downstream: a default `usize` of `0` also shouldn't be used in all (or most) `usize` contexts. Having a default doesn't suggest it's appropriate for all cases, so downstream if it doesn't make sense to use this default, we aren't forcing it anywhere 🤷‍♀️

PR #481, opened 2025-06-17
fix: alternative layouts and formatting bugs related to tailwind typography deprecation

sgarcia-dev:
Addressed multiple text/alignment bugs related to tailwind typography deprecate, and added new alternative layout options. Video demo can be found in Slack due to size restrictions

### DIFF
--- a/wdl-doc/theme/dist/style.css
+++ b/wdl-doc/theme/dist/style.css
@@ -1,476 +1,5 @@
 /*! tailwindcss v4.1.10 | MIT License | https://tailwindcss.com */
 @layer properties;
-@layer theme, base, components, utilities;
-@layer theme {
-  :root, :host {
-    --font-sans: "DM Sans";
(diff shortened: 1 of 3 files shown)


a-frantz:
Looks great so far! Some changes I'd like to see:

1. See page `docs/data_structures/flag_filter/index.html`, and you'll see that the the markdown headers and paragraphs are too close to each other. ~Can you introduce an override for these alignments?~ just checked the source, and this looks like a Rust bug I can fix trivially. Seems this markdown section isn't being wrapped in `markdown-body` like it should be. Ignore this for this PR, and I'll introduce the fix after merging 💪 
2. the clickable VS non-clickable breadcrumb elements are no longer being differentiated (without hovering). Can we fix that?
3. see screenshot: is this too wide for prose? Should `markdown-body` have a max-width?
Screenshot 2025-06-17 at 9 58 53 AM

a-frantz (approved):

PR #468, opened 2025-06-04
feat: add support for lsp `goto_definitions`

dead8309:
Added:

- `Struct definitions`
- `Call definitions`
  - Tasks
  - Workflows
- `Namespaced calls`
- `Namespace usage → import declaration`
- `Access expressions`
  - Struct member access
  - Call output access
  - Array element access
  - Chained access expressions
- `Scope-based resolution`
  - Variable references
  - Scatter variables
  - Conditional variables
  - Task inputs/outputs
  - Workflow inputs/outputs
- `Cross document navigation`
- `Aliased imports`
- `Global document resolution`

Before submitting this PR, please make sure:

For external contributors:

- [x] You have read the [CONTRIBUTING](https://github.com/stjude-rust-labs/wdl/blob/main/CONTRIBUTING.md) guide in its entirety.
- [x] You have not used AI on any parts of this pull request.

For all contributors:

- [x] You have added a few sentences describing the PR here.
- [x] Your code builds clean without any errors or warnings.
- [ ] You have added tests (when appropriate).
- [x] You have updated the README or other documentation to account for these
      changes (when appropriate).
- [x] You have added an entry to the relevant `CHANGELOG.md` (see
      ["keep a changelog"] for more information).
- [x] Your commit messages follow the [conventional commit] style.
- [x] Your PR title follows the [conventional commit] style.

[conventional commit]: https://www.conventionalcommits.org/en/v1.0.0/#summary
["keep a changelog"]: https://keepachangelog.com/en/1.1.0/

PR #484, opened 2025-06-18
feat: add support for lsp `textDocument/references`

dead8309:
Adds support for find references to wdl lsp

For external contributors:

For all contributors:

      changes (when appropriate).

### DIFF
--- a/wdl-analysis/src/graph.rs
+++ b/wdl-analysis/src/graph.rs
@@ -21,6 +21,7 @@ use petgraph::stable_graph::StableDiGraph;
 use petgraph::visit::Bfs;
 use petgraph::visit::EdgeRef;
 use petgraph::visit::Visitable;
+use petgraph::visit::Walker;
 use reqwest::Client;
 use rowan::GreenNode;
 use tokio::runtime::Handle;
@@ -627,5 +628,13 @@ impl DocumentGraph {
 
         self.cycles
             .retain(|(from, to)| !collected.contains(from) && !collected.contains(to));
+    }
+
+    /// Gets all nodes that have a dependency on the given node.
+    pub fn transitive_dependents(
+        &self,
+        index: petgraph::graph::NodeIndex,
+    ) -> impl Iterator<Item = NodeIndex> {
+        Bfs::new(&self.inner, index).iter(&self.inner)
     }
 }
--- a/wdl-analysis/src/handlers/find_all_references.rs
+++ b/wdl-analysis/src/handlers/find_all_references.rs
@@ -1,0 +1,177 @@
+//! Handlers for "find all references" requests.
+//!
+//! This module implements the LSP "textDocument/references" functionality for
+//! WDL files. It finds all references to a symbol by first resolving the
+//! symbol's definition, then searches through all the appropriate documents.
+//!
+//! See: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references
+
+use anyhow::Context;
+use anyhow::Result;
+use anyhow::anyhow;
+use lsp_types::Location;
+use url::Url;
+use wdl_ast::AstNode;
+use wdl_ast::SyntaxKind;
+use wdl_ast::TreeToken;
+
+use super::goto_definition::position_to_offset;
+use crate::SourcePosition;
+use crate::SourcePositionEncoding;
+use crate::graph::DocumentGraph;
+use crate::handlers;
+use crate::handlers::location_from_span;
+use crate::handlers::position;
+
+/// Represents a target definition for which references are being searched.
+#[derive(Debug)]
+struct TargetDefinition {
+    /// The identifier text of the target symbol.
+    name: String,
+    /// The location where the target is defined.
+    location: Location,
+}
+
+/// Finds all references to the identifier at the given position.
+///
+/// It first resolves the definition of the identifier at the specified
+/// position, then searches through the appropriate scope of
+/// documents to find all references to that definition.
+pub fn find_all_references(
+    graph: &DocumentGraph,
+    document_uri: Url,
+    position: SourcePosition,
+    encoding: SourcePositionEncoding,
+    include_declaration: bool,
+) -> Result<Vec<Location>> {
+    let definition_location = handlers::goto_definition(graph, document_uri, position, encoding)
+        .context("failed to resolve symbol definition")?
+        .ok_or_else(|| {
+            anyhow!(
+                "no definition location found for symbol at position: {}:{}",
+                position.line,
+                position.character
+            )
+        })?;
+
+    let doc_index = graph
+        .get_index(&definition_location.uri)
+        .ok_or_else(|| anyhow!("definition document not in graph"))?;
+
+    let node = graph.get(doc_index);
+    let document = node
+        .document()
+        .ok_or_else(|| anyhow!("definition document not analyzed"))?;
+
+    let lines = node
+        .parse_state()
+        .lines()
+        .ok_or_else(|| anyhow!("missing line index for target"))?;
+
+    let offset = position_to_offset(
+        lines,
+        SourcePosition::new(
+            definition_location.range.start.line,
+            definition_location.range.start.character,
+        ),
+        encoding,
+    )
+    .context("failed to convert position to offset")?;
+
+    let token = document
+        .root()
+        .inner()
+        .token_at_offset(offset)
+        .find(|t| t.kind() == SyntaxKind::Ident)
+        .ok_or_else(|| anyhow!("could not find target token at definition site"))?;
+
+    let target = TargetDefinition {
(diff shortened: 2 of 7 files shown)


dead8309:
on top of #468. Will rebase it once that gets merged. Thanks!

wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -0,0 +1,227 @@
+//! Implements find all references functionality
```
claymcleod, line 1:
```suggestion
//! Handlers for "find all references" requests.
```

wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -0,0 +1,227 @@
...
+use crate::handlers;
+use crate::handlers::location_from_span;
+use crate::handlers::position;
+
+/// Represents a target definition for which references are being searched.
+#[derive(Debug)]
+struct TargetDefinition {
+    /// The identifier text of the target symbol.
+    name: String,
+    /// The syntax kind of the target's parent node, used to determine reference
```
claymcleod, line 30:
```suggestion
    /// The syntax kind of the target's parent node.
    ///
    /// This is used to determine reference context.
```

wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -0,0 +1,227 @@
...
+        collect_references_from_document(graph, doc_index, &target, encoding, &mut locations)
+            .with_context(|| {
+                format!("failed to collect references from document at index {doc_index:?}")
+            })?;
+    }
+
+    Ok(locations)
+}
+
+/// Resolves the target definition from a definition location.
```
claymcleod, line 74: I don't know what this means just based on this sentence alone. Can we refine this docstring?

wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -0,0 +1,227 @@
...
+    encoding: SourcePositionEncoding,
+    locations: &mut Vec<Location>,
+) -> Result<()> {
+    let root = document.root().inner().clone();
+
+    for token in root
+        .descendants_with_tokens()
+        .filter_map(|el| el.into_token())
+    {
+        if token.kind() == SyntaxKind::Ident && token.text() == target.name {
```
claymcleod, line 202: Is it _guaranteed_ that variables cannot be shadowed in WDL documents that depend on an upstream document? If so, it might be worth adding a comment here explaining that this invariant must hold. Else, to do it a bit more advanced, you should check for a shadow definition of the same name in this document at a minimum.

claymcleod (approved):
This looks pretty good to me! I'll see what Adam thinks being new to this part of the code.

wdl-analysis/src/graph.rs
```diff
@@ -629,3 +629,40 @@ impl DocumentGraph {
...
+pub trait DocumentGraphExt {
+    /// Gets all nodes that have a dependency on the given node.
+    fn dependents(
+        &self,
+        index: petgraph::graph::NodeIndex,
+    ) -> Box<dyn Iterator<Item = petgraph::graph::NodeIndex> + '_>;
+}
+
+impl DocumentGraphExt for DocumentGraph {
+    fn dependents(
```
acfoltzer, line 643: Above, we already have `DocumentGraph::dependencies()` that gets the immediate incoming neighbors of a node. Calling this `dependents()` by analogy implies that it does not do a full traversal for all the dependents. I recommend we call this something like `transitive_dependents()` or `all_dependents()` instead to differentiate.

wdl-analysis/src/graph.rs
```diff
@@ -629,3 +629,40 @@ impl DocumentGraph {
...
+            for edge in self
+                .inner
+                .edges_directed(current_index, petgraph::Direction::Outgoing)
+            {
+                let dependent_index = edge.target();
+                dependents.insert(dependent_index);
+                queue.push(dependent_index);
+            }
+        }
+        Box::new(dependents.into_iter())
```
acfoltzer, line 666: This is a fine implementation of BFS, but since `petgraph` offers a [built-in option](https://docs.rs/petgraph/latest/petgraph/visit/struct.Bfs.html), we should use that.

wdl-analysis/src/graph.rs
```diff
@@ -629,3 +629,40 @@ impl DocumentGraph {
...
+        &self,
+        index: petgraph::graph::NodeIndex,
+    ) -> Box<dyn Iterator<Item = petgraph::graph::NodeIndex> + '_>;
+}
+
+impl DocumentGraphExt for DocumentGraph {
+    fn dependents(
+        &self,
+        index: petgraph::graph::NodeIndex,
+    ) -> Box<dyn Iterator<Item = petgraph::graph::NodeIndex> + '_> {
```
acfoltzer, line 646: Assuming we can make this an inherent (rather than trait) method, we should also be able to simplify this return type:
```suggestion
    ) -> impl Iterator<Item = NodeIndex> {
```

wdl-lsp/src/server.rs
```diff
@@ -741,4 +742,32 @@ impl LanguageServer for Server {
 
         Ok(result)
     }
+
+    async fn references(&self, mut params: ReferenceParams) -> RpcResult<Option<Vec<Location>>> {
+        normalize_uri_path(&mut params.text_document_position.text_document.uri);
+
+        debug!("received `textDocument/gotoDefinition` request: {params:#?}");
```
acfoltzer, line 749: copy/paste error

wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -0,0 +1,217 @@
+//! Handlers for "find all references" requests.
```
acfoltzer, line 1: This would be a good spot to include a high-level description of what this is. Perhaps include a link to the spec, although it is pretty light on details too: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_references

acfoltzer (commented):
Hi Vaibhav,

I'm passing along the comments I've made so far in my review, as I have a hard stop for today and want to get you the feedback ASAP. I will finish reviewing `find_all_references.rs` tomorrow.

So far everything looks great, I just have some style and library design/reuse points.

acfoltzer (changes requested):
Thanks for your patience on this review; I wanted to take my time to really try and understand the subtle aspects of the WDL spec around naming and how this implementation grapples with them.

My main concern about this implementation is trying to make sense of the `SyntaxKind`-based rules for where we are looking for references. It's not clear to me that the visibility of a definition is unambiguous based solely on its syntactic category, e.g. `BoundDeclNode`s which can appear both as private declarations in task/workflow bodies and as `input` or `output` declarations visible to the enclosing scope(s).

I asked more specific questions in a comment on `determine_search_scope()`, but in general I'm curious what other approaches we might take here. I am still learning the codebase, but I see we build up scope structures and dependency dags between tasks and workflows for other analysis steps. Is there possibly room for tracking the use sites of declarations as we perform the lookups necessary for, e.g., typechecking?

Finally, if `determine_search_scope()` is used solely to improve performance rather than as a crucial part of returning correct results, perhaps a good first step would be to ship an implementation that doesn't bother filtering out irrelevant scopes. Inefficient, yes, but perhaps not a noticeable hit for most WDL workspaces?

wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -0,0 +1,244 @@
...
+/// Determines the search scope for finding references based on the target
+/// definition type.
+///
+/// # Returns
+///
+/// A vector of document indices that should be searched for references:
+/// - For local symbols: only the target document
+/// - For imported symbols: the target document + all dependent documents
+/// - For unknown symbol types: defaults to target document only
+fn determine_search_scope(
```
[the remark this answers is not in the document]
dead8309: 1. yes only for optimization, the correctness is ensured by `goto_definition`
2. Oops i added those which were supported by goto-definition to test out things after adding `unimplemented!()` for default case. 

It will only have definition `SyntaxKind`. Removed👍

wdl-analysis/src/handlers/find_all_references.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        _ => unimplemented!("find all references not implemented for: {:?}", target.kind),
+    }
+}
+
+/// Collects references to the target symbol form a single document.
+///
+/// 1. Traverse all tokens in the document's CST
+/// 2. Filter for identifier tokens matching the target name
+/// 3. For each match, resolve its definition using goto definition
```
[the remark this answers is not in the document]
dead8309: yes call and workflow inputs were not added, opening a pr
image

acfoltzer (approved):
Thank you!!

PR #490, opened 2025-06-24
fix: align left sidebar with Figma

sgarcia-dev:
Changelog in Slack

### DIFF
--- a/wdl-doc/src/docs_tree.rs
+++ b/wdl-doc/src/docs_tree.rs
@@ -471,14 +471,13 @@ impl DocsTree {
         html! {
             @for (category, workflows) in workflows_by_category {
                 li class="" {
-                    div class="left-sidebar__category" {
+                    div class="left-sidebar__content-item w-full" {
                         img src=(self.get_asset(base, "category-selected.svg")) class="left-sidebar__icon" alt="Category icon";
                         p class="" { (category) }
                     }
                     ul class="" {
                         @for node in workflows {
-                            li x-data=(format!(r#"{{
-                                    hover: false,
+                            a href=(diff_paths(self.root_abs_path().join(node.path()), base).expect("should diff paths").to_string_lossy()) x-data=(format!(r#"{{
                                     node: {{
                                         current: {},
                                         icon: '{}',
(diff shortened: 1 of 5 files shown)


wdl-doc/src/docs_tree.rs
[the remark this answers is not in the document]
sgarcia-dev: Actually, that's not necessarily true. Tailwind CSS best practices actually **recommend** inlining the styles as it helps with colocation. As reuse is usually achieved using "Components" (templates)

https://tailwindcss.com/docs/styling-with-utility-classes#managing-duplication 

Our case is somewhat different/unique in that our templates are tightly coupled with Rust code and don't follow separation of concerns best practices between Server/Frontend code, and do not promote component re-use as well. So I've been trying to strike a balance. My general rules are:

- Anything that will be reused by markup should be moved to a class
- Anything that is only present in a single place can be inlined

Having said that, even this approach isn't perfect because Rust Templates lack access to typical Front End tooling meaning we don't get things like linters/autocomplete and its very easy for uncaught typos and syntax errors to exist in the "frontend" rust code without us ever noticing compared to other common/popular templating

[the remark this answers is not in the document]
a-frantz: There is another concern we have to consider. We will have some users who will want to customize the CSS used to render _their_ docs site, and we do want to support that use case in some way. Ideally, we want users to able to modify the final CSS without recompiling the rust binaries. The less inlined CSS we have, the better we can support this use case.

Granted, this is a use case that's rather low on our priority list (especially at this point in development), but some extra context for you and part of my request.

wdl-doc/src/docs_tree.rs
```diff
@@ -788,28 +786,26 @@ impl DocsTree {
...
+                        img src=(self.get_asset(base, "x-mark.svg")) class="absolute right-2 top-1/2 -translate-y-1/2 size-6 hover:cursor-pointer" alt="Clear icon" x-show="search !== ''" x-on:click="search = ''";
                     }
-                    div class="left-sidebar__tabs-container" {
-                        div x-on:click="showWorkflows = true; search = ''" class="left-sidebar__tabs" x-bind:class="! showWorkflows ? 'text-slate-400 hover:text-slate-300' : 'text-slate-50 border-b-slate-50'" {
+                    div class="left-sidebar__tabs-container mt-4" {
+                        button x-on:click="showWorkflows = true; search = ''" class="left-sidebar__tabs" x-bind:class="! showWorkflows ? 'text-slate-400 hover:text-slate-300' : 'text-slate-50 border-b-slate-50'" {
                             img src=(self.get_asset(base, "list-bullet-selected.svg")) class="left-sidebar__icon" alt="List icon";
                             p { "Workflows" }
                         }
-                        // TODO clicking this should scroll the active page into view
```
a-frantz, line 805: This TODO doesn't look fixed to me. Can you put this comment back?

wdl-doc/src/docs_tree.rs
```diff
@@ -846,8 +844,9 @@ impl DocsTree {
                         li class="flex place-content-center" {
                             img x-show="search !== '' && searchedNodes.length === 0" src=(self.get_asset(base, "search.svg")) class="size-8" alt="Search icon";
                         }
-                        li class="flex place-content-center" {
-                            p x-show="search !== '' && searchedNodes.length === 0" class="" x-text="'No results found for \"' + search + '\"'" {}
+                        li x-show="search !== '' && searchedNodes.length === 0" class="flex gap-1 place-content-center text-center break-words whitespace-normal text-sm text-slate-500" {
+                            span x-text="'No results found for '" {}
+                            span x-text="` \"${search}\"`" class="text-slate-50" {}
```
a-frantz, line 849:
```suggestion
                            span x-text="`\"${search}\"`" class="text-slate-50" {}
                            span x-text="'" {}
```
I think there's a missing trailing comma. This should fix it

wdl-doc/src/docs_tree.rs
sgarcia-dev: Nice catch! Added

a-frantz (approved):

PR #502, opened 2025-07-01
fix: multiple reported bugs

sgarcia-dev:
Changelog

*   Left Sidebar
    *   Row icons are shrinking
    *   Row item styles for hover/focus incorrect
    *   Folder collapse/expand state and styles missing
    *   When sidebar tab is inactive, folder icon remains white
    *   Alternative Sidebar Layout user preference should be persisted to session storage, which takes precedence over defaulting logic.
    *   Alternative Sidebar Layout buttons need better icons ([@ira.baranova](https://stjude.slack.com/team/U03PB30452L) to define)
*   Center Column
    *   Breadcrumbs should be centered with content
    *  Left sidebar control buttons should also be centered
*   Content Changes
    *   Run button styles can be improved.
    *   Tables are missing styles


wdl-doc/src/runnable.rs
```diff
@@ -153,8 +155,9 @@ pub(crate) trait Runnable {
                     div class="main__grid-header-cell" { "Name" }
                     div class="main__grid-header-cell" { "Type" }
                     div class="main__grid-header-cell" { "Description" }
-                    div class="main__grid-header-separator" {}
+                    // div class="main__grid-header-separator" {}
                     @for param in iter {
+                        div class="main__grid-row-separator" {}
```
sgarcia-dev, line 160: Hello @a-frantz  👋 in order to add separators between rows, we need to update all tables on the site to render both a `main__grid-header-separator` element under the header cells, as well as `main__grid-row-separator` between regular rows. I did not perform the changes because the table code is duplicated (you'll locate it quicker than I would), and the iterator logic in rust. As I already wasted some time already trying to figure out how the images were being processed somewhat strangely in rust  😅

I think it should be doable to do something like `if (currentRowIndex < totalRows) renderSeparator` as that would ensure we don't accidentally render a separator under the final row.

PR #519, opened 2025-07-08
feat: add support for auto-completions

dead8309:
For external contributors:

For all contributors:

      changes (when appropriate).

### DIFF
--- a/wdl-lsp/tests/completions.rs
+++ b/wdl-lsp/tests/completions.rs
@@ -450,7 +450,8 @@ async fn should_complete_versions() {
     let Some(CompletionResponse::Array(items)) = response else {
         panic!("expected a response, got none");
     };
-    println!("{items:?}");
 
     assert_contains(&items, "1.0");
+    assert_contains(&items, "1.1");
+    assert_contains(&items, "1.2");
 }


wdl-analysis/src/handlers/completions.rs
```diff
@@ -0,0 +1,453 @@
...
+                detail: Some(p.left_type().to_string()),
+                ..Default::default()
+            });
+
+            items.push(CompletionItem {
+                label: "right".to_string(),
+                kind: Some(CompletionItemKind::FIELD),
+                detail: Some(p.right_type().to_string()),
+                ..Default::default()
+            });
```
dead8309, line 308: _This was done as an experiment_

TODO: I don't know how but if time permits in a future PR i would want to do _inferred_ completions for maps.

```wdl
version 1.2

workflow test_map {
  Map[String, Int] string_to_int = { "a": 1, "b": 2 }
  Map[File, Array[Int]] file_to_ints = {
    "/path/to/file1": [0, 1, 2],
    "/path/to/file2": [9, 8, 7]
  }

  output {
    Int b = string_to_int["b"]  # evaluates to 2
    Array[Int] ints = file_to_ints["/path/to/file1"]  # evaluates to [0, 1, 2]
  }
}
```

so someone typing `file_to_ints["`, LSP should suggest  `"/path/to/file1"`, `"/path/to/file2"` in completions list. Obvisouly this would only work if they can be inferred.

Not a scope for this PR tho

wdl-analysis/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 #### Added
 
+* Added `auto-completions` support for WDL Language Server ([#519](https://github.com/stjude-rust-labs/wdl/pull/519)).
```
claymcleod, line 12:
```suggestion
* Added autocompletion support for WDL Language Server ([#519](https://github.com/stjude-rust-labs/wdl/pull/519)).
```

claymcleod, line 12:
```suggestion
* Added code completion support for the WDL Language Server ([#519](https://github.com/stjude-rust-labs/wdl/pull/519)).
```

wdl-analysis/src/handlers/completions.rs
```diff
@@ -0,0 +1,453 @@
...
+/// kinds, and descriptions.
+fn add_keyword_completions(token_set: &TokenSet, items: &mut Vec<CompletionItem>) {
+    items.extend(token_set.iter().map(|raw| {
+        let token = Token::from_raw(raw);
+        let label = token
+            .describe()
+            .trim_start_matches("`")
+            .split("`")
+            .next()
+            .unwrap_or("")
```
claymcleod, line 189: Should this panic instead? I don't think adding a empty completion item seems right here. I could be wrong, let me know.

wdl-analysis/src/handlers/completions.rs
```diff
@@ -0,0 +1,453 @@
...
+
+/// Adds completions for member access expressions.
+///
+/// Handles different types of member access completions:
+/// - Namespace access
+/// - Struct member access
+/// - Call output access
+/// - Pair element access (when accessing `.left` and `.right` of pair types)
+///
+/// For namespace access, it directly looks up the identifier before the dot.
```
claymcleod, line 209: Can you give a little more on what `node` is here? What's expected to be contained within?

wdl-analysis/src/handlers/completions.rs
```diff
@@ -0,0 +1,453 @@
...
+                        ..Default::default()
+                    });
+                }
+
+                return Ok(());
+            }
+        }
+    }
+
+    // NOTE: We do type evaluation only for non namespaces or complex types
```
claymcleod, line 256:
```suggestion
    // NOTE: we do type evaluation only for non namespaces or complex types
```

wdl-analysis/src/handlers/completions.rs
```diff
@@ -0,0 +1,453 @@
...
+
+    let Some(target_node) = target_element.as_node() else {
+        return Ok(());
+    };
+
+    let Some(target_expr) = Expr::cast(target_node.clone()) else {
```
claymcleod, line 262: So at this point in the code, I see we're tracking this down as an expression on the left side of a dot. Can you give me an example of what this might look like in practice? Do we have a test for this case?

wdl-analysis/src/queue.rs
```diff
@@ -654,51 +711,59 @@ where
         }
 
         let total = tasks.len();
-        self.tokio
-            .block_on((self.progress)(context.clone(), kind, 0, total));
+        if completed.is_some() {
```
claymcleod, line 714: Is this right? Seems like we shoudl send progress updates even if the completed callback is not provided. Perhaps I'm missing something with this.

wdl-analysis/src/stdlib.rs
```diff
@@ -1651,26 +1758,150 @@ pub static STDLIB: LazyLock<StandardLibrary> = LazyLock::new(|| {
                         .parameter(PrimitiveType::Integer)
                         .parameter(PrimitiveType::Integer)
                         .ret(PrimitiveType::Integer)
+                        .definition(
```
claymcleod, line 1761: For polymorphic signatures, let's `const`-ize the definition just above the declaration and refer to that instead of duplicating.

wdl-analysis/src/stdlib.rs
```diff
@@ -1759,6 +2188,41 @@ pub static STDLIB: LazyLock<StandardLibrary> = LazyLock::new(|| {
                         .parameter(PrimitiveType::String)
                         .parameter(PrimitiveType::String)
                         .ret(PrimitiveType::String)
+                        .definition(
+                            r#"
+Given three String parameters `input`, `pattern`, `replace`, this function replaces all non-overlapping occurrences of `pattern` in `input` by `replace`. `pattern` is a [regular expression](https://en.wikipedia.org/wiki/Regular_expression) and is evaluated as a [POSIX Extended Regular Expression (ERE)](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended).
```
claymcleod, line 2193:
```suggestion
Given three `String` parameters `input`, `pattern`, `replace`, this function replaces all non-overlapping occurrences of `pattern` in `input` by `replace`. `pattern` is a [regular expression](https://en.wikipedia.org/wiki/Regular_expression) and is evaluated as a [POSIX Extended Regular Expression (ERE)](https://en.wikipedia.org/wiki/Regular_expression#POSIX_basic_and_extended).
```

wdl-analysis/src/stdlib.rs
claymcleod: Some of the polymorphic signatures have specific markup for the arguments. Those should stay as they are (e.g., a stdlib function that can take an object, map, and dict). For ones like this one though, where the content stays the same across the polymorphic signatures, let's `const`.

wdl-lsp/tests/workspace/completions/source.wdl
```diff
@@ -0,0 +1,27 @@
...
+        bar: Bar {
+            baz: "hello",
+        },
+    }
+
+    call util.greet as t1 { name = my_foo.bar.baz }
+
+    String n = my_foo.
+
+    output  {
```
claymcleod, line 24:
```suggestion
    output {
```

wdl-analysis/src/queue.rs
dead8309: This was to make the analysis **block** in case on completions and **cancellable** in other cases - what i asked in slack that we would need analyzed document to access methods like `structs`, `namespaces`, `tasks` etc on it

wdl-analysis/src/stdlib.rs
dead8309: What if we use those parameter and return types from function signatures  instead of a generic `Float | Int` 🤔

wdl-analysis/src/handlers/completions.rs
dead8309: yes 
```rs
#[tokio::test]
async fn should_complete_struct_members() {
    let mut ctx = setup().await;

    // Position of cursor `String n = my_foo.`
    let response = completion_request(&mut ctx, "source.wdl", Position::new(21, 22)).await;
    let Some(CompletionResponse::Array(items)) = response else {
        panic!("expected a response, got none");
    };

    assert_eq!(items.len(), 1, "should only complete the single member");
    assert_contains(&items, "bar");
    assert_not_contains(&items, "baz");
}

#[tokio::test]
async fn should_complete_with_partial_word() {
    let mut ctx = setup().await;
    // Position of cursor at `Int out = qux.n`
    let response = completion_request(&mut ctx, "partial.wdl", Position::new(13, 23)).await;
    let Some(CompletionResponse::Array(items)) = response else {
        panic!("expected a response, got none");
    };

    assert_eq!(items.len(), 1, "should only have a single item");
    assert_contains(&items, "num");
}

#[tokio::test]
async fn should_complete_namespace_members() {
    let mut ctx = setup().await;
    // Position of cursor at `call lib.`
    let response = completion_request(&mut ctx, "namespaces.wdl", Position::new(5, 13)).await;
    let Some(CompletionResponse::Array(items)) = response else {
        panic!("expected a response, got none");
    };

    assert_eq!(items.len(), 1);
    assert_contains(&items, "greet");
}
```

wdl-analysis/src/handlers/completions.rs
```diff
@@ -0,0 +1,720 @@
...
+            kind: Some(CompletionItemKind::MODULE),
+            detail: Some(format!("import alias {name}")),
+            ..Default::default()
+        });
+    }
+}
+
+/// Adds completions for the members of the implicit `task` variable.
+fn add_task_variable_completions(items: &mut Vec<CompletionItem>) {
+    const TASK_FIELDS: &[&str] = &[
```
claymcleod, line 561: For every field that is a `const` like this, I'd prefer if we defined the arrays in `wdl-grammar` right next to the keywords. The reason for this is I want all of this kind of information to be stored in one part of the codebase. It's much more likely that, if, for example, a new keyword is added in the future, the engineer will update the list if it lives there than if it is buried here in the LSP code. We can still import it and use it here obviously.

claymcleod (changes requested):
Here are a few more changes. Getting very close!

claymcleod:
1. Can we add version autocompletion (1.0, 1.1, 1.2).
2. Can we give a description to each of the task variables? Right now, there is no text describing what they represent, just the type.

wdl-analysis/src/handlers/completions.rs
```diff
@@ -0,0 +1,720 @@
...
+            if n.text() == TASK_VAR_NAME
+                && document.version()
+                    >= Some(wdl_ast::SupportedVersion::V1(wdl_ast::version::V1::Two))
+            {
+                if let Some(parent) = n.parent() {
+                    if matches!(
+                        parent.kind(),
+                        SyntaxKind::CommandSectionNode | SyntaxKind::OutputSectionNode
+                    ) {
+                        add_task_variable_completions(items);
```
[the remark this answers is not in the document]
dead8309: SPEC had examples of `task` variable usage under `command` and `output`. 

Removing this condition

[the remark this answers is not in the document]
dead8309: We do check for
1. If its a `Dot` token
2. If the `Dot` token has a parent/previous sibling
3. If the `Dot` token's previous sibling was a `Node`
4. If that `Node`'s text is `task`
5. If the currernt document version is `1.2`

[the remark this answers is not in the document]
claymcleod: Oh! I see it now. Okay thanks.

wdl-lsp/src/server.rs
```diff
@@ -398,7 +398,7 @@ impl LanguageServer for Server {
                 references_provider: Some(OneOf::Left(true)),
                 completion_provider: Some(CompletionOptions {
                     resolve_provider: Some(false),
-                    trigger_characters: Some(vec![".".to_string(), " ".to_string()]),
+                    trigger_characters: Some(vec![".".to_string()]),
```
dead8309, line 401: Added that for a better ux - like user typing `version` then hitting space would show `1.0`,`1.1`,`1.2`, this is great but it was causing a lot of unwanted completions requests for other sections.

With the current changes. `version` completions would work as:
1.  When user types `version` and triggers auto completion (CTRL+SPACE)
2. When user types `version` and then starts typing a known supported version `1.*`

claymcleod (approved):

PR #534, opened 2025-07-17
refactor: use `libtest-mimic` for custom test executables

acfoltzer:
This change adopts [`libtest-mimic`](https://docs.rs/libtest-mimic) for all test targets that do not use the standard Rust test harness.

A number of the test targets in the codebase use `harness = false` in their `Cargo.toml` definitions. This is necessary because they build their test targets dynamically from directories and files present nearby when the tests are run, rather than from Rust functions annotated with `#[test]`.

`libtest-mimic` provides functions for replicating the command-line interface and runtime semantics of the standard Rust test harness as much as possible while still allowing for dynamic logic to generate test targets (called `Trial`s). This allows us to remove several repetitive elements from most of these existing tests:

- command-line parsing code to filter which tests to run
- unwind/panic handling
- result tallying and terminal output formatting
- explicit parallelization

Beyond the saved maintenance burden, compatibility with the standard test harness sets us up for using [`cargo-nextest`](https://nexte.st/) as a test runner both locally and in CI, which offers a variety of potential benefits (though currently appears to be slower than the normal runner).

I also noticed that the `wdl-doc/tests/doc.rs` target does not dynamically generate test targets, so I converted it to use the standard test harness. Since `libtest-mimic` does not perfectly replicate the behavior of the standard test harness, this is the preferred option when possible.

Removal of shared `Analyzer`s
-----------------------------

When restructuring several of these tests, I moved us away from a model of a single `wdl_analysis::Analyzer` being used for an entire test suite. Although the filtering of shared analysis results seems like it's been effective for isolating (mis)behavior of individual tests, this type of interdependency between tests can lead to unexpected and inconsistent behavior.

Since I'm newer to the project, I want to make sure that this change does not inadvertently reduce the confidence we derive from these tests, though. Reviewers, please let me know if it'd be valuable to recover some of the shared analysis behavior.

Improved performance 🚀
----------------------

These figures are not particularly scientific, just a couple of runs each on my M4 Max with all the test artifacts prebuilt with `cargo build --all-features --workspace --tests`.

Before:

```
cargo test --all-features --workspace  7.41s user 7.91s system 71% cpu 21.478 total
```

After:

```
cargo test --all-features --workspace  7.23s user 8.44s system 116% cpu 13.464 total
```

**Running our full test suite locally now takes only ~65% of the time it did previously! 🎉**

Strangely, `nextest` seems to take quite a bit _longer_. This is inconsistent with my past experience with that tool, but I have not before used it in combination with `libtest-mimic` so there might be some unfortunate interaction there:

```
( cargo nextest run --all-features --workspace && cargo test --all-features --workspace --doc )  5.42s user 4.18s system 27% cpu 34.933 total
```

---

For all contributors:

      changes (when appropriate).


claymcleod (approved):
Looks like a solid change. Good job!

PR #538, opened 2025-07-23
fix: enable `pair` serialization to `inputs.json` and `outputs.json` during execution

adthrasher:
Serialize `Pair` as `Object` for execution-level `inputs.json` and `outputs.json`.

For external contributors:

For all contributors:

      changes (when appropriate).

### DIFF
--- a/wdl-engine/src/value.rs
+++ b/wdl-engine/src/value.rs
@@ -3005,7 +3005,7 @@ pub struct ValueSerializer<'a> {
 
 impl<'a> ValueSerializer<'a> {
     /// Constructs a new `ValueSerializer`.
-    pub(crate) fn new(value: &'a Value, allow_pairs: bool) -> Self {
+    pub fn new(value: &'a Value, allow_pairs: bool) -> Self {
         Self { value, allow_pairs }
     }
 }
@@ -3043,7 +3043,7 @@ pub struct CompoundValueSerializer<'a> {
 
 impl<'a> CompoundValueSerializer<'a> {
     /// Constructs a new `CompoundValueSerializer`.
-    pub(crate) fn new(value: &'a CompoundValue, allow_pairs: bool) -> Self {
+    pub fn new(value: &'a CompoundValue, allow_pairs: bool) -> Self {
         Self { value, allow_pairs }
     }
 }


peterhuene:
I'd also be good with supporting this PR in the short term and pushing the `Pair` serialization format and the coercions upstream to the spec.

We'd _technically_ be off spec, but to me this is a failure of the spec that's causing unnecessary problems for users.

adthrasher:
> I'd also be good with supporting this PR in the short term and pushing the `Pair` serialization format and the coercions upstream to the spec.
> 
> We'd _technically_ be off spec, but to me this is a failure of the spec that's causing unnecessary problems for users.

Do you (or @claymcleod) have an impression of how difficult it would be to get this upstream in the spec?

peterhuene:
I think it wouldn't be considered a breaking change (in my opinion, removing an explicit error condition is backwards compatible), so I could see it being adopted in 1.3, but progress on 1.3 is quite slow right now as the broader ecosystem is still trying to digest 1.1 and 1.2.

My first inclination is to fix it for `inputs.json` and `outputs.json` (which should also fix printing the final outputs at the end of Sprocket's evaluation) to unblock flowing a `Pair` through a workflow internally and then do battle with the spec to get `Pair` to work as something users can serialize with `write_json` or pass an input from a JSON inputs file.

wdl-engine/src/inputs.rs
```diff
@@ -310,8 +312,27 @@ impl Serialize for TaskInputs {
     where
         S: serde::Serializer,
     {
-        // Only serialize the input values
-        self.inputs.serialize(serializer)
+        let mut map = serializer.serialize_map(Some(self.inputs.len()))?;
+        self.inputs.iter().for_each(|(key, value)| match value {
```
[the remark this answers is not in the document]
peterhuene: Instead, we could refactor the current serialization implementation of `Value` and `CompoundValue` out into new wrapper types `SerializableValue<'a>(&'a Value, bool)` and `SerializableCompoundValue<'a>(&'a CompoundValue, bool)`, where the second value is a flag for whether or not we serialize `Pair`.

Then we could implement `Serialize` for `Value` (and likewise for `CompoundValue`) like so:

```rust
impl serde::Serialize for Value {
    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> {
        SerializableValue::new(self, false).serialize(serializer)
    }
}
```

And use `SerializableValue::new(v, true)` in the implementations of `Serialize` for `WorkflowInputs`, `TaskInputs`, and `Outputs`.

wdl-engine/src/outputs.rs
```diff
@@ -65,28 +65,12 @@ impl Serialize for Outputs {
...
-        }
-
-        impl serde::Serialize for Serialize<'_> {
-            fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
-            where
-                S: serde::Serializer,
-            {
-                self.value.serialize(serializer)
-            }
-        }
```
adthrasher, line 81: @peterhuene - I'm not certain I fully understand what this was doing before.

wdl-engine/src/inputs.rs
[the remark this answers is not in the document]
adthrasher: Thanks @peterhuene, I'd already started down the `serializable_with_pairs` route, but can revisit if you prefer the `SerializableValue` approach.

wdl-engine/src/value.rs
```diff
@@ -2353,6 +2375,32 @@ impl CompoundValue {
...
+                match self.0 {
+                    CompoundValue::Pair(pair) => {
+                        let mut state = serializer.serialize_map(Some(2))?;
+                        let left = pair.left().serializable_with_pairs();
+                        let right = pair.right().serializable_with_pairs();
+                        state.serialize_entry("left", &left)?;
+                        state.serialize_entry("right", &right)?;
+                        state.end()
+                    }
+                    _ => self.0.serialize(serializer),
```
peterhuene, line 2397: We can't fall back to the default `serialize` implementation here as there might be more pair values contained internally; for example, we should be able to serialize a `Map[String, Pair[String, String]]` just fine, but here we're falling back to `serialize` for `Map`, so its value gets serialized without support for pairs.

I think the easiest thing to do is to extract out the existing serialization logic from `Value` and `CompoundValue` into a `ValueSerializer` and `CompoundValueSerializer` types that impl `serde::Serialize`.  We remove the implementation of `serde::Serialize` from `Value` and `CompoundValue` so that we _must_ use `ValueSerializer` and `CompoundValueSerializer` to serialize a value and that will catch all of the places we're currently relying on the implementation that doesn't allow serializing `Pairs`.

In the `CompoundValueSerializer` implementaiton, we have `CompoundValue::Pair(pair) if self.allow_pairs => { ... }` and a `CompoundValue::Pair(_) if !self.allow_pairs { /* error */ }` in the match. For every place where we serialize an inner value, we construct a new `ValueSerializer` for it, but we pass through the `allow_pairs` flag.

wdl-engine/src/inputs.rs
```diff
@@ -311,7 +312,15 @@ impl Serialize for TaskInputs {
         S: serde::Serializer,
     {
         // Only serialize the input values
-        self.inputs.serialize(serializer)
+        let mut map = serializer.serialize_map(Some(self.inputs.len()))?;
+        self.inputs.iter().for_each(|(k, v)| {
+            let serialized_value = v.serializable_with_pairs();
+            map.serialize_entry(k, &serialized_value)
+                .unwrap_or_else(|e| {
+                    panic!("failed to serialize value `{v}` for input `{k}`: {e}");
```
peterhuene, line 320: I'd be more comfortable bubbling up the serialization error here instead of panicking.

wdl-engine/src/inputs.rs
```diff
@@ -630,7 +639,15 @@ impl Serialize for WorkflowInputs {
     {
         // Note: for serializing, only serialize the direct inputs, not the nested
         // inputs
-        self.inputs.serialize(serializer)
+        let mut map = serializer.serialize_map(Some(self.inputs.len()))?;
+        self.inputs.iter().for_each(|(k, v)| {
+            let serialized_value = v.serializable_with_pairs();
+            map.serialize_entry(k, &serialized_value)
+                .unwrap_or_else(|e| {
+                    panic!("failed to serialize value `{v}` for input `{k}`: {e}");
```
peterhuene, line 647: Same comment as above.

wdl-engine/src/outputs.rs
peterhuene: I think this might have been a hold over from a time when `Value` wasn't directly serializable; it does seem entirely unnecessary.

adthrasher:
Thanks @peterhuene !

PR #543, opened 2025-07-28
feature: wdl-engine options to allow exceeding max CPU and memory

acfoltzer:
In response to https://github.com/stjude-rust-labs/sprocket/issues/144.

This adds two new configuration options on the task section of `wdl-engine`'s config struct that allow task execution to be attempted even if the known maximum value of either CPU or memory is exceeded by a task's requirements. It does so by clamping the requirement to that max, logging a warning, and then continuing.

This is only tested for the local backend, and for the Docker backend when not configured to use a swarm. For other backends, `wdl-engine` will still attempt to run tasks that exceed limits, but since the actual maximums are managed by other systems, behavior is not guaranteed.

Given the lack of consistency across backends, we're not going to pursue the addition of new command-line flags to `sprocket` at this time. Advanced users are able to configure this via `sprocket.toml` if desired.

Tests are added in the `wdl-engine/tests/tasks` suite, which has also been extended to look for and use potentially-platform-specific serialized config files instead of hardcoded default configs. This is necessary to exercise the behavior of the new flags, and could be extended to other test suites in the future if desired.

      changes (when appropriate).

### DIFF
--- a/wdl-engine/CHANGELOG.md
+++ b/wdl-engine/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Added
+
+* Added `cpu_limit_behavior` and `memory_limit_behavior` options to task execution configuration ([#543](https://github.com/stjude-rust-labs/wdl/pull/543))
+
 #### Changed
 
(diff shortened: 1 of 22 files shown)


a-frantz (approved):
Looks great! Thanks Adam

PR #553, opened 2025-07-31
fix: remove `wdl-engine` task test config override

acfoltzer:
I inadvertently left some temporary debugging changes in when #543 was merged.

This happy accident revealed we didn't have any tests in this part of the suite exercising the limiting behavior for CPU and memory, so I added a couple and validated that they would have failed if present at the time of my committing the wrong code.

Unfortunately getting the new tests to match the expected output was not as simple as just running a single `BLESS=1` invocation, as the error messages for exceeded resources include the host- or Docker-derived maximums found at runtime.

I considered extending the test fixture's `normalize()` function to rewrite the messages specific to these tests, but (1) hardcoding for specific test output in this way is very gross and (2) would require the escalation of pulling in `regex` in order to accomplish vs the literal search/replace currently performed by `normalize()` for path separators and line endings.

Instead of this, I opted to add a new `wdl-engine` config flag that signals a desire to suppress environment-specific items from logging output and error messages. There are, of course, many more parts of the code that emit env-specific output, some unavoidably so. For the purposes of this PR, I only addressed the locations in question for these tests, and tried my best to caveat the incompleteness of the flag's implementation in its docstring.

      changes (when appropriate).


a-frantz (dismissed):
apologies, this fell of my radar!

This seems like a fine band-aid to me, but I would hope we can find a better solution in the future. Good to merge once the conflicts are resolved IMO 👍

a-frantz (approved):

PR #565, opened 2025-08-06
feat: ignore files

a-frantz:
Describe the problem or feature in addition to a link to the issues._

for #539 which will stay open till this makes it into Sprocket

see https://github.com/stjude-rust-labs/sprocket/pull/158

For external contributors:

For all contributors:

      changes (when appropriate).


wdl-analysis/src/analyzer.rs
```diff
@@ -409,11 +409,22 @@ where
...
-            for result in WalkDir::new(&path).follow_links(true) {
-                let entry = result.with_context(|| {
-                    format!("failed to read directory `{path}`", path = path.display())
-                })?;
-                if !entry.file_type().is_file()
+            let walker = WalkBuilder::new(&path)
+                .follow_links(true)
+                .add_custom_ignore_filename(".sprocketignore")
+                .standard_filters(false) // don't use git options
+                .hidden(true)
```
a-frantz, line 416: should we or should we not search hidden dirs for WDL files?

wdl-analysis/Cargo.toml
```diff
@@ -32,7 +33,6 @@ tokio = { workspace = true }
 tracing = { workspace = true }
 url = { workspace = true }
 uuid = { workspace = true, features = ["v4"] }
-walkdir = { workspace = true }
```
a-frantz, line 35: `walkdir` is used in `wdl-engine`. Should it be replaced there with the `ignore` crate so we can drop this dep? Or is that not worth dealing with?

wdl-analysis/src/analyzer.rs
```diff
@@ -409,11 +414,19 @@ where
...
+            if let Some(ignore_filename) = config.ignore_filename() {
+                walker.add_custom_ignore_filename(ignore_filename);
+            }
+            let walker = walker.standard_filters(false).parents(true).build();
+
+            for result in walker {
+                let entry = match result {
+                    Ok(entry) => entry,
+                    Err(_) => continue, // ignore any errors parsing the ignore files
+                };
```
acfoltzer, line 427: I don't think this is the behavior we want... an error in an ignore file would make us skip every entry, the result being we'd silently add nothing to the analyzer context. Better to `?` this error out, I'd say (probably with an informative `context`).
```suggestion
                let entry = entry?;
```

wdl-analysis/src/config.rs
```diff
@@ -108,6 +114,26 @@ impl Config {
...
+    ///
+    /// Specifying `None` for `filename` disables ignore behavior. This is also
+    /// the default.
+    ///
+    /// `Some(filename)` will use `filename` as the ignorefile basename to
+    /// search for. Child directories _and_ parent directories are searched
+    /// for a file with the same basename as `filename` and if a match is
+    /// found it will attempt to be parsed as an ignorefile. Errors during
+    /// this parsing are discarded, as it is a non-critical feature and should
+    /// not block further anlaysis.
```
a-frantz, line 129: should this doc comment link out to something that defines ignorefiles, or is it just expected anyone this deep in the guts of `wdl_analysis` is familiar with them?

wdl-analysis/src/analyzer.rs
```diff
@@ -409,11 +414,19 @@ where
...
+
+                if !entry.file_type().is_some_and(|e| e.is_file())
                     || entry.path().extension().and_then(OsStr::to_str) != Some("wdl")
                 {
                     continue;
```
acfoltzer, line 432: Whoa, TIL about `Option::is_some_and()`, that's a neat function! Between that, the negation, and the `||`ing with another condition, though, I think it is difficult to read and understand the intent here. It also made it easy to miss that an entry without a file type would actually fall through and attempt to get analyzed, probably exploding at some later point.

Here's how I would rewrite this. It's more verbose and is somewhat repetitive, but imo requires less mental overhead to understand.

```suggestion
                // Skip entries without a file type
                let Some(file_type) = entry.file_type() else {
                    continue;
                };
                // Skip non-files
                if !file_type.is_file() {
                    continue;
                }
                // Skip files without a `.wdl` extension
                if entry.path().extension() != Some(&OsStr::new("wdl")) {
                    continue;
```

wdl-analysis/src/config.rs
```diff
@@ -119,6 +145,8 @@ struct ConfigInner {
     /// See [`Config::with_fallback_version()`]
     #[serde(default)]
     fallback_version: Option<SupportedVersion>,
+    /// See [`Config::with_ignore_filename()`]
+    ignore: Option<String>,
```
acfoltzer, line 149: Mild preference to make this variable name match the methods.
```suggestion
    ignore_filename: Option<String>,
```

acfoltzer (changes requested):

wdl-analysis/src/analyzer.rs
```diff
@@ -409,13 +414,28 @@ where
...
+            let mut walker = WalkBuilder::new(&path);
+            if let Some(ignore_filename) = config.ignore_filename() {
+                walker.add_custom_ignore_filename(ignore_filename);
+            }
+            let walker = walker.standard_filters(false).parents(true).build();
+
+            for result in walker {
                 let entry = result.with_context(|| {
-                    format!("failed to read directory `{path}`", path = path.display())
+                    "error during recursive file search\nthis may be due to parsing an ignorefile"
```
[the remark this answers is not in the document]
acfoltzer: I would trim it down a bit, "recursive file search" might not mean much to a typical user. "error reading ignore file {ignore_filename}" would also give the user a hint of what filename to look for (it doesn't look like the display/`Error` impls of the `ignore::Error` point to a path)

acfoltzer (approved):

PR #586, opened 2025-08-26
fix: fix deserialization of `Object` to not require WDL identifiers.

### LINKED ISSUE #549: 1.1 spec test failure: `serde_pair`: deserialization of JSON objects to `Object` prevents allowable coercions
**Describe the bug**

Evaluate the following WDL:

```wdl
version 1.1

task tail {
  input {
    Pair[File, Int] to_tail
  }

  command <<<
  tail -n ~{to_tail.right} ~{to_tail.left}
  >>>

  output {
    Array[String] lines = read_lines(stdout())
  }
}

workflow serde_pair {
  input {
    Map[File, Int] to_tail
  }

  scatter (item in as_pairs(to_tail)) {
    call tail {
      input: to_tail = item
    }
    Pair[String, String]? two_lines = 
      if item.right >= 2 then (tail.lines[0], tail.lines[1]) else None
  }

  output {
    Map[String, String] tails_of_two = as_map(select_all(two_lines))
  }
}
```

With the inputs:

```json
{
  "serde_pair.to_tail": {
    "cities.txt": 2,
    "hello.txt": 1
  }
}
```

With `cities.txt` containing:

```
Houston
Chicago
Piscataway
```

And `hello.txt` containing:

```
hello
```

This fails with:

```
error: invalid input key `serde_pair.to_tail`

Caused by:
    object key `cities.txt` is not a valid WDL identifier
```

This is due to first deserializing the JSON map to an `Object` which we then attempt to coerce to the destination type of `Map[File, Int]`.

Note that `cities.txt` is _not_ a valid object key, but is a valid `File`.

**Expected behavior**

The evaluation should complete successfully with this output:

```json
{
  "serde_pair.tails_of_two": {
    "Houston": "Chicago"
  }
}
```

peterhuene:
This PR loosens the restriction on `Object` deserialization such that the keys are no longer required to be valid WDL identifiers.

This requirement was not in the WDL spec and prevented using the `Object` as an intermediary that would be coerced to a different value type, such as `Map`, for accessing the keys.

Additionally, this PR also implements support for type coercions between `Map` and `Object`/`Struct` where the key type is coercible to and from `String` instead of the previous requirement that the key type *be* `String`.

Fixes #549.

For external contributors:

For all contributors:

      changes (when appropriate).


a-frantz (commented):
this is a fun windows error message 😅 
```
tail: cannot open 'D:awdlwdlwdl-enginetestsworkflowsserde-paircities.txt' for reading: No such file or directory
```
LGTM otherwise. Will do another scan once the windows CI passes

peterhuene:
Yeah, not entirely sure why it's ending up that way (I suspect the normalization we try to do to make Windows paths make sense, but I would expect slashes instead of no backslashes).

Looking into the failure.

a-frantz (changes requested):
I think `wdl-analysis` also needs a changelog entry

wdl-analysis/src/types.rs
```diff
@@ -621,10 +623,10 @@ impl Coercible for CompoundType {
                 true
             }
 
-            // Struct -> Map[String, X], Struct -> Map[String, X]?, Struct? -> Map[String, X]? (if
-            // all struct members are coercible to X)
+            // Struct -> Map[X, Y], Struct -> Map[X, y]?, Struct? -> Map[X, Y]? where: String -> X
```
claymcleod, line 626:
```suggestion
            // Struct -> Map[X, Y], Struct -> Map[X, Y]?, Struct? -> Map[X, Y]? where: String -> X
```

peterhuene:
So apparently auto-merge skips in-progress CI? 👀

PR #592, opened 2025-09-03
[WIP]revise: lint rule toggling

a-frantz:


wdl-cli/src/analysis.rs
```diff
@@ -194,11 +210,32 @@ fn get_diagnostics_config(exceptions: &HashSet<String>) -> DiagnosticsConfig {
...
+                .any(|exception| exception.eq_ignore_ascii_case(rule.id()));
+        if enable {
+            enabled_rules.push(rule.id());
+        } else {
+            disabled_rules.push(rule.id());
+        }
+        enable
+    }));
+    info!("enabled lint rules: {:?}", enabled_rules);
+    info!("disabled lint rules: {:?}", disabled_rules);
```
a-frantz, line 239: this logs once per analysis thread. @peterhuene is there a better way we can log this only once?

peterhuene: I was just about to comment on this.

Probably the easiest way is to move the call of `get_lint_visitor` out of the closure (making it an `Option<Linter>` outside of the closure) which is cloned inside the closure and added to the validator if `Some`.

We'd need `Linter` to implement `Clone`, but there's no reason it couldn't be (should probably be a custom `Clone` impl that clones `rules`, but default initializes `document_exceptions`).

peterhuene (commented):
Other than the extraneous logging issue, I think this looks good!

wdl-cli/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+#### Changed
+
+* `Analysis` had it's `lint()` method replaced with `enabled_lint_tags()` and `disabled_lint_tags()` ([#592](https://github.com/stjude-rust-labs/wdl/pull/592)).
```
a-frantz, line 12:
```suggestion
* `Analysis` had its `lint()` method replaced with `enabled_lint_tags()` and `disabled_lint_tags()` ([#592](https://github.com/stjude-rust-labs/wdl/pull/592)).
```

a-frantz:
This should wait till this is resolved: https://github.com/stjude-rust-labs/sprocket/pull/164#issuecomment-3271745710
Then https://github.com/stjude-rust-labs/sprocket/pull/169 will be updated to be reviewed in tandem with this PR

